PR #9, opened 2020-02-24
Print Attributes in stdout exporter

nacho-undefinedlabs:
Print Span attributes in stdout exporter, so it can be more useful for debugging.

### DIFF
--- a/Sources/Exporters/Stdout/StdoutExporter.swift
+++ b/Sources/Exporters/Stdout/StdoutExporter.swift
@@ -32,6 +32,7 @@ public class StdoutExporter: SpanExporter {
             print("ParentSpanId: \(span.parentSpanId?.hexString ?? "no Parent")")
             print("Start: \(span.startEpochNanos)")
             print("Duration: \(span.endEpochNanos - span.startEpochNanos)")
+            print("Attributes: \(span.attributes)")
             print("------------------\n")
         }
         return .success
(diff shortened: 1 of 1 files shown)


Sources/Exporters/Stdout/StdoutExporter.swift
```diff
@@ -32,6 +32,7 @@ public class StdoutExporter: SpanExporter {
             print("ParentSpanId: \(span.parentSpanId?.hexString ?? "no Parent")")
```
SergeyKanzhelev: this PR is fine. Generally though we are trying to output to stdout in a form of JSON in a single line. It is more applicable for server side probably where stdout can be consumed programmatically by fluentBit or something like this

Sources/Exporters/Stdout/StdoutExporter.swift
```diff
@@ -32,6 +32,7 @@ public class StdoutExporter: SpanExporter {
...
             print("Start: \(span.startEpochNanos)")
             print("Duration: \(span.endEpochNanos - span.startEpochNanos)")
```
SergeyKanzhelev:
```suggestion
            print("Duration: \(span.endEpochNanos - span.startEpochNanos) nanoseconds")
```

Sources/Exporters/Stdout/StdoutExporter.swift
nacho-undefinedlabs: Yes, it has much more sense. Created an issue to track this change: #11

PR #10, opened 2020-02-24
Keep the scope in the span context

nacho-undefinedlabs:
Keep the scope where a span was created in the span context, so we can end it when the span ends.

Make the handling as a protocol extension.

### DIFF
--- a/Sources/OpenTelemetryApi/Trace/Span.swift
+++ b/Sources/OpenTelemetryApi/Trace/Span.swift
@@ -128,6 +128,16 @@ public enum SpanAttributeConstants: String {
 }
 
 extension Span {
+    public func end() {
+        context.scope?.close()
+    }
+
+    public func end(endOptions: EndSpanOptions) {
+        end()
+    }
+}
+
+extension Span {
     public func setAttribute(key: String, value: String?) {
         return setAttribute(key: key, value: AttributeValue.string(value))
     }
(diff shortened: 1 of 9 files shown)


Sources/OpenTelemetryApi/Trace/Span.swift
```diff
@@ -127,6 +127,16 @@ public enum SpanAttributeConstants: String {
     case httpRouteKey = "http.route"
 }
 
+extension Span {
+    public func end() {
+        context.scope?.close()
+    }
+
+    public func end(endOptions: EndSpanOptions) {
```
SergeyKanzhelev: why are you not taking the timestamp from the `EndSpanOptions`? Is it because it's just an abstract class implementation?

Tests/OpenTelemetrySdkTests/Trace/SpanBuilderSdkTests.swift
```diff
@@ -221,12 +220,11 @@ class SpanBuilderSdkTest: XCTestCase {
 
     func testParentCurrentSpan() {
```
SergeyKanzhelev: should any of these tests validate that the original span was restored on the context?

Sources/OpenTelemetryApi/Trace/Span.swift
nacho-undefinedlabs: Yes, its like an abstract class implementation, it has no notion of timestamp

Tests/OpenTelemetrySdkTests/Trace/SpanBuilderSdkTests.swift
nacho-undefinedlabs: Absolutely, just added a test to check it. Thanks a lot

Issue #19, opened 2020-03-25
 Implementation of the B3 propagation protocol.


nacho-undefinedlabs:
It should be located in /OpenTelemetryApi/Trace/Propagation. 
See [https://github.com/openzipkin/b3-propagation](https://github.com/openzipkin/b3-propagation). 
Java opentelemetry client can be a good reference  for implementation details: [https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/propagation/B3Propagator.java](https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/propagation/B3Propagator.java)

RayZhao1998:
I'm working on this now.

nacho-undefinedlabs:
Closed with #26

PR #26, opened 2020-04-17
feat: support b3-propagation

RayZhao1998:
Support b3-propagation and add b3-propagation unit tests. #19


nacho-undefinedlabs:
Very good work,  just missing some tests for the invalid results when format is invalid. 
(nitpick some test names show "sindleHeader" instead of "SIngleHeader")

PR #30, opened 2020-04-28
Fix Scope Testing action for external PR

nacho-undefinedlabs:
GitHub Actions doesn't support accessing the repository secrets when running in an external pull request so it fails.
Its better having this value public than avoiding external PR to run the action and tests


SergeyKanzhelev:
should I remove it from secrets than?

nacho-undefinedlabs:
Yes, it is not needed anymore. I didn’t know  GitHub actions didn’t allow PR access the configured secrets as other CI do.

PR #35, opened 2020-05-11
Span exporter changes

nacho-undefinedlabs:
Span exporter changes:
* SpanExporter protocol includes a flush() function
* return only failure or succes (not failedRetryable or failedNotRetryable)
* TracerSDK provider now implements a forceFlush() method that calls current processor flush


Sources/OpenTelemetrySdk/Trace/Export/SpanExporter.swift
```diff
@@ -24,6 +24,9 @@ public protocol SpanExporter: AnyObject {
     /// - Parameter spans: the list of sampled Spans to be exported.
     @discardableResult func export(spans: [SpanData]) -> SpanExporterResultCode
 
+    ///Exports the collection of sampled Spans that have not yet been exported.
```
SergeyKanzhelev, line 27: does it just schedule for sending or actually sending it no matter how long it takes?

Sources/OpenTelemetrySdk/Trace/TracerSdkProvider.swift
```diff
@@ -78,4 +78,9 @@ public class TracerSdkProvider: TracerProvider {
         }
         sharedState.stop()
     }
+
+    ////Requests the active span processor to process all span events that have not yet been processed.
```
SergeyKanzhelev, line 82: is there a difference in swift - how many slashes comment starts wih? In this PR - there are 2, 3 and 4 slashes comments

Sources/OpenTelemetrySdk/Trace/Export/SpanExporter.swift
nacho-undefinedlabs: This is the abstract method to be implemented, no currently exporter implements it further than simple return.

Sources/OpenTelemetrySdk/Trace/TracerSdkProvider.swift
nacho-undefinedlabs: Yes, when using 3 slashes Xcode parses the code to generate documentation, two slashes used for standard comments . This 4 slashes was a typo :)

PR #36, opened 2020-05-11
Changes in AttributeValue

nacho-undefinedlabs:
Changes in AttributeValue:
* Add support for arrays in values
* Allow usage of empty strings
* Remove attribute when using nil value


Sources/Exporters/Jaeger/Adapter.swift
```diff
@@ -109,6 +114,18 @@ final class Adapter {
         case let .double(value):
             vType = .double
             vDouble = value
+        case .stringArray(let value):
```
SergeyKanzhelev, line 117: I wonder if array of arrays supported?

nacho-undefinedlabs: No, only arrays of previous possible attribute values: string, bool and numeric (double or int). Should array of arrays be supported?

PR #40, opened 2020-06-04
Small changes to sampling

nacho-undefinedlabs:
Don't make Sample depend on SpanId
Allow changing sampling probability from TraceConfig with a percentage directly


SergeyKanzhelev:
@nacho-undefinedlabs can you please make sure to sign EasyCLA. So we can switch repo to it

PR #42, opened 2020-06-11
Testing on forks

nacho-undefinedlabs:
It uses the public `SCOPE_DSN` for pull requests but It uses the repository `secret.SCOPE_DSN` on push so forks can test before creating a PR

Also force to use  previous agent version (0.6.2), since versions from 0.7.0 uses opentelemetry internally and when testing  swift clashes with the symbols and does funny thing on runtime.

### DIFF
--- a/.github/workflows/ScopeTesting.yml
+++ b/.github/workflows/ScopeTesting.yml
@@ -16,6 +16,7 @@ jobs:
         dsn: ${{ secrets.SCOPE_DSN }}
         platform: macos
         codePath: true
+        agentVersion: 0.6.2
     - name: Testing for macOS (only run on pull_request)
       if: github.event_name == 'pull_request'
       uses: undefinedlabs/scope-for-swift-action@v1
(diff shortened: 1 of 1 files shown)


nacho-undefinedlabs:
in the process of signing EasyCLA, will wait until EasyCLA is solved before merging

.github/workflows/ScopeTesting.yml
```diff
@@ -9,9 +9,19 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Testing for macOS
+    - name: Testing for macOS (only on push)
+      if: github.event_name == 'push'
+      uses: undefinedlabs/scope-for-swift-action@v1
+      with:
+        dsn: ${{ secrets.SCOPE_DSN }}
```
SergeyKanzhelev, line 16: I believe the secret was removed from the repo (https://github.com/open-telemetry/opentelemetry-swift/pull/30/files). Or it refers to another secret?

nacho-undefinedlabs: Yes, but I added it again ( with the same value that is published) I could not think any other way a fork can use its own value in its own commits and use the official one when doing a PR. The root problem is that Pull Requests in GitHub actions don’t inherit secrets from the repo where they are creating it, so need to use its own value for every commit, and use the global only on PR

PR #43, opened 2020-06-15
Add a Prometheus exporter

### LINKED ISSUE #39: Implement a Prometheus exporter for metrics

nacho-undefinedlabs:
It includes both the exporter and a simple example app. 
Also some changes to the API needed to create the exporter

Close #39

### DIFF
--- a/Sources/Exporters/Prometheus/PrometheusMetric.swift
+++ b/Sources/Exporters/Prometheus/PrometheusMetric.swift
@@ -28,19 +28,18 @@ struct PrometheusMetric {
     var description: String
     var type: String = ""
 
-    func write() -> String {
-        
+    func write(timeStamp: String) -> String {
         var output = ""
-        
+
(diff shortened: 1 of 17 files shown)


Sources/Exporters/Prometheus/PrometheusMetric.swift
```diff
@@ -0,0 +1,117 @@
...
+        
+        if !type.isEmpty {
+            output += "# TYPE "
+            output += name
+            output += " "
+            output += type
+            output += "\n"
+        }
+        
+        let now = String(Int64((Date().timeIntervalSince1970 * 1000.0).rounded()))
```
SergeyKanzhelev, line 52: should we have a single `now` value for all metrics in a single output?

nacho-undefinedlabs: Yes, thats a very good idea, metrics are already collected and we wont be adding the time taken for writing them, I will add this change

PR #47, opened 2020-06-19
Implement  a Zipkin exporter

### LINKED ISSUE #18: Implement a zipkin exporter

nacho-undefinedlabs:
Also added some imporvemtns to Api and Sdk that were obvious while developing this exporter

Close #18


Sources/Exporters/Zipkin/Implementation/ZipkinConversionExtension.swift
```diff
@@ -0,0 +1,159 @@
...
+            processResources(state: &attributeEnumerationState, key: $0.key, value: $0.value)
+        }
+
+        var localEndpoint = defaultLocalEndpoint
+
+        var serviceName = attributeEnumerationState.serviceName
+
+        if serviceName != nil, !serviceName!.isEmpty {
+            if let serviceNamespace = attributeEnumerationState.serviceNamespace, !serviceNamespace.isEmpty {
+                serviceName = serviceNamespace + "." + serviceName!
```
SergeyKanzhelev, line 61: I think this logic was implemented in places, but spec forbid to combine namespace and name

Sources/Exporters/Zipkin/Implementation/ZipkinConversionExtension.swift
```diff
@@ -0,0 +1,159 @@
...
+    private static func toSpanKind(otelSpan: SpanData) -> String {
+        switch otelSpan.kind {
+        case .server:
+            return "SERVER"
+        case .producer:
+            return "PRODUCER"
+        case .consumer:
+            return "CONSUMER"
+        default:
+            return "CLIENT"
```
SergeyKanzhelev, line 125: I think default is empty string or something. Please refer to the spec

Sources/Exporters/Zipkin/Implementation/ZipkinSpan.swift
```diff
@@ -0,0 +1,94 @@
...
+                object["timestamp"] = $0.timestamp
+                object["value"] = $0.value
+                return object
+            }
+
+            output["annotations"] = annotationsArray
+        }
+
+        if tags.count > 0 {
+//            let tagsArray: [Any] = tags.map {
```
SergeyKanzhelev, line 83: do you want to keep commented lines?

Sources/Exporters/Zipkin/Implementation/ZipkinConversionExtension.swift
nacho-undefinedlabs: True, fixed and added the namespace as a Zipkin tag as the spec says.

Sources/Exporters/Zipkin/Implementation/ZipkinConversionExtension.swift
nacho-undefinedlabs: True, should not be set

Sources/Exporters/Zipkin/Implementation/ZipkinSpan.swift
nacho-undefinedlabs: Nope, thanks for catching

PR #49, opened 2020-06-23
Added Open telemetry protocol GRPC exporter

### LINKED ISSUE #22: OTel proto's as models?
👋 I was just stumbling onto this repo, and noticed that it didn't reflect any of the datamodels from the protobuf IDLs in https://github.com/open-telemetry/opentelemetry-proto

I didn't want to disrupt this, but let you know that I grabbed those core protos and created a swift package to expose them. Wanted to share in case it would be useful to you:

https://github.com/heckj/OpenTelemetryModels

It's just the proto files generated into models, nothing more - but may still be useful to you.

### LINKED ISSUE #29: Implement exporter for OpenTelemetry collector

### LINKED ISSUE #48: iOS support
Hello and thanks for your job, you're doing great!
I'm an iOS developer and I was wondering if it was possible to use this library for a mobile application.
I was pretty happy to see the last commits about the Zipkin exporter but after I imported the package on my project I realize that it seems to be only for the macOS environment, isn't it?

On my side, I can say I'm not very confident with this kind of development or technology, in fact, I'm asking you if it is possible or not.

nacho-undefinedlabs:
* Created the OpenTelemetryProtocolExporter and tests for it
* Added swift-proto models
* Changes to Api and sSdk: make timestamps and epoch times Uint64 instead of Int64
* Also fixed Zipkin exporter compilation on iOS 

Close #29 
Close #22
Fixes #48


Examples/Logging Tracer/LoggingSpan.swift
```diff
@@ -67,23 +67,23 @@ class LoggingSpan: Span {
...
+    public func addEvent(name: String, timestamp: UInt64) {
         Logger.log("Span.addEvent(\(name) timestamp:\(timestamp))")
     }
 
     public func addEvent(name: String, attributes: [String: AttributeValue]) {
         Logger.log("Span.addEvent(\(name), attributes:\(attributes) )")
     }
 
-    public func addEvent(name: String, attributes: [String: AttributeValue], timestamp: Int64) {
+    public func addEvent(name: String, attributes: [String: AttributeValue], timestamp: UInt64) {
```
SergeyKanzhelev, line 78: I'm curious if there is a built-in type for timestamp in swift?

nacho-undefinedlabs: The most native way for timestamps in iOS is using TimeInterval that is basically a Double value containing number of seconds (with sub milliseconds precision). I thought about using it at the beginning of development, but initially discarded. I can reevaluate now with most of the library ready if it flows well with all the api's

PR #51, opened 2020-06-24
Support for creating TraceId with 64 bit hexString

### LINKED ISSUE #32: B3 Propagator should left-pad 64-bit trace IDs with zeros
As noted in https://github.com/open-telemetry/opentelemetry-go/issues/686 and discussed at the SIG Spec meeting today, we should left-pad 64-bit trace ID values in B3 propagators to allow for interoperability with systems that produce 64-bit trace IDs.  It appears that the [traceID initializer](https://github.com/open-telemetry/opentelemetry-swift/blob/master/Sources/OpenTelemetryApi/Trace/TraceId.swift#L129) does not allow for trace IDs that are less than 32 hex digits.

nachoBonafonte:
B3 propagators can use 64 bit for trace ID

Fixes #32


Sources/OpenTelemetryApi/Trace/TraceId.swift
```diff
@@ -127,17 +127,24 @@ public struct TraceId: Comparable, Hashable, CustomStringConvertible, Equatable
...
+        if hex.count >= 32 + offset {
+            let firstIndex = hex.index(hex.startIndex, offsetBy: offset)
+            let secondIndex = hex.index(firstIndex, offsetBy: 16)
+            let thirdIndex = hex.index(secondIndex, offsetBy: 16)
+            if let idHi = UInt64(hex[firstIndex ..< secondIndex], radix: 16),
+                let idLo = UInt64(hex[secondIndex ..< thirdIndex], radix: 16) {
+                self.init(idHi: idHi, idLo: idLo)
+                return
+            }
+        } else if hex.count >= 32 + offset {
```
SergeyKanzhelev, line 139: is it the same condition as above in `if`?

nachoBonafonte: Yes, it is. It also means the test I added was not correct

PR #52, opened 2020-06-24
Update Readme with latest develpments added

nachoBonafonte:
Also fixed zipkin exporter taking a random ips for local endpoints, take always first


nacho-undefinedlabs:
@SergeyKanzhelev, after all this changes, what would be missing in the project for releasing an official beta?

PR #56, opened 2020-06-26
Change the ZipkinTraceExporterOption to be configurable from outside of the package

xxZap:


xxZap:
@nachoBonafonte Hey, I'm not very practical on this kind of operation. I think I firstly committed with my company account and not my personal one and now that commit is missing my personal User's ID. Should I restart from zero e re-commit with my personal account?

PR #64, opened 2020-07-25
Zipkin duration unit is microseconds, calculate it correctly

### LINKED ISSUE #63: ZipkinSpan wrong duration
Hey guys,
I'm using ZipkinExporter to send some simple Spans to Zipkin but I'm noticing some errors:
```
2020-07-22 15:55:09.125 DEBUG 1 --- [orker-epoll-2-1] z.s.i.ZipkinHttpCollector                : Cannot store spans [0b844295a4fdf705d4eb85f794e4ec34/07f1937b25588e92] due to CompletionException(com.linecorp.armeria.common.grpc.protocol.ArmeriaStatusException: Invalid 2020-08-14T23:00:31.216435-07:00: Invalid end time)
java.util.concurrent.CompletionException: com.linecorp.armeria.common.grpc.protocol.ArmeriaStatusException: Invalid 2020-08-14T23:00:31.216435-07:00: Invalid end time
	at java.util.concurrent.CompletableFuture.encodeThrowable(Unknown Source) ~[?:?]
	at java.util.concurrent.CompletableFuture.completeThrowable(Unknown Source) ~[?:?]
	at java.util.concurrent.CompletableFuture$UniApply.tryFire(Unknown Source) ~[?:?]
	at java.util.concurrent.CompletableFuture.postComplete(Unknown Source) ~[?:?]
	at java.util.concurrent.CompletableFuture.complete(Unknown Source) ~[?:?]
```

It seems that the problem is **Invalid end time** in fact the sent date is in August but we still are in July so, after some investigation I noticed that when the `ZipkinConversionExtension` is parsing the Span and it's creating the `ZipkinSpan`, it does the following:

```
return ZipkinSpan(traceId: ZipkinConversionExtension.EncodeTraceId(traceId: otelSpan.traceId, useShortTraceIds: useShortTraceIds),
                          parentId: parentId,
                          id: ZipkinConversionExtension.EncodeSpanId(spanId: otelSpan.spanId),
                          kind: ZipkinConversionExtension.toSpanKind(otelSpan: otelSpan),
                          name: otelSpan.name,
                          timestamp: otelSpan.startEpochNanos / 1000,
                          duration: ( otelSpan.endEpochNanos - otelSpan.startEpochNanos) * 1000000,
                          localEndpoint: localEndpoint,
                          remoteEndpoint: remoteEndpoint,
                          annotations: annotations,
                          tags: attributeEnumerationState.tags,
                          debug: nil,
                          shared: nil)
```

Is it right that the duration is calculated in that way? 
Shouldn't be
`                          duration: ( otelSpan.endEpochNanos - otelSpan.startEpochNanos) / 1000`

nachoBonafonte:
Fixes #63

### DIFF
--- a/Sources/Exporters/Zipkin/Implementation/ZipkinConversionExtension.swift
+++ b/Sources/Exporters/Zipkin/Implementation/ZipkinConversionExtension.swift
@@ -91,7 +91,7 @@ struct ZipkinConversionExtension {
                           kind: ZipkinConversionExtension.toSpanKind(otelSpan: otelSpan),
                           name: otelSpan.name,
                           timestamp: otelSpan.startEpochNanos / 1000,
-                          duration: ( otelSpan.endEpochNanos - otelSpan.startEpochNanos) / 1000,
+                          duration: (otelSpan.endEpochNanos - otelSpan.startEpochNanos) / 1000,
                           localEndpoint: localEndpoint,
                           remoteEndpoint: remoteEndpoint,
                           annotations: annotations,
--- a/Sources/Exporters/Zipkin/Implementation/ZipkinConversionExtension.swift
+++ b/Sources/Exporters/Zipkin/Implementation/ZipkinConversionExtension.swift
@@ -91,7 +91,7 @@ struct ZipkinConversionExtension {
                           kind: ZipkinConversionExtension.toSpanKind(otelSpan: otelSpan),
                           name: otelSpan.name,
                           timestamp: otelSpan.startEpochNanos / 1000,
-                          duration: ( otelSpan.endEpochNanos - otelSpan.startEpochNanos) * 1000000,
+                          duration: ( otelSpan.endEpochNanos - otelSpan.startEpochNanos) / 1000,
                           localEndpoint: localEndpoint,
                           remoteEndpoint: remoteEndpoint,
                           annotations: annotations,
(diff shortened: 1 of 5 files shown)


Sources/Exporters/Zipkin/Implementation/ZipkinConversionExtension.swift
```diff
@@ -91,7 +91,7 @@ struct ZipkinConversionExtension {
                           kind: ZipkinConversionExtension.toSpanKind(otelSpan: otelSpan),
                           name: otelSpan.name,
                           timestamp: otelSpan.startEpochNanos / 1000,
-                          duration: ( otelSpan.endEpochNanos - otelSpan.startEpochNanos) * 1000000,
+                          duration: ( otelSpan.endEpochNanos - otelSpan.startEpochNanos) / 1000,
```
SergeyKanzhelev, line 94: while you are on this - maybe also check if it's greater than zero.

SergeyKanzhelev, line 94:
```suggestion
                          duration: (otelSpan.endEpochNanos - otelSpan.startEpochNanos) / 1000,
```

nachoBonafonte: duration is already unsigned, so no negative number here

PR #69, opened 2020-10-07
First version of the Datadog Exporter

nachoBonafonte:
*This is a work in progress, and currently in an alpha stage, should not be used in production.*
* It will convert Opentelemetry spans to Datadog traces and span Events to Datadog logs. There is still no support for metrics
* It is currently only using the endpoints for mobile applications and Client tokens, will be expanded in the future.


SergeyKanzhelev (approved):
glanced over, lgtm as long as it was tested

PR #70, opened 2020-10-14
Datadog exporter: Fully synchronize on flush

nachoBonafonte:
Previous to this change, flush method didn't block before sending all remaining spans.

### DIFF
--- a/Sources/Exporters/DatadogExporter/Persistence/FileReader.swift
+++ b/Sources/Exporters/DatadogExporter/Persistence/FileReader.swift
@@ -62,6 +62,21 @@ internal final class FileReader {
         return nil
     }
 
+    func onRemainingBatches(process: (Batch)->()) -> Bool {
+        queue.sync {
+            do {
+                try orchestrator.getAllFiles(excludingFilesNamed: Set(filesRead.map { $0.name }))?.forEach {
+                    let fileData = try $0.read()
+                    let batchData = dataFormat.prefixData + fileData + dataFormat.suffixData
+                    process(Batch(data: batchData, file: $0))
+                }
+            } catch {
+                return false
+            }
+            return true
+        }
+    }
+
     // MARK: - Accepting batches
 
     func markBatchAsRead(_ batch: Batch) {
@@ -70,7 +85,7 @@ internal final class FileReader {
         }
     }
 
-    private func synchronizedMarkBatchAsRead(_ batch: Batch) {
+    func synchronizedMarkBatchAsRead(_ batch: Batch) {
         orchestrator.delete(readableFile: batch.file)
         filesRead.append(batch.file)
     }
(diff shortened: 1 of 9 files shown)


Sources/Exporters/DatadogExporter/Persistence/FilesOrchestrator.swift
```diff
@@ -113,6 +113,16 @@ internal class FilesOrchestrator {
         }
     }
 
+    func getAllFiles(excludingFilesNamed excludedFileNames: Set<String> = []) -> [ReadableFile]? {
+        do {
+            return try directory.files()
+                .filter({ excludedFileNames.contains($0.name) == false })
+        } catch {
+            print("🔥 Failed to obtain readable files: \(error)")
```
SergeyKanzhelev, line 121: is it typical practice to log with emojis? Interesting...

Sources/Exporters/DatadogExporter/Persistence/FileReader.swift
```diff
@@ -62,6 +62,21 @@ internal final class FileReader {
         return nil
     }
 
+    func onRemainingBatches(process: (Batch)->()) -> Bool {
+        queue.sync {
+            do {
+                try orchestrator.getAllFiles(excludingFilesNamed: Set(filesRead.map { $0.name }))?.forEach {
```
SergeyKanzhelev, line 68: so this will take the list of files once and any new file will be ignored, correct? Nothing prevents code from writing new ones while flushing, right?

SergeyKanzhelev: just making sure there is no chance flush will never end keep reading and producing files...

SergeyKanzhelev (approved):
lgtm with couple questions

Sources/Exporters/DatadogExporter/Persistence/FileReader.swift
SergeyKanzhelev: perhaps say it explicitly

Sources/Exporters/DatadogExporter/Persistence/FilesOrchestrator.swift
nachoBonafonte: It is on swift, since it has always had full unicode support, and in fact is quite useful for visualising errors in a long log

Sources/Exporters/DatadogExporter/Persistence/FileReader.swift
nachoBonafonte: Writer should not be working when this code is run, but anyway the getAllFIles() method gets all the remaining file at once, so if any other file appeared after that read while the previous are being sent it wont be read. 
But as you said much better having this info as a comments.

nachoBonafonte:
Hi @SergeyKanzhelev, it looks like EasyCLA is frozen, I even tried updating the Readme without much results. Could you merge it with your admin superpowers?

SergeyKanzhelev:
> Hi @SergeyKanzhelev, it looks like EasyCLA is frozen, I even tried updating the Readme without much results. Could you merge it with your admin superpowers?

Same on other repos... looking into it

SergeyKanzhelev:
reopened for EasyCLA to kick in. It was an issue on EasyCLA side

PR #74, opened 2020-10-26
Datadog Exporter: Minor improvements and new example

nachoBonafonte:
* Use a non default source for logs
* Improve error encoding, now we use Opentelemetry status instead of a boolean attribute
* Add a simple datadog example where configuration of exporter is shown

### DIFF
--- a/Examples/Datadog Sample/main.swift
+++ b/Examples/Datadog Sample/main.swift
@@ -1,0 +1,79 @@
+// Copyright 2020, OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


Examples/Datadog Sample/main.swift
```diff
@@ -0,0 +1,79 @@
...
+            childScope.close()
+        }
+        childSpan.end()
+        scope.close()
+    }
+    span.end()
+}
+
+let exporterConfiguration = ExporterConfiguration(
+    serviceName: "Otel exporter Example",
```
SergeyKanzhelev, line 59: one suggestion is to take the `servicename` from resources API. See [cnvention](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/semantic_conventions/README.md#user-content-service:~:text=Logical%20name%20of%20the%20service.%20%5B1%5D) and example [in this post](https://medium.com/opentelemetry/opentelemetry-beyond-getting-started-5ac43cd0fe26#e300:~:text=%7B%20%E2%80%9Cservice.name%E2%80%9D%2C%20%E2%80%9CMyBackEndApp%E2%80%9D%20%7D%2C)

SergeyKanzhelev (approved):

Examples/Datadog Sample/main.swift
nachoBonafonte: That's a good idea, but after trying it it looks like the api for modifying the resources is not very functional. I will add it as an issue to the project for a future fix.

PR #76, opened 2020-10-27
RecordsEventReadableSpan: Fix potential thread read conditions when adding attributes from different threads

nachoBonafonte:
There are a pair of thread race conditions fixed. Access to `hasEnded` and `events` members could happen from different threads


SergeyKanzhelev:
> There are a pair of thread race conditions fixed. Access to `hasEnded` and `events` members could happen from different threads

so what's wrong with this race? Wouldn't it be ok just to add another event even if span ended? It may be better than adding a lock on every add.

nachoBonafonte:
Yes, maybe is too overwhelming solution. I'll close and think more about this issue

PR #77, opened 2020-10-29
Datadog exporter: Add a config setting to avoid exporting unsampled spans or logs from unsampled spans

nachoBonafonte:
Datadog exporter: Add a config setting to avoid exporting unsampled spans or logs from unsampled spans

### DIFF
--- a/Sources/Exporters/DatadogExporter/DatadogExporter.swift
+++ b/Sources/Exporters/DatadogExporter/DatadogExporter.swift
@@ -33,8 +33,12 @@ public class DatadogExporter: SpanExporter {
 
     public func export(spans: [SpanData]) -> SpanExporterResultCode {
         spans.forEach {
-            spansExporter.exportSpan(span: $0)
-            logsExporter.exportLogs(fromSpan: $0)
+            if $0.traceFlags.sampled || configuration.exportUnsampledSpans {
+                spansExporter.exportSpan(span: $0)
+            }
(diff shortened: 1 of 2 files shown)


Sources/Exporters/DatadogExporter/DatadogExporter.swift
```diff
@@ -33,8 +33,12 @@ public class DatadogExporter: SpanExporter {
 
     public func export(spans: [SpanData]) -> SpanExporterResultCode {
         spans.forEach {
-            spansExporter.exportSpan(span: $0)
-            logsExporter.exportLogs(fromSpan: $0)
+            if $0.traceFlags.sampled || configuration.exportUnsampledSpans {
```
SergeyKanzhelev, line 36: I'm surprised sampled spans coming to exporter at all. Batching processor suppose to take care of filtering them out. I will LGTM anyway, but definitely something to look at going forward

SergeyKanzhelev (approved):

Sources/Exporters/DatadogExporter/DatadogExporter.swift
nachoBonafonte: Yes, both BatchSpanProcessor and SimpleSpanProcessor filter unsampled spans by default, but you can configure them to not filter. I have a use case where I want to export the logs but not the spans where they were associated, and I can imagine other special cases were this configuration can be useful.

PR #80, opened 2020-11-25
Datadog Exporter: Fix some system version support for macOS and tvOS

nachoBonafonte:
Some of the availability checks for OS versions were not correct, and it failed compiling on them


SergeyKanzhelev (approved):
I have no context here, but looks ok

PR #83, opened 2020-12-11
Use numbered version of opentracing-objc library

nachoBonafonte:
Swift Package manager forces to use a numbered version of opentracing-objc library, so our library can also be used with a numbered version


SergeyKanzhelev (approved):
I have no context here, but looks OK

PR #84, opened 2020-12-16
Update Tracing (Api and Sdk) to latest spec

nachoBonafonte:
Many files touched, but mostly are renames for consistency, changes include:
* renaming several classes and attributes to be more consistent with current spec
* Added helpers to simplify seconds -> nanoseconds conversion
* Reworked status values to: ok, error, unset
* Simplified api moving Events and Links to be part of SpanData in Sdk
* Remove TimedEvent, since Event now has the timestamp property always
* Increased default maximum number of attributes, events and links to many values, reducing is easy, but wont lose values by default


Sources/OpenTelemetrySdk/Internal/MillisClock.swift
```diff
@@ -23,10 +23,10 @@ public class MillisClock: Clock {
     }
 
     public var now: UInt64 {
```
SergeyKanzhelev, line 25: should this class be public? Also why wouldn't SDK operate with `Date` and only convert to nanos on export?

Tests/ExportersTests/OpenTelemetryProtocol/SpanAdapterTests.swift
```diff
@@ -91,104 +91,104 @@ class SpanAdapterTests: XCTestCase {
...
-        eventWithAttrib.name = "test_with_attributes"
-
-        var attribute = Opentelemetry_Proto_Common_V1_KeyValue()
-        attribute.key = "key_string"
-        attribute.value = Opentelemetry_Proto_Common_V1_AnyValue()
-        attribute.value.stringValue = "string"
-        eventWithAttrib.attributes = [attribute]
-
-        XCTAssertEqual(SpanAdapter.toProtoSpanEvent(event: TimedEvent(name: "test_with_attributes", epochNanos: 12345, attributes: ["key_string": AttributeValue.string("string")])), eventWithAttrib)
+//        status = Opentelemetry_Proto_Trace_V1_Status()
```
SergeyKanzhelev, line 94: just remove the commented code?

SergeyKanzhelev (approved):
reviewed API changes - looks good.

Sources/OpenTelemetrySdk/Internal/MillisClock.swift
nachoBonafonte: Yes, probably shouldnt be public, I will change. 
About using directly `Date` I remember having doubt about it at the beggining of the project, in fact there are some functions that support `Date` for users, but don't remember why I decided to use nanos at the end. I can revisit it in a future. I will create an issue with it

Tests/ExportersTests/OpenTelemetryProtocol/SpanAdapterTests.swift
nachoBonafonte: Just commented to avoid the test to fail temporarily, I plan to update the proto spec in the next PR so some of this will be needed.

PR #90, opened 2021-01-05
Add startTimestamp to MetricData

nachoBonafonte:
Add startTimestamp to MetricData, so we can support time ranges.
It implies changing Aggregator from protocol to class.
Minor fix in Prometheus exporter


SergeyKanzhelev (approved):
I suggest to remove commented lines in Sources/OpenTelemetrySdk/Metrics/Aggregators/Aggregator.swift

PR #91, opened 2021-01-08
Use Date instead of nanoseconds

nachoBonafonte:
Use time constructions based on native Date instead of using nanoseconds, convert to nanoseconds only on exporting.
It makes usage more natural from a user standpoint and avoids conversion on its side.
It also modifies some parameter names to be consistent with actual parameters


Sources/Exporters/DatadogExporter/Persistence/FilesOrchestrator.swift
```diff
@@ -183,3 +183,24 @@ internal func fileCreationDateFrom(fileName: String) -> Date {
     let millisecondsSinceReferenceDate = TimeInterval(UInt64(fileName) ?? 0) / 1_000
     return Date(timeIntervalSinceReferenceDate: TimeInterval(millisecondsSinceReferenceDate))
 }
+
+private enum FixedWidthIntegerError<T: BinaryFloatingPoint>: Error {
```
SergeyKanzhelev, line 187: cannot find the use of this new enum. Is it part of this PR?

SergeyKanzhelev (approved):

Sources/Exporters/DatadogExporter/Persistence/FilesOrchestrator.swift
nachoBonafonte: yes is only used privately in this Datadog exporter module, and is used in line 202.
This was added here after removing SwiftExtension file because some of those hekpers were moved to Api

PR #96, opened 2021-01-15
Several minor improvements including:

nachoBonafonte:
Several minor improvements including:
* Import "swift-atomics" library and use it to protect hasEnded attribute in RecordEventsReadableSpan
* Provide a isError attribute to Status so it can easily be checked, now that unset is the default
* Datadog exporter: Fix reporting unset as error, using previous attribute
* Clean some api tests (previously in sdk) that were incorrectly importing OpenTelemetrySdk

### DIFF
--- a/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
+++ b/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
@@ -87,11 +87,8 @@ public class RecordEventsReadableSpan: ReadableSpan {
     /// The end time of the span.
     public private(set) var endTime: Date?
     /// True if the span is ended.
-    private var endAtomic = ManagedAtomic<Bool>(false)
-    public private(set) var hasEnded: Bool {
-        set {
-            self.endAtomic.store(newValue, ordering: .relaxed)
-        }
+    fileprivate var endAtomic = ManagedAtomic<Bool>(false)
+    public var hasEnded: Bool {
         get {
             return self.endAtomic.load(ordering: .relaxed)
         }
@@ -222,11 +219,10 @@ public class RecordEventsReadableSpan: ReadableSpan {
     }
 
     public func setAttribute(key: String, value: AttributeValue?) {
-        if hasEnded {
-            return
-        }
-
         attributesSyncLock.withLockVoid {
+            if !isRecording {
+                return
+            }
             if value == nil {
                 attributes.removeValueForKey(key: key)
             }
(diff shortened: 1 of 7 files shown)


Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
```diff
@@ -86,7 +87,15 @@ public class RecordEventsReadableSpan: ReadableSpan {
     /// The end time of the span.
     public private(set) var endTime: Date?
     /// True if the span is ended.
-    public private(set) var hasEnded: Bool = false
+    private var endAtomic = ManagedAtomic<Bool>(false)
```
SergeyKanzhelev, line 90: Why is it needed? it doesn't synchronize threads and the only thing this variable protects is whether the new attribute will be added after the span has ended. Not a big deal I think and may not worth an atomic variable on a critical path

SergeyKanzhelev:
> Clean some api tests (previously in sdk) that were incorrectly importing OpenTelemetrySdk

is it part of PR? couldn't find it

Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
nachoBonafonte: Is is also checked when adding events, I am finding this issue in my workflow because I am adding events from a secondary thread while the span ends

nachoBonafonte:
Yes, just a pair of deleted imports in the last 2 files

Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
SergeyKanzhelev: you are not achieving any synchronization here, correct? Without atomic the only think will get better is when even was being added exactly at the moment span is being ended. Not much improvement over non-atomic?

nachoBonafonte: I modified the code a bit, so it doesn't have any atomic check in the critical path. 
It still uses the hasEnded atomic value to avoid the span to be ended twice from different threads and because is part of the public API and don't know how it is going to be used from outside.
But for the addAttribute() or addEvent() the code is now using the isRecording variable instead, that is only updated after end() has been called.

SergeyKanzhelev (approved):

Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
SergeyKanzhelev: great idea to split the check into two variables.

PR #99, opened 2021-01-22
Fix excessive memory usage in PushMetricController and BatchSpanProcessor

nachoBonafonte:
As their job is done in an infinite loop, wrap each execution in an autoreleasepool so memory can be recovered in a timely manner, if not memory continue growing until stopped.


SergeyKanzhelev (approved):

SergeyKanzhelev:
swift is an interesting language requiring this... =)

nachoBonafonte:
Yes, it's not usually needed, but in some special cases it is. Memory is marked as deleted but memory is only reclaimed at the end of the event loop so it can grow quite a bit.

PR #101, opened 2021-01-25
Fix package URL in README

johnbley:
The URL in the README points to the old repo, meaning that people who blindly copy+paste from there will only ever get 0.1.0 and not newer versions.


nachoBonafonte (approved):

nachoBonafonte:
good catch, thanks

PR #108, opened 2021-02-08
PR for otelp-metrics

bryce-b:
I'm opening this draft PR to get some feedback on the changes I've made to the existing api to support the otelp for metrics. 

Here are key notes for these changes: 
```
- Made MeterSdkProvider init mimic the TraceProvider init

        I decided to change the MetricSdkProvider.init method to take shared state objects as parameters,
        since this is how the TraceProvider operates. Alternatively, I could have move the default values
        into the MeterSharedState construction, but that would require making the NoopMetricProcessor &
        NoopMetricExporter public, which I'm not against, but seemed intentionally private.

- replaced MeterRegisteryKey with InstrumentationLibraryInfo

- Injected Resource into the MetricSdk stack
        I followed the same pattern as with the Trace Sdk.
```


nachoBonafonte:
It looks great so far. 
About making Noop... public, I cannot imagine a use case for a user of the library to use a NoopMetricProcessor or  a NoopMetricExporter. I would only make public what can be useful for the user.

bryce-b:
>  I cannot imagine a use case for a user of the library to use a NoopMetricProcessor or a NoopMetricExporter
I agree, the issue I ran into was using them as default values for the sharedMeterState's constructor, which would have required them to be public since they would be available in the constructor definition.

bryce-b:
I'm looking for some guidance in the toProtoMetric method in Sources/Exporters/OpenTelemetryProtocol/metric/MetricsAdapter.swift

I'm not seeing a way to set min/max values. The closest proto metric I could find for the summary metrics was a histogram. I was thinking that the explicitBounds might satisfy the "min/max" issue, but I don't think that's quite right. Would exemplars be the means to capture the min max values?

nachoBonafonte:
Not an expert here, but I would say so. Checking go code , they have this [method](https://github.com/open-telemetry/opentelemetry-go/blob/e50a1c8cecb8c584bdfada6cb21fe0b173a22646/exporters/otlp/internal/transform/metric.go#L253) and in line 261 they convert to histomgram with [this](https://github.com/open-telemetry/opentelemetry-go/blob/e50a1c8cecb8c584bdfada6cb21fe0b173a22646/exporters/otlp/internal/transform/metric.go#L478). I hope it can help you

nachoBonafonte:
Sorry for the delay in reviewing, i was OOO since last Wednesday. 
The code looks really great, and I think that your changes made the usage much nicer and consistent. Just to be picky, I saw some styling issues mainly with multiple line breaks, but can be merges like this. (I use swiftformat for auto fixing style)

nachoBonafonte (approved):

nachoBonafonte:
Thanks a lot for your contribution.
Let me know if you cannot merge the PR and I will do.

bryce-b:
I'll make sure I use swiftformat in the future! I'm not able to merge however; take it away!

PR #113, opened 2021-02-25
Update to opentelemetry spec 1.0.1

nachoBonafonte:
This PR adds several minor changes to comply with opentelemetry spec v 1.0.1, including:
- Swiftify Status struct, converting it to an enum and associated description only for error status
- Renamed TraceConfig to SpanLimits, and moved Sampler out to TracerSharedState
- Renamed Probability sampler to TraceIdRatioBased
- Allow usage of ParentBased sampler (it was not possible to create it before from a user perspective)
- Update OTLP protocol to 0.7.0
- Update grpc-swift to 1.0.0 and generate swift-proto files with this version ( Fixes #96)
- Rename IdsGenerator protocol to IdGenerator

I would also like to use this version, if no issues arise with these changes, for the first Official beta version of OpenTelemetry Swift library. I would also like it to be announced publicly to try to increase usage and adoption. @SergeyKanzhelev, do you think is a good idea? How should I proceed with it? Thanks


nachoBonafonte:
Hi @SergeyKanzhelev, did you have time to review it? They are mainly name changes

SergeyKanzhelev (approved):

PR #114, opened 2021-03-11
added TraceConstant class with exception semantics

bryce-b:
I didn't see these values defined anywhere.


Sources/OpenTelemetrySdk/Trace/TraceConstants.swift
```diff
@@ -0,0 +1,35 @@
+// Copyright 2020, OpenTelemetry Authors
```
SergeyKanzhelev, line 1: already 2021. My recommendation is just omit the year mentions across all files

Sources/OpenTelemetrySdk/Trace/TraceConstants.swift
```diff
@@ -0,0 +1,35 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+public enum TraceConstants: String {
```
SergeyKanzhelev, line 18: so this is a public class on SDK? Is it assumed that libraries that instrument code will take dependency on SDK rather than API?

Sources/OpenTelemetrySdk/Trace/TraceConstants.swift
```diff
@@ -0,0 +1,35 @@
...
+    
+    /// Exception event conventions
+
+    /// An exception event must be named "exception"
+    case exception = "exception";
+    /// The type of the exception (its fully-qualified class name, if applicable). The dynamic type of the exception should be preferred over the static type in languages that support it.
+    case exceptionType = "exception.type";
+    /// SHOULD be set to true if the exception event is recorded at a point where it is known that the exception is escaping the scope of the span.
+    case exceptionEscaped = "exception.escaped";
+    /// The exception message.
```
SergeyKanzhelev, line 28: I suggest to note whether it should be localized or not (if this is possible in Swift to take a neutral - `en` version of a string).

Sources/OpenTelemetrySdk/Trace/TraceConstants.swift
```diff
@@ -0,0 +1,35 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+public enum TraceConstants: String {
+    
+    /// Exception event conventions
```
SergeyKanzhelev, line 20: is there a spec for this. May worth referring to it.

Sources/OpenTelemetrySdk/Trace/TraceConstants.swift
SergeyKanzhelev: maybe not public on not on SDK. Also, would it be better to create wrappers for errors. Is it unnetural to operate with the word "exception" on swift?

nachoBonafonte:
For generic attibutes, there is already a file in the API: `SemanticAttributes.swift`, imho all this standard attributes should be located there.
About supporting exceptions, I initially had omitted it on purpose because its status was experimental at least on the exceptions spec, but if you are really interested in supporting it, I would prefer to see the whole feature landing in the repository, including both the [record exception](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception) and the [attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md).

Sources/OpenTelemetrySdk/Trace/TraceConstants.swift
bryce-b: that's what I get for copy-pasting 😆

Sources/OpenTelemetrySdk/Trace/TraceConstants.swift
bryce-b: > so this is a public class on SDK? Is it assumed that libraries that instrument code will take dependency on SDK rather than API? 

That's correct.

> maybe not public on not on SDK. Also, would it be better to create wrappers for errors. Is it unnatural to operate with the word "exception" on swift?

I'm following the established pattern in OpenTelemetrySDK/Resources/ResourcesConstant.swift per public and formatting
These specific strings are pulled from the exception spec : https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md
These constants are intended to be used in that context, so I think leaving it using a generic term "exceptions" should be ok.

I'd like to add a recordException() to span object, but this is good first step.

bryce-b:
> For generic attibutes, there is already a file in the API: SemanticAttributes.swift, imho all this standard attributes should be located there.
Good point. Per the spec it should be in the API! I was getting thrown off by the ResourceConstants.swift file in the SDK Resource folder. Should that get moved up to the API too?

> About supporting exceptions, I initially had omitted it on purpose because its status was experimental at least on the exceptions spec, but if you are really interested in supporting it, I would prefer to see the whole feature landing in the repository, including both the record exception and the attributes.

I avoided adding an RecordException api on spans because it's experimental, but I figured having these constants isn't a bad thing. The api is essentially a convenience method for adding an event with these specific attributes, and I'm using them to record network errors. I can revise this PR to add the methods, but we might want to discuss which objects it will accept: NSException, Error, NSError? I think there is reason to support all three.

Sources/OpenTelemetrySdk/Trace/TraceConstants.swift
bryce-b: I think that depends on open-telemetry API spec. Should semantic conventions be localized? If so we should create a Issue to review all the semantic conventions to update localization.

bryce-b: I've opened a ticket #115 for this.

nachoBonafonte:
> Good point. Per the spec it should be in the API! I was getting thrown off by the ResourceConstants.swift file in the SDK Resource folder. Should that get moved up to the API too?
Resources is part of the SDK, thats why that file is locate there, but yes, its location couls also be in API.

> I avoided adding an RecordException api on spans because it's experimental, but I figured having these constants isn't a bad thing. The api is essentially a convenience method for adding an event with these specific attributes, and I'm using them to record network errors. I can revise this PR to add the methods, but we might want to discuss which objects it will accept: NSException, Error, NSError? I think there is reason to support all three.
Yes, its trus, now thinking about it  I can see a use case for the attributes without using the API. We can talk about adding the RecordException API in teh next SIG meeting, but I also think adding support for all three objects is useful, specially for usability.

nachoBonafonte (approved):

bryce-b:
just fyi, I don't have merge perms so I'll need someone else to merge the pr when it's approved.

PR #118, opened 2021-03-18
added nsurlsession instrumentation

bryce-b:
pretty messy first pass; it will need some clean up, but I wanted to get this pr up for comment.

### DIFF
--- a/Sources/Instrumentation/NSURLSession/DDURLSessionDelegate.swift
+++ b/Sources/Instrumentation/NSURLSession/DDURLSessionDelegate.swift
@@ -1,0 +1,38 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// The `URLSession` delegate object which enables network requests instrumentation. **It must be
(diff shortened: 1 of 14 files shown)


Sources/Instrumentation/NSURLSession/DDURLSessionDelegate.swift
```diff
@@ -0,0 +1,38 @@
...
+@objc
+open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate {
+    var interceptor: URLSessionInterceptorType?
+
+    @objc
+    override public init() {
+        interceptor =   URLSessionAutoInstrumentation.instance?.interceptor
+        if interceptor == nil {
+            print("""
+                `Agent.start()` must be called before initializing the `DDURLSessionDelegate` and
```
bryce-b, line 22: review messaging here

nachoBonafonte:
My main concern with this approach is the need for the user to use the provided session delegate, so users will need to modify their code in order to use this instrumentation, they may already have a session delegate that should inherit from the provided one. If application uses some library that does networking it could be impossible to instrument this third party library without modifying the code.

bryce-b:
It should be pretty straight forward to add some instrumentation to intercept the delegate when it is passed to the NSURLSession. It's probably a good idea to make a couple of options available for users, which ever they are.

nachoBonafonte:
I have added my proposal also as a PR: [here](https://github.com/open-telemetry/opentelemetry-swift/pull/119)

PR #119, opened 2021-03-23
Instrumentation for URLSession 

nachoBonafonte:
This is my proposal for URLSession instrumentation:
* It works just by initializing `URLSessionInstrumentation`, for any code or library based on URLSession, no changes for the user code or third party libraries.
* User can implement any extra behaviour by setting the callbacks in `URLSessionConfiguration`. e.g. exclude URL's for instrumentation or header injection, report extra tags, inject extra headers...

Added a very simple example that can be used for experimenting
Missing tests for the code

### DIFF
--- a/Examples/Network Sample/main.swift
+++ b/Examples/Network Sample/main.swift
@@ -1,0 +1,46 @@
+// Copyright 2020, OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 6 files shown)


nachoBonafonte:
I am adding this PR to a branch so the code can be used and improved from more people

Examples/Network Sample/main.swift
```diff
@@ -0,0 +1,46 @@
+// Copyright 2020, OpenTelemetry Authors
```
SergeyKanzhelev, line 1:
```suggestion
// Copyright 2021, OpenTelemetry Authors
```

SergeyKanzhelev (approved):

PR #135, opened 2021-03-26
Print events in StdoutExporter debug log

johnbley:
I'm using the StdoutExporter as a debug aide and found that it was missing the span events.


nachoBonafonte (approved):

nachoBonafonte:
LGTM, thanks for your contribution

PR #147, opened 2021-03-29
TracerProviderSdk must log if the name is empty

### LINKED ISSUE #139: Null/empty name passed to `TracerProvider.getTracer()` should log the error
See [this part](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer) - value remains the same, but we need to log it, as it's _technically_ an error ;)

nachoBonafonte:
TracerProviderSdk must log if the name is empty and use a default name. Fixes #139
Also make TracerProviderSdk() non public, since user should use the get() interface
Added tests for TracerProviderSdk
Also added default resources to EnvVarResources


Sources/OpenTelemetrySdk/Trace/TracerSdkProvider.swift
```diff
@@ -20,19 +20,36 @@ import OpenTelemetryApi
 public class TracerSdkProvider: TracerProvider {
     private var tracerProvider = [InstrumentationLibraryInfo: TracerSdk]()
     internal var sharedState: TracerSharedState
+    internal static let emptyName = "unknown"
 
     /// Returns a new TracerSdkProvider with default Clock, IdGenerator and Resource.
-    public init(clock: Clock = MillisClock(),
+    init(clock: Clock = MillisClock(),
```
bryce-b, line 26: I'm a little concerned about making this private. It will limit the configurability of the `TracerSdkProvider`; or is the intended way to make custom configurations subclassing the `TracerSdkProvider`?

per the spec: 
> Configuration (i.e., SpanProcessors, IdGenerator, SpanLimits and Sampler) MUST be managed solely by the TracerProvider and it MUST provide some way to configure all of them that are implemented in the SDK, at least when creating or initializing it.

nachoBonafonte: Yes, you are true, I was following exactly that spec, understanding that configuration should come from the TracerProvider in the API, but we need configuration in the SDK side, since some of those settigns are on the Sdk. I will revert this line

bryce-b (approved):

bryce-b (approved):

SergeyKanzhelev (approved):

nachoBonafonte:
TracerProviderSdk must log if the name is empty and use a default name. Fixes #139
Added tests for TracerProviderSdk
Also added default resources to EnvVarResources

PR #156, opened 2021-04-04
Update Trace/Resource Attribute Constants

Sherlouk:
~Closes #116. Slight clash with #144.~

Superseded by #157.

PR #157, opened 2021-04-04
Automate Semantic Attributes Generation

### LINKED ISSUE #116: Update Opentelemetry Constants
Check:
SemanticConventions.swift
Resource constants

Sherlouk:
Hey folks 👋 

I created #156 manually this morning which synchronises our resource constants and trace attributes with the latest OpenTelemetry specification.

I also discovered (after doing that) that as part of the specification there exists YML files which detail the attributes in a machine readable way, and indeed [build tools](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions) exist to automate the generation.

This PR adds a new generate script which will create nicely documented attributes, easily synced to a specific version of the spec (currently 1.1.0 but this can be changed).

**:warning: This might be a breaking change.** 
Of the attributes that have tests for them, and of the ones I manually checked, nothing has changed - but I'm sure there will be at least one somewhere which has changed casing ever so slightly. I think it's best to take the hit early on as it'll make it easy to upgrade in the future.

For transparency, the generate script was taken from the Java repository [here](https://github.com/open-telemetry/opentelemetry-java/blob/main/buildscripts/semantic-convention/generate.sh) and the Jinja template is inspired from [theirs](https://github.com/open-telemetry/opentelemetry-java/blob/main/buildscripts/semantic-convention/templates/SemanticAttributes.java.j2) too.

Supersedes #156. Closes #116. Slight clash with #144.


Sherlouk:
I've also gone through and generated the default member sets which the specification outlines for certain key/value pairs. I use either a struct or an enum dependant on whether or not the specification supports custom values. There's no nice API for it currently but this could certainly be expanded in the future.

Sources/OpenTelemetryApi/Trace/SemanticAttributes.swift
```diff
@@ -12,141 +12,1494 @@
...
+    */
+    public struct FaasInvokedProviderValues: CustomStringConvertible {
+        /**
+        Amazon Web Services.
+        */
+        static let AWS = FaasInvokedProviderValues("aws")
+        /**
+        Amazon Web Services.
+        */
+        static let Azure = FaasInvokedProviderValues("azure")
```
Sherlouk, line 1400: Mistake in the specification - PR here: https://github.com/open-telemetry/opentelemetry-specification/pull/1595

nachoBonafonte (approved):
Great contribution, it both adds documentation and saves time in the future for any update. Thanks a lot

PR #158, opened 2021-04-06
Implement a Jaeger propagator for context and baggage

### LINKED ISSUE #153: Implement a Jaeger propagator
As per the spec, a Jaeger Propagator must be implemented (for propagating both span and baggage values).

nachoBonafonte:
Closes #153


SergeyKanzhelev (approved):
this will increase the API surface and size. But I think having it is beneficial for users. So lgtm

PR #159, opened 2021-04-06
Add Locks to SwiftMetricsShim

Sherlouk:
Also change access control on `metricName` setters to private to minimise leakage outside of the class.

Hopefully addresses the concerns of @ktoso in their code review on #155.

### DIFF
--- a/Sources/SwiftMetricsShim/MetricHandlers.swift
+++ b/Sources/SwiftMetricsShim/MetricHandlers.swift
@@ -18,7 +18,7 @@ import OpenTelemetryApi
 
 class SwiftCounterMetric: CounterHandler, SwiftMetric {
     
-    var metricName: String
+    private(set) var metricName: String
     var metricType: MetricType = .counter
     let counter: AnyCounterMetric<Int>
     let labels: [String: String]
@@ -41,7 +41,7 @@ class SwiftCounterMetric: CounterHandler, SwiftMetric {
 
 class SwiftGaugeMetric: RecorderHandler, SwiftMetric {
     
-    var metricName: String
+    private(set) var metricName: String
     var metricType: MetricType = .gauge
     let counter: AnyCounterMetric<Double>
     let labels: [String: String]
@@ -64,7 +64,7 @@ class SwiftGaugeMetric: RecorderHandler, SwiftMetric {
 
 class SwiftHistogramMetric: RecorderHandler, SwiftMetric {
     
-    var metricName: String
+    private(set) var metricName: String
     var metricType: MetricType = .histogram
     let measure: AnyMeasureMetric<Double>
     let labels: [String: String]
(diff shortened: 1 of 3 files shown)


nachoBonafonte (approved):
LGTM, I will give some time for @ktoso to review before merging

Sources/SwiftMetricsShim/MetricHandlers.swift
```diff
@@ -87,7 +87,7 @@ class SwiftHistogramMetric: RecorderHandler, SwiftMetric {
 
 class SwiftSummaryMetric: TimerHandler, SwiftMetric {
     
-    var metricName: String
+    private(set) var metricName: String
```
ktoso, line 90: Do those ever need to be set though? 🤔 

I don't see any other modifications of them other than during initialization, am I missing something?

ktoso (approved):
Looks good :) 

Might want to add a test that creates a bunch of metrics concurrently? :-)

ktoso:
Thanks for the ping! More than happy to review 👍

Sources/SwiftMetricsShim/MetricHandlers.swift
Sherlouk: I checked setting them to let - can't be done due to protocol conformance 😞

Sherlouk:
Thanks for taking a look Konrad 🙂

Sources/SwiftMetricsShim/MetricHandlers.swift
ktoso: Hm? which conformance?

I only see:

```
protocol SwiftMetric {
    var metricName: String { get }
    var metricType: MetricType { get }
    init(name: String, labels: [String: String], meter: Meter)
}
```

Which are get-only properties so they can be witnessed by `let metricName` and `let metricType`.

I even checked locally and yea, seems just fine, do you have something else in mind?

Sherlouk: I've written Swift for half a decade and I've never realised that a `let variableName` is valid for a protocol which defines `var variableName` - today I learned!

What I was referring to is that I tried changing the `var` within the protocol definition itself to a `let` and that doesn't work. What I didn't realise did work is changing the `var` to a `let` in each of the classes but leaving the protocol unaffected. I had naively assumed that it would flag up an inconsistency between class and conformance.

Apologies for being so dense, great shout and has now been changed.

Thanks for educating me!

ktoso: Ha, okey :-) Lovely, happy to help.

Yes it's a neat pattern to implement get only requirements with a `let`, very happy to have spread the knowlage about this a bit more then 👍

ktoso:
Looking good, now with the `let` even better :-) Thanks for the ping

Issue #141, opened 2021-03-26
Consider simplifying `Baggage` section


carlosalberto:
For a little while we had the same discussion in Java, and we ended up removing `BaggageManager`, offering all functionality through `Baggage.builder()` and module/static functions for `Context` handling (similar to moving them out of `Tracer`).

We also happened to hide other members as `EmptyBaggage`, by providing the same through static methods (i.e. `Baggage.empty()`).

Finally - we removed `DefaultBaggage` and made this the **official** de facto class used underneath.

(Overall I _think_ with all these changes, we made the API surface smaller/simpler).

alolita:
Fixed. Closing issue.

Issue #38, opened 2020-06-04
Linux Support & NIO Integration


pauljohanneskraft:
Hello - first of all, I want to say, that I really like your efforts towards providing tracing capabilities in Swift! Thank you for your efforts towards providing a Swift implementation for OpenTelemetry.

Since I would like to use OpenTelemetry-Swift on Linux, I tried to compile the current version for a Ubuntu-based Docker container, but due to dependencies to `os.activity` this unfortunately failed. Since most servers run on Linux distributions and/or Linux-based Docker containers, adding Linux support would potentially open up OpenTelemetry-Swift to a lot more users.

To remove the dependency on `os.activity`, I dug a little deeper into this library and noticed that the dependency on `os.activity` is present in `ContextUtils`, `CorrelationContextInScope` and `SpanInScope`. Especially `ContextUtils` seems to be quite hard to substitute since it provides static functions with attributes depending on `os.activity`.

As far as I understood the APIs being used by this library, it enters and leaves different execution contexts which are used for tracing. Is this actually required to be handled by `os.activity` or could it easily be replaced with a different API? I thought about the possible usage of [NIO](https://github.com/apple/swift-nio) where a new context could be opened whenever a context is switched from one eventLoop to the next. I am, however, not deeply familiar with the inner workings of Swift-NIO, which is why this approach might not be feasible.

Please let me know what you think about adding Linux support to OpenTelemetry-Swift and whether or not it might make sense to use NIO for automatic context propagation. It might also be a temporary workaround to disable certain functionality (e.g. any automatic context propagation mechanism) on Linux to open up the library to a lot more users.

nacho-undefinedlabs:
Hi, yes, one of the goals of the project is that it can be build and used on Linux, and as you said I think that the only current  limitation is in os.activity library that only works on Apple platform ( and, as you can see from the implementation, even when the library comes from Apple it is not well supported from Swift and some low levels stuff is needed to use it).

This library uses system services that allows to get and set a context from every point in code without the need of carrying it with you, automatically keeping the status in threads and asynchronous tasks.

This library could be replaced with other API that provided the same functionality,  I don't know how NIO manages contexts, but if it can keep it without user intervention in code that doesn't know about this library it would be a great option. Or maybe it could use a different implementation on Linux using some system services that Linux can provide for it (I have no Linux experience), or as you said temporary removing  automatic context propagation on Linux or even using some kind of stack based context propagation that would partially solve it in many cases.

[...]

PR #160, opened 2021-04-08
Reworked context handling in the library

### LINKED ISSUE #126: Consider hiding `Span.scope()`
Exposing the [actual Scope](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift#L80) can become trick, starting with the fact that a `Span` can be passed among different threads and end up being associated to many `Scopes`. 

Besides that, probably it makes sense to be a little conservative and hide something like this (it could be exposed later on if truly needed).

### LINKED ISSUE #154: Context layer should not reference directly trace/baggage portions
As discussed in the previous SIG call, the `Context` layer that exists on top of `Activity` should not directly reference any other signals (trace, baggage, metrics), in order to keep it independent, and allow the user to, if desired, extend it himself by storing/propagating custom values.

nachoBonafonte:
* Add initial support for implementing a different contextProvider, by default uses a contextProvider based on Activity.framework
* Removed Scope concept from the library
* Now context is handled in OpenTelemetry.instance.contextProvider, which uses the desired ContextProvider
This rework also implied modifiying the default Baggage, whihch is now an optional set to nil instead of an EmptyBaggage instance, these empty classes are not public anymore.
Some more cleaning can be done in Baggage handling, but the PR is big enough

Fixes #126, Fixes #154, partially improves #141
Some changes ease development for #38


Sources/OpenTelemetryApi/Baggage/BaggageBuilder.swift
```diff
@@ -43,6 +43,6 @@ public protocol BaggageBuilder: AnyObject {
     /// - Parameter key: the EntryKey which will be removed.
     @discardableResult func remove(key: EntryKey) -> Self
 
-    /// Creates a Baggage from this builder.
-    func build() -> Baggage
+    /// Creates a Baggage from this builder or nil if it has no entries itself or the parent
```
SergeyKanzhelev, line 46: please rephrase. This sentence is hard to understand

Sources/OpenTelemetryApi/Baggage/BaggageBuilder.swift
```diff
@@ -43,6 +43,6 @@ public protocol BaggageBuilder: AnyObject {
...
+    func build() -> Baggage?
```
SergeyKanzhelev, line 47: returning nil may increase possibility for errors in using the library. We do it in .NET anyway. In Java we always return empty objects instead to minimize NRE. Raising comment to review if this is indeed what is wanted.

SergeyKanzhelev (commented):
also fully commented file may be removed from the repo completely

bryce-b (approved):

Sources/OpenTelemetryApi/Baggage/BaggageBuilder.swift
nachoBonafonte: Yes, you are true. I will remove the change,it was trying to explain the implementation of the `DefaultBaggageBuilder`, and this is the protocol. Just made no sense

Sources/OpenTelemetryApi/Baggage/BaggageBuilder.swift
nachoBonafonte: I was returning nil to avoid forcing to return an EmptyBaggage class,  which I thought I had to export. But after thinking more about it and checking that I can still the empty class internal, I am reverting this change.

nachoBonafonte:
friendly ping @SergeyKanzhelev

SergeyKanzhelev (approved):

PR #165, opened 2021-04-12
Added Resource Extension to generate iOS attributes

bryce-b:
this provides a number of resource generating objects that captures various iOS related attributes
for use with TracerProviders.

I'm open to ideas of different names for the resource extension package, as well as where it lives in the project structure. I placed them where I did because I'm not very creative.


Tests/InstrumentationTests/sdk-resource-extension-tests/ApplicationResourceProviderTests.swift
```diff
@@ -0,0 +1,80 @@
...
+#if canImport(UIKit)
+    class ApplicationResourceProviderTests: XCTestCase {
+        func testContents() {
+            let appData = mockApplicationData(name: "appName", identifier: "com.bundle.id", version: "1.2.3", build: "9876")
+
+            let provider = ApplicationResourceProvider(source: appData) as ResourceProvider
+
+            let resource = provider.create()
+
+            XCTAssertTrue(appData.name == resource.attributes[ResourceAttributes.serviceName.rawValue]?.description)
```
Sherlouk, line 30: nit: should we use `XCTAssertEqual` in order to get better failure messages?

question: Are there any OTel guidelines with regards to the service name? I know internally we have a fairly strict format (so we would personally need a route to override this) but for example we don't support spaces or 'special characters' (basically non [a-z\-_]) in them. I'm not sure if there is anything in the spec for this.

nachoBonafonte (commented):
I would like Opentelemetry-swift supports also tvOS and macOS. Would it be possible to add support also for tvOS and macOS values?

bryce-b:
@nachoBonafonte I'll look into that

Tests/InstrumentationTests/sdk-resource-extension-tests/ApplicationResourceProviderTests.swift
nachoBonafonte: About the question you make, I think that it supports any AttributeValue, so the supported characters in the strings are not limited

nachoBonafonte (changes requested):
Just some minor issues:
* Please update the README now that not only UIKIt is supported and add it to the exclude list in the package, like in other exporters (it fixes a warning)
* Now is the only target whose paths are in small caps and words separated by `-`, instead of capitals. Would you mind  changing it (not mandatory though)

nachoBonafonte (approved):

nachoBonafonte:
Thanks a lot, merging now

PR #166, opened 2021-04-13
added optional URLSessionInstrumentation config to customize span name

bryce-b:
I've added an optional config to allow customization of the span name in the URLSession instrumentation which is allowed per the spec : 
```Therefore, HTTP client spans SHOULD be using conservative, low cardinality names formed from the available parameters of an HTTP request, such as "HTTP {METHOD_NAME}". Instrumentation MUST NOT default to using URI path as span name, but MAY provide hooks to allow custom logic to override the default span name.``` 

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md

also added some tests to verify the new addition operates as expected
some of the other changes are results of running swiftformat

### DIFF
--- a/Tests/InstrumentationTests/URLSessionTests/URLSessionConfigurationTests.swift
+++ b/Tests/InstrumentationTests/URLSessionTests/URLSessionConfigurationTests.swift
@@ -50,8 +50,10 @@ final class URLSessionConfigurationTests: XCTestCase {
         XCTAssertEqual(true, checker.nameSpanCalled)
 
         XCTAssertEqual(1, URLSessionLogger.runningSpans.count)
-        let span = URLSessionLogger.runningSpans["id"]!
-        XCTAssertEqual("new name", span.name)
+        XCTAssertNotNil(URLSessionLogger.runningSpans["id"])
+        if let span = URLSessionLogger.runningSpans["id"] {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Tests/InstrumentationTests/URLSessionTests/URLSessionConfigurationTests.swift [hunk 1]
```diff
@@ -0,0 +1,106 @@
...
+            receivedResponse: nil,
+            receivedError: nil))
+
+        URLSessionLogger.processAndLogRequest(request, sessionTaskId: "id", instrumentation: instrumentation, shouldInjectHeaders: true)
+
+        XCTAssertEqual(true, checker.nameSpanCalled)
+
+        XCTAssertEqual(1, URLSessionLogger.runningSpans.count)
+        let span = URLSessionLogger.runningSpans["id"]!
+        XCTAssertEqual("new name", span.name)
```
nachoBonafonte, line 54: How about avoiding the force unwrap and move the optional to the assert, we avoid a crash in case it fails

nachoBonafonte (approved):
Looks good to me, and thanks for adding the first tests to the network instrumentarion

nachoBonafonte:
I don't know if you can merge yourself, as it's not the main branch. Let me know if not.

PR #168, opened 2021-04-19
Improve usability of MetricProvider

### LINKED ISSUE #167: Add `addMetricProcessor` function to MetricProvider
https://cloud-native.slack.com/archives/C01NCHR19SB/p1618582139006900?thread_ts=1618580592.006800&cid=C01NCHR19SB

We need to add consistency between the span provider and metric provider. Currently for metrics you have to initialise your own `MeterProviderSdk` which is not necessary for spans.

Desired syntax:
> OpenTelemetrySDK.instance.meterProvider.addMetricProcessor(metricProcessor)

nachoBonafonte:
Add some consistency between the span provider and metric provider, so the metric provider can be configured without needing to use the constructor.
Also add support for multiple metric exporters
Fixes #167


nachoBonafonte:
I know this API is highly experimental, but usability was really bad without it

Sherlouk (approved):
Got a super busy week ahead of me so may not be able to test it immediately in our app but the code itself looks great.

Sources/OpenTelemetrySdk/Metrics/Export/MetricExporter.swift
```diff
@@ -19,12 +19,48 @@ public enum MetricExporterResultCode {
     case success
     case failureNotRetryable
     case failureRetryable
+
+    /// Merges the current result code with other result code
+    /// - Parameter newResultCode: the result code to merge with
+    public mutating func mergeResultCode(newResultCode: MetricExporterResultCode) {
```
SergeyKanzhelev, line 25: does it have to be public?

SergeyKanzhelev (approved):

Sources/OpenTelemetrySdk/Metrics/Export/MetricExporter.swift
nachoBonafonte: No, it does not. Changed this and also the one in SpanExporterResultCode.  Thanks a lot.

PR #169, opened 2021-04-19
move URLSession Config `requestCreated` call to before the span is cr…

bryce-b:
…eated

The config describes this callback as a means to customize the span builder before
the span is started, but it was getting called after `spanBuilder.startSpan()` was called.

Though, now, the request hasn't been "created" since `createdRequest` is called before `instrumentedRequest`.  Maybe this configuration should be called "customizeSpanBuilder"?

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionLogger.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionLogger.swift
@@ -61,6 +61,8 @@ class URLSessionLogger {
             spanBuilder.setAttribute(key: $0.key, value: $0.value)
         }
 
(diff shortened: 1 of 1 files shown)


nachoBonafonte:
Yes, you are true. This is an interesting problem, because it would be great to get the request with the tracing headers, so you can log that value if interested, but we cannot get the headers until the span is started.

nachoBonafonte:
I think I prefer the callback passes the created span instead of the span builder, what do you think?

nachoBonafonte:
I have moved on with replacing SpanBuilder to Span in the callback, sometimes during instrumentation we replace the original request, so I think is better if the callback receive the real request sent instead of a possibly old one.

PR #170, opened 2021-04-21
Allow configuration on the TraceProviderSDK

nachoBonafonte:
Allow configuration on the TraceProviderSDK, after hiding the SharedState from public access, configuration of some fields was not possible anymore.

### DIFF
--- a/Sources/OpenTelemetrySdk/Trace/TracerProviderSdk.swift
+++ b/Sources/OpenTelemetrySdk/Trace/TracerProviderSdk.swift
@@ -16,7 +16,6 @@
 import Foundation
 import OpenTelemetryApi
 
-/// This class is not intended to be used in application code and it is used only by OpenTelemetry.
 public class TracerProviderSdk: TracerProvider {
     private var tracerProvider = [InstrumentationLibraryInfo: TracerSdk]()
     internal var sharedState: TracerSharedState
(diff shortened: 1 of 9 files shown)


Sources/OpenTelemetrySdk/Trace/TracerProviderSdk.swift
```diff
@@ -66,16 +65,61 @@ public class TracerProviderSdk: TracerProvider {
...
+        return sharedState.sampler
+    }
+
+    /// Updates the active Sampler.
+    public func updateActiveSampler(_ newSampler: Sampler) {
+        sharedState.setSampler(newSampler)
+    }
+
+    /// Returns the active SpanProcessor.
+    public func getActiveSpanProcessor() -> SpanProcessor {
```
SergeyKanzhelev, line 119: what does "active" span processor means? There may be many configured, correct? And there is not corresponding `updateActiveSpanProcessor` either

nachoBonafonte: Yes you are true, the name is not well chosen and maybe the functionality is not needed. I added it for having access to the span processors configured. In the case of having several span processors it would return the MultiSpanProcessor that wraps them. Maybe I should return the array of span processors included in the multispanprocessor instead and change the name. What do you think?

nachoBonafonte: I have reworked those methods, now they are: 
    `getActiveSpanProcessors()` returns an array with all the configured Span Processors
    `addSpanProcessor(_ spanProcessor: SpanProcessor)` add a new Span processor to the already configured 
   `resetSpanProcessors()` it removes all configured SpanProcessors and replace them with a Noop one

SergeyKanzhelev (approved):

PR #171, opened 2021-04-21
Add automatic Network instrumentation for URLSession

nachoBonafonte:
Automatically instruments URLSession just by initializing the class.
Extra behaviour can be added by setting the callbacks in URLSessionInstrumentationConfiguration.
It includes a simple example with the minimum code that can also be used for testing.

The code has been tested independently and with external code, and in last Swift SIG meeting we decided it was ready for merging.


SergeyKanzhelev:
I wonder if anybody with more swift knowledge can review first. If not happy to review to the best of my understanding

nachoBonafonte:
Maybe @bryce-b or @Sherlouk can provide a code review.

Examples/Network Sample/main.swift
```diff
@@ -0,0 +1,75 @@
...
+}
+
+
+class SessionDelegate: NSObject, URLSessionDataDelegate, URLSessionTaskDelegate {
+    let semaphore = DispatchSemaphore(value: 0)
+
+//    func urlSession(_ session: URLSession, task: URLSessionTask, didFinishCollecting metrics: URLSessionTaskMetrics) {
+//
+//        let a = 0
+//    }
```
Sherlouk, line 46: nit: Could probably be cleaned up

Examples/Network Sample/main.swift
```diff
@@ -0,0 +1,75 @@
...
+
+func simpleNetworkCall() {
+    let url = URL(string: "http://httpbin.org/get")!
+    let request = URLRequest(url: url)
+    let semaphore = DispatchSemaphore(value: 0)
+
+    let task = URLSession.shared.dataTask(with: request) { data, _, _ in
+        if let data = data {
+            let string = String(data: data, encoding: .utf8)
+            print(string ?? "")
```
Sherlouk, line 30: nit: We can remove the optionality by doing `String(decoding: data, as: UTF8.self)` which will output a non-optional string.

Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -0,0 +1,518 @@
...
+private var idKey: Void?
+
+public class URLSessionInstrumentation {
+    private var requestMap = [String: NetworkRequestState]()
+
+    var configuration: URLSessionInstrumentationConfiguration
+
+    private let queue = DispatchQueue(label: "com.datadoghq.ddnetworkinstrumentation")
+
+    static var instrumentedKey = "com.datadoghq.instrumentedCall"
```
Sherlouk, line 34: question: Is there any need for our URL instrumentation to include references to DataDog? While not wrong, I feel like this could lead to confusion down the line?

Sherlouk (commented):
Quick glance - looks good on the whole for a first stab, open to iteration 👍

Examples/Network Sample/main.swift
nachoBonafonte: Sure, forgot after doing some tests.

Examples/Network Sample/main.swift
nachoBonafonte: Nice one, didn't know about it. I also replaced some other uses of it.

Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
nachoBonafonte: Absolutely not, I just forgot to rename them after moving this code from the implementation I was using previously.

SergeyKanzhelev (approved):
lgtm, but I'd suggest more reviews from people more familiar with it

nachoBonafonte:
Thanks a lot. I will bring this status to the SIG meeting today, to check if anyone would like to review it more. If not I will merge.

bryce-b:
@SergeyKanzhelev @nachoBonafonte I've been using this instrumentation quite extensively and have also been digging around it it quite a bit. 👍 from me to merge.

bryce-b (approved):

PR #172, opened 2021-04-28
The Resource created here should be overwritten with new data

bryce-b:
the current implementation behaves by preserving the default values.

e.g.: `unknown_service:{process.executable.name}` is preserved in `service.name` as opposed to the new value presented in the application resource provider.


nachoBonafonte (changes requested):
Please add a test that would have found this bug

nachoBonafonte (approved):

PR #180, opened 2021-05-15
Typo fix: ipv5 used for ipv6 in the ZipkinExporter

sodabrew:
This file originated whole-cloth in #47 with this typo in place. I searched around to see if there might be some quirk that ipv5 is used by convention in Zipkin-related code but didn't find anything else. Assuming this is simply a typo, here's a simple fix. Thank you!


nachoBonafonte (approved):
Good catch, thanks a lot for reporting and creating a PR.

PR #185, opened 2021-05-20
Add new SignPostIntegration instrumentation

### LINKED ISSUE #179: Integrate with os.signpost functionality
It would be great integrating Opentelemetry with system signpost functionality. It would allow reusing all the opentelemetry instrumentation when doing profiling for example.
Using a dedicated SpanProcessor that simply calls `os_signpost_interval_begin` and `os_signpost_interval_end` when starting or ending the span would be great.

nachoBonafonte:
It integrateds spans created in opentelemetry with os_signpost functionality, allowing the spans to appear automatically in `Instruments` when profiling the app:

Screenshot 2021-05-20 at 15 10 33

Closes #179


Sources/Instrumentation/SignPostIntegration/SignPostIntegration.swift
```diff
@@ -0,0 +1,43 @@
+// Copyright 2020, OpenTelemetry Authors
```
SergeyKanzhelev, line 1: let's remove the year everywhere =)

Sources/Instrumentation/SignPostIntegration/SignPostIntegration.swift
```diff
@@ -0,0 +1,43 @@
...
+@available(macOS 10.14, *)
+public class SignPostIntegration: SpanProcessor {
+    public let isStartRequired = true
+    public let isEndRequired = true
+    public let osLog = OSLog(subsystem: "OpenTelemetry", category: .pointsOfInterest)
+
+    public init() {}
+
+    public func onStart(parentContext: SpanContext?, span: ReadableSpan) {
+        let signpostID = OSSignpostID(log: osLog, object: self)
```
SergeyKanzhelev, line 32: is this static data that can be added to the resource?

Sources/Instrumentation/SignPostIntegration/SignPostIntegration.swift
nachoBonafonte: I will replace with the format I found in opentelemetry-java:
```
/*
 * Copyright The OpenTelemetry Authors
 * SPDX-License-Identifier: Apache-2.0
 */
```
If approved I will create another PR changing to this header in all project files

Sources/Instrumentation/SignPostIntegration/SignPostIntegration.swift
nachoBonafonte: No, it creates a different ID each time it is called

SergeyKanzhelev (approved):

PR #194, opened 2021-05-28
Fix Jaeger exporter

nachoBonafonte:
Jaeger exporter was broken since commit c0dfb7df52b80552ecfc4bc9061d491f8990c552, after a time calculus change it ended up using milliseconds instead of microseconds.  Replace variables names in tests so it is obvious that the microseconds is the unit, so it cannot get broken again.

Also fix status, since it was reporting error when .undefined, that is the default now

Also modify another test on Datadog exporter that fails randomly on local to not fail anymore

### DIFF
--- a/Tests/ExportersTests/Jaeger/AdapterTests.swift
+++ b/Tests/ExportersTests/Jaeger/AdapterTests.swift
@@ -19,11 +19,11 @@ class AdapterTests: XCTestCase {
     static let parentSpanId = "0000000000aef789"
 
     func testProtoSpans() {
-        let duration = 900 // ms
-        let startMs = UInt64(Date().timeIntervalSince1970 * 1000)
-        let endMs = startMs + UInt64(duration)
+        let duration = 900 // microseconds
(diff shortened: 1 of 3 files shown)


Tests/ExportersTests/Jaeger/AdapterTests.swift
```diff
@@ -245,12 +245,12 @@ class AdapterTests: XCTestCase {
                         instrumentationLibraryInfo: InstrumentationLibraryInfo(),
                         name: "GET /api/endpoint",
                         kind: .server,
-                        startTime: Date(timeIntervalSince1970: Double(startMs) / 1000),
+                        startTime: Date(timeIntervalSince1970: Double(startMicroseconds) / 1000000),
```
SergeyKanzhelev, line 248: maybe make `1000000` a constant

SergeyKanzhelev (approved):

Tests/ExportersTests/Jaeger/AdapterTests.swift
nachoBonafonte: yes, much better. Thanks

PR #199, opened 2021-06-07
Bryce/metric processor improvements

### LINKED ISSUE #198: MeterProviderSdk -> MetricPushController retain loop
The MeterProviderSdk has a [member object](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/OpenTelemetrySdk/Metrics/MeterProviderSdk.swift#L33), pushMetricController, which in turn takes a [reference])https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/OpenTelemetrySdk/Metrics/Export/PushMetricController.swift#L9) to the MeterProviderSdk object. This relationship isn't flagged as `unowned` in the PushMetricController, which results in a retention loop, between the two objects, meaning they will never be released. 

This can be easily solved by adding `unowned var metricProvider: MeterProviderSdk` in the member variable declaration in the PushMetricController class referenced above. However, this will cause the async dispatch task to crash when the PushMetricController is released, because there isn't any `wait` that occurs in the `deinit` of that class. 

This can be fixed by using a DispatchSourceTimer to which provides `cancel` functionality to the dispatch thread: https://github.com/open-telemetry/opentelemetry-swift/pull/199

bryce-b:
This PR fixes a retention loop between MetricProviderSdk & PushMetricController, and also fixes the resulting crash by using a DispatchSourceTimer to clean up the dispatched thread when the PushMetricController is released.

fixes #198

### DIFF
--- a/Sources/OpenTelemetrySdk/Metrics/Export/PushMetricController.swift
+++ b/Sources/OpenTelemetrySdk/Metrics/Export/PushMetricController.swift
@@ -26,7 +26,7 @@ class PushMetricController {
                     return
                 }
                 let start = Date()
-                let values = self.meterProvider.getMeters().values
+                let values = meterProvider.getMeters().values
                 values.forEach {
                     $0.collect()
                 }
--- a/Sources/OpenTelemetrySdk/Metrics/Export/PushMetricController.swift
+++ b/Sources/OpenTelemetrySdk/Metrics/Export/PushMetricController.swift
@@ -9,10 +9,9 @@ class PushMetricController {
     unowned var meterProvider: MeterProviderSdk
 
     let pushMetricQueue = DispatchQueue(label: "org.opentelemetry.PushMetricController.pushMetricQueue")
-    let metricPushTimer : DispatchSourceTimer
+    let metricPushTimer: DispatchSourceTimer
 
     init(meterProvider: MeterProviderSdk, meterSharedState: MeterSharedState, shouldCancel: (() -> Bool)? = nil) {
-
         self.meterProvider = meterProvider
         self.meterSharedState = meterSharedState
         metricPushTimer = DispatchSource.makeTimerSource(flags: DispatchSource.TimerFlags(), queue: pushMetricQueue)
@@ -21,7 +20,7 @@ class PushMetricController {
                 guard let self = self else {
                     return
                 }
-                if(shouldCancel?() ?? false) {
+                if shouldCancel?() ?? false {
                     self.metricPushTimer.cancel()
                     return
                 }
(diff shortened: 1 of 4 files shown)


Sources/OpenTelemetrySdk/Metrics/Export/PushMetricController.swift
```diff
@@ -6,35 +6,45 @@ import Foundation
...
+            autoreleasepool {
+                guard let self = self else {
+                    return
+                }
+                if shouldCancel?() ?? false {
+                    self.metricPushTimer.cancel()
+                    return
+                }
+                let start = Date()
+                let values = self.meterProvider.getMeters().values
```
nachoBonafonte, line 28:
```suggestion
                let values = meterProvider.getMeters().values
```

Sources/OpenTelemetrySdk/Metrics/Export/PushMetricController.swift
```diff
@@ -6,35 +6,45 @@ import Foundation
...
-                autoreleasepool {
-                    let start = Date()
-                    let values = self.meterProvider.getMeters().values
-                    values.forEach {
-                        $0.collect()
-                    }
+        metricPushTimer = DispatchSource.makeTimerSource(flags: DispatchSource.TimerFlags(), queue: pushMetricQueue)
+        metricPushTimer.setEventHandler { [weak self] in
+            autoreleasepool {
+                guard let self = self else {
```
nachoBonafonte, line 20:
```suggestion
                guard let self = self,
                      let meterProvider = self.meterProvider else {
```
Here we retain meterProvider by storing in a local variable so we can read all values without potential deletion

Sources/OpenTelemetrySdk/Metrics/Export/PushMetricController.swift
```diff
@@ -6,35 +6,45 @@ import Foundation
 
 class PushMetricController {
     var meterSharedState: MeterSharedState
-    var meterProvider: MeterProviderSdk
+    unowned var meterProvider: MeterProviderSdk
```
nachoBonafonte, line 9:
```suggestion
    weak var meterProvider: MeterProviderSdk?
```
Being unowned meterProvider could be destroyed while reading the values and the app would crash, being `weak` allows us to retain the instance in the timer event

nachoBonafonte (changes requested):
Hi Bryce, your changes look very good, and the approach using the timer is very good. I only can see an issue with a potential crash when accessing the unowned variable. I have added suggestions so that the problem wont happen, it implies using a weak var instead of unowned and retaining in the timer event. What do you think?

bryce-b:
Looks good @nachoBonafonte ! Thanks for corrections!

nachoBonafonte (approved):

PR #202, opened 2021-06-10
Open otlp exporter

bryce-b:


nachoBonafonte:
I have creacted #207 to solve the issue you wanted to solve with it, let me know what you think

PR #203, opened 2021-06-15
Add aggregationTemporality to OTEL metrics exporter

Sherlouk:
Not 100% sure if `cumulative` is the correct answer here, but it does work for my use case. Without this, these metrics do not pass the OTEL collector's exporter validation (tested forwarding events to Prometheus) as it complains about the invalid temporality setting.

With this fix, we're seeing events successfully go through the collector and land in Prom.

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocol/metric/MetricsAdapter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocol/metric/MetricsAdapter.swift
@@ -100,6 +100,7 @@ struct MetricsAdapter {
                     protoDataPoint.labels.append(kvp)
                 }
 
+                protoMetric.doubleSum.aggregationTemporality = .cumulative
                 protoMetric.doubleSum.dataPoints.append(protoDataPoint)
             case .doubleSummary:
 
@@ -136,6 +137,7 @@ struct MetricsAdapter {
                     protoDataPoint.labels.append(kvp)
                 }
 
+                protoMetric.intSum.aggregationTemporality = .cumulative
                 protoMetric.intSum.dataPoints.append(protoDataPoint)
             case .intSummary:
                 guard let summaryData = $0 as? SummaryData<Int> else {


Sherlouk:
Another error is being flagged up by Prometheus for "out of order timestamps". I'm not sure if it's caused by this, but I've also noticed that we have an issue somewhere which is duplicating metrics.

This is a sample log where I'm dumping the metric name and value in my own custom StdOut metric exporter. Notice how metrics appear multiple times (for example the play count) which might be throwing Prometheus off if it's seeing multiple data entries.

```
videoErrorCount 1
videoPlayCount 4
videoLoadTime Optional(2.335054874420166)
videoErrorCount 0
videoPlayCount 1
videoWatchPercentage Optional(2.000297263537837)
videoBufferPercentage Optional(0.0)
videoLoadTime Optional(0.4247908592224121)
```

nachoBonafonte:
LGTM, but lets wait if anyone has other opinion. Would you prefer this change merged, or also add the fix the other issue as part of this PR

Sherlouk:
I'm honestly not too sure what the fix is for the second issue.

I've patched it temporarily by wrapping the OTEL metric exporter in my own metric exporter, and filtering out any duplicates (only leaving the most recent values). This shouldn't be necessary though and means the project is using up more and more memory as time goes on, which is clearly not good. I think we need to cleanup the metrics once they've been pushed up successfully - any ideas?

This is what I'm using (which appears to work fine):
```swift
class MultiplexMetricExporter: MetricExporter {
    let underlying: MetricExporter
    
    init(underlying: MetricExporter) {
        self.underlying = underlying
    }
    
    func export(metrics: [Metric], shouldCancel: (() -> Bool)?) -> MetricExporterResultCode {
        var mutable = [Metric]()
        
        metrics.reversed().forEach { metric in
            if mutable.contains(where: { $0.name == metric.name }) == false {
                mutable.append(metric)
            }
        }
        
        return underlying.export(metrics: mutable, shouldCancel: shouldCancel)
    }
}
```

nachoBonafonte:
Ok, I am going to merge this fix and have moved the second issue to #205

nachoBonafonte (approved):

PR #206, opened 2021-06-17
Reset collected metrics properly in MetricProcessorSdk

### LINKED ISSUE #205: Metrics are being exported duplicated
From a report in #203 
```
Another error is being flagged up by Prometheus for "out of order timestamps". I'm not sure if it's caused by this, but I've also noticed that we have an issue somewhere which is duplicating metrics.

This is a sample log where I'm dumping the metric name and value in my own custom StdOut metric exporter. Notice how metrics appear multiple times (for example the play count) which might be throwing Prometheus off if it's seeing multiple data entries.

videoErrorCount 1
videoPlayCount 4
videoLoadTime Optional(2.335054874420166)
videoErrorCount 0
videoPlayCount 1
videoWatchPercentage Optional(2.000297263537837)
videoBufferPercentage Optional(0.0)
videoLoadTime Optional(0.4247908592224121)
```

nachoBonafonte:
Reset collected metrics in `MetricProcessorSdk.finishCollectionCycle()`
Fixes #205
Remove `UngroupedBatcher` metric processor, it works the same than `MetricProcessorSDK` but not concurrently safe.
Use `MetricProcessorSDK` everywhere the other was used.


Sherlouk:
Looks good - worth noting this is now technically a breaking change for clients.

nachoBonafonte:
I have reverted deleting UngroupedBatcher, but has marked it as deprecated so it will warn users if the change but wont break any code

bryce-b (approved):

Issue #204, opened 2021-06-15
Expose `CallOptions` in OTLP exporters


bryce-b:
The `TraceServiceClient` & `MetricServiceClient` (full class name like : `Opentelemetry_Proto_Collector_Trace_V1_TraceServiceClient`) are used to export service requests using GRPC via an `export` method ( [example 1](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Exporters/OpenTelemetryProtocol/trace/OtlpTraceExporter.swift#L32))
This `export` method as an optional second parameter, `CallOptions` : https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Exporters/OpenTelemetryProtocol/proto/trace_service.grpc.swift#L57

This `CallOptions` provides access to setting [GPRC metadata](https://github.com/grpc/grpc-swift/blob/main/Sources/GRPC/CallOptions.swift) which can be useful for use in setting things like AUTH headers.

I'd like to propose exposing an optional configuration to these exporters to allow passing of a `CallOptions` to the constructor. Maybe something like : 

```
public class OtlpTraceExporter: SpanExporter {
    let channel: GRPCChannel   
    let traceClient: Opentelemetry_Proto_Collector_Trace_V1_TraceServiceClient    
    let timeoutNanos: Int64
    let options: CallOptions?

    public init(channel: GRPCChannel, timeoutNanos: Int64 = 0, options: CallOptions? = nil) {
       ...
   }

    public func export(spans: [SpanData]) -> SpanExporterResultCode {
        ...
        let export = traceClient.export(exportRequest, options)
```

nachoBonafonte:
Totally agree, in fact it looks like a really needed feature

bryce-b:
This actually may be solved by implemented https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md

specifically the OTEL_EXPORTER_OTLP_HEADERS field.

PR #213, opened 2021-06-22
Add functionality for Exporter headers env-var #204

bryce-b:


bryce-b:
I did some research on env-vars in Xcode after a hunch I got looking at the way env-vars are set in Xcode (only possible through the `run` option). Env-vars will only apply to app state ran though xcode, and wont be transmitted to archive builds. It seems obvious after the fact. I think that we may need another method of configuration akin to Env-var config available in other sdks. 
@nachoBonafonte do you think we should allow these configurations to be set up through plists?

Sources/OpenTelemetrySdk/Trace/TracerProviderBuilder.swift
```diff
@@ -0,0 +1,56 @@
...
+        spanProcessors.append(spanProcessor)
+        return self
+    }
+
+    public func add(spanProcessors: [SpanProcessor]) -> Self {
+        self.spanProcessors.append(contentsOf: spanProcessors)
+        return self
+    }
+
+    public func build() -> TracerProvider {
```
nachoBonafonte, line 53: we could return here TracerProviderSdk

nachoBonafonte, line 53:
```suggestion
    public func build() -> TracerProviderSdk {
```
we could return here TracerProviderSdk

nachoBonafonte (approved):
It looks really good to me

PR #217, opened 2021-07-16
added NetworkStatus package & injected it into network instrumentation

bryce-b:
implements attributes defined here : https://github.com/open-telemetry/opentelemetry-specification/pull/1759

I think I need to re-generate the protobuf spec as well?


bryce-b:
Seems like there is a permissions issue using another github project. Will it be an issue to use a 3rd party framework here?

nachoBonafonte:
I did an initial review it looks generally good, some minor things only mostly stylistic.

nachoBonafonte:
> Seems like there is a permissions issue using another github project. Will it be an issue to use a 3rd party framework here?

What do you mean by permissions issue, license related? While the third party project has a compatible license there should be no problem.

bryce-b:
>  What do you mean by permissions issue, license related? While the third party project has a compatible license there should be no problem.

The build is failing due to unable to access the `reachability` repo

nachoBonafonte:
> >  What do you mean by permissions issue, license related? While the third party project has a compatible license there should be no problem.
> 
> 
> 
> The build is failing due to unable to access the `reachability` repo
> 
> 

I think the problem is in the URL used in Package.swift for the repo, can you try using: `https://github.com/ashleymills/Reachability.swift` ?

Tests/InstrumentationTests/NetworkStatusTests/NetworkStatusTests.swift
```diff
@@ -0,0 +1,152 @@
+// Copyright © 2021 Elasticsearch BV
```
SergeyKanzhelev, line 1: is the different copyright intentional?

bryce-b: Oops! usually I catch this. I'll correct it right away.

nachoBonafonte (approved):

PR #219, opened 2021-08-11
Added initializer for NetworkStatusInjector

bryce-b:
Not sure if we want the logger in there, I'm not sure we log errors like this anywhere else in the framework. Let me know if I should take it out.


nachoBonafonte (approved):
LGTM, about using the system logger I have no problems with it, we can talk about it in the next meeting and decide a clear path for the future

PR #227, opened 2021-09-02
Datadog exporter: migrate to intake v2

nachoBonafonte:
Updated the endpoints and the format.
It support both API key and client keys in the same configuration parameter
Added support for compressing the payloads


SergeyKanzhelev (approved):
just glanced over, lgtm

Sources/Exporters/DatadogExporter/ExporterConfiguration.swift
```diff
@@ -37,15 +38,15 @@ public struct ExporterConfiguration {
     /// Option to add a custom prefix to all the metrics sent by the exporter
     var metricsNamePrefix: String?
 
-    public init(serviceName: String, resource: String, applicationName: String, applicationVersion: String, environment: String, clientToken: String?, apiKey: String?, endpoint: Endpoint, uploadCondition: @escaping () -> Bool, performancePreset: PerformancePreset = .default, exportUnsampledSpans: Bool = true, exportUnsampledLogs: Bool = true, hostName: String? = nil, metricsNamePrefix: String? = "otel") {
+    public init(serviceName: String, resource: String, applicationName: String, applicationVersion: String, environment: String, apiKey: String, endpoint: Endpoint, source: String = "otel-swift-exporter", uploadCondition: @escaping () -> Bool, performancePreset: PerformancePreset = .default, exportUnsampledSpans: Bool = true, exportUnsampledLogs: Bool = true, hostName: String? = nil, metricsNamePrefix: String? = "otel") {
```
ncreated, line 41: Wasn't `source` set to `"ios"` before? 
I'm not exactly sure about the impact of using `"otel-swift-exporter"` 💭 - I think this won't install default "iOS" processing pipelines (in Logs) when the first data comes in.

Sources/Exporters/DatadogExporter/Logs/LogsExporter.swift
```diff
@@ -39,7 +35,7 @@ internal class LogsExporter {
             dateProvider: SystemDateProvider()
         )
 
-        let dataFormat = DataFormat(prefix: "", suffix: "", separator: "\n")
+        let dataFormat = DataFormat(prefix: "[", suffix: "]", separator: ",")
```
ncreated, line 38: 👍

Sources/Exporters/DatadogExporter/Upload/DataUploadStatus.swift
```diff
@@ -0,0 +1,86 @@
...
+    let needsRetry: Bool
+
+    // MARK: - Debug Info
+
+    /// Upload status description printed to the console if SDK `.debug` verbosity is enabled.
+    let userDebugDescription: String
+
+    /// An optional error printed to the console if SDK `.error` (or lower) verbosity is enabled.
+    /// It is meant to indicate user action that must be taken to fix the upload issue (e.g. if the client token is invalid, it needs to be fixed).
+    let userErrorMessage: String?
```
ncreated, line 63: I don't see if we make use of these two in `DataUploadWorker`. In `dd-sdk-ios` we print the `userDebugDescription` to the user (on upload failure) and send `userErrorMessage` for internal monitoring. I think we can remove it from here and cleanup.

ncreated (approved):
Looks good 👍, I left few clean-up comments.

Sources/Exporters/DatadogExporter/ExporterConfiguration.swift
nachoBonafonte: Oh, I was not aware of that. I will change it back

Sources/Exporters/DatadogExporter/Upload/RequestBuilder.swift
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        self.url = urlComponents?.url ?? url
+        self.precomputedHeaders = precomputedHeaders
+        self.computedHeaders = computedHeaders
+    }
+
+    /// Creates `URLRequest` for uploading given `data` to Datadog.
+    /// - Parameter data: data to be uploaded
+    /// - Returns: the `URLRequest` object and `DD-REQUEST-ID` header value (for debugging).
```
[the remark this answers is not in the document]
nachoBonafonte: Removed

Sources/Exporters/DatadogExporter/Upload/DataUploadStatus.swift
nachoBonafonte: Cleaned up, thanks for your review

PR #228, opened 2021-09-02
OTLP Exporter example to send traces and metrics

vvydier:
This example uses the OTLP Exporter to send traces and metrics to otel-collector which is configured to export traces to zipkin and metrics to prometheus respectively


Examples/OTLP Exporter/main.swift
```diff
@@ -0,0 +1,97 @@
...
+    parentSpan.setAttribute(key: sampleKey, value: sampleValue)
+    for _ in 1...10 {
+        doWork(parentSpan: parentSpan)
+    }
+    Thread.sleep(forTimeInterval: 0.5)
+    parentSpan.end()
+}
+
+func doWork(parentSpan: Span) {
+    OpenTelemetry.instance.contextProvider.setActiveSpan(parentSpan)
```
nachoBonafonte, line 58: This line should moved to the `createSpan()` function after parent span has been created (can be called before or after the `setAttribute()` method), since the active span is stored between function calls. Calling it here repeatedly is superfluous. The `parentSpan` parameter can be omitted then

nachoBonafonte (approved):
It looks good to me, followed the README and everything works as expected.

PR #232, opened 2021-09-15
Bring zipkin exporter's treatment of status into spec

johnbley:
Use proper names (otel.status_code, error) and criteria for when/how they should be added as zipkin tags.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/zipkin.md#status


nachoBonafonte (approved):
LGTM, thanks a lot for your contribution

PR #235, opened 2021-09-16
Fix thread race condition in URLSessionInstrumentation

nachoBonafonte:
Cover all accesses to requestMap using the proper queue


bryce-b (approved):
pretty straight forward

PR #240, opened 2021-09-21
Update swift-atomics to 1.0.0

johnbley:
Pro: swift-atomics went 1.0 about a month ago; the final api changes were fairly minimal and don't affect our code base; and it does bring good benefits like support for library evolution.  

Con: Some of our (otel-swift's) users may also depend on swift-atomics. Unfortunately, Swift Package Manager doesn't have a way to express that otel-swift would be happy with "0.x OR 1.x".  So, there is a small risk of user impact where somebody would be forced into upgrading from swift-atomics 0.0.x to 1.0 (which, again, should be very easy to do).

Balancing everything I think it's OK to move now, but I can also understand wanting to wait a while.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/apple/swift-atomics.git",
         "state": {
           "branch": null,
-          "revision": "26e346cd64f6b92d4089e7fafe2f8e82f60ddb8c",
-          "version": "0.0.2"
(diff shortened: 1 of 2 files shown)


bryce-b:
This looks good to me

nachoBonafonte (approved):
It looks good to me also, and as you say, the risk on user impact is really low.

PR #242, opened 2021-09-30
reverse merge on env-var resource fixes #237

### LINKED ISSUE #237: service.name not set from environment variable
When I run the OTLP Exporter example, the spans that are generated do not seem to pick up service.name from the env variable OTEL_SERVICE_NAME as per
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md
I also OTEL_RESOURCE_ATTRIBUTES="service.name=TEST" which also did not work and as per the spec link above, OTEL_SERVICE_NAME should take precedence over OTEL_RESOURCE_ATTRIBUTES

bryce-b:
It seems like the env-var resource was using merge backwards, where the env-vars should take precedence.


nachoBonafonte (approved):

nachoBonafonte:
It looks like EasyCLA got stuck, could you force a change?

PR #255, opened 2021-10-21
fix env variables as per changes in the otel spec

vvydier:
1. Changed the env variable to OTEL_RESOURCE_ATTRIBUTES as per the spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable
2. Changing the environment variables to TRACEPARENT and TRACESTATE for inter-process propagation https://github.com/open-telemetry/opentelemetry-specification/issues/74


bryce-b:
the link for `TRACEPARENT` & `TRACESTATE` goes to a git ignore update pr. can you link the intended spec doc? I was having issues finding it

PR #270, opened 2021-11-03
Fix URLSessionInstrumentation concurrent network requests

nachoBonafonte:
URLSessionInstrumentation was not reporting concurrent network requests

The task identifier was being set out of the callback implementation so it was being reused, if the session tried to send concurrent requests, only the first one was being reported.


danielqsj (approved):
/lgtm

SergeyKanzhelev (approved):

PR #273, opened 2021-11-04
Implement ForceFlush spec adding a timeout parameter.

### LINKED ISSUE #186: Implement ForceFlush spec in built-in Span processors
As specified here:
[https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush-1](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush-1)

Missing implementing a timeout in the protocol, and return if success. Do not break existing API, and add deprecation notice.

Also Exporters `forceFlush` are missing timeout parameter

nachoBonafonte:
This parameter takes effect only in BatchSpanProcessor built-in processors

Fixes #186


Sources/OpenTelemetrySdk/Trace/SpanProcessor.swift
```diff
@@ -33,5 +33,12 @@ public protocol SpanProcessor {
 
     /// Processes all span events that have not yet been processed.
     /// This method is executed synchronously on the calling thread
-    func forceFlush()
+    func forceFlush(timeout: TimeInterval?)
```
SergeyKanzhelev, line 36: should the new argument be documented? Specifically, end users will appreciate some information on what does zero timeout means and how to make it infinite

Sources/OpenTelemetrySdk/Trace/SpanProcessors/BatchSpanProcessor.swift
```diff
@@ -109,33 +109,33 @@ private class BatchWorker: Thread {
...
-    private func exportBatch(spanList: [ReadableSpan]) {
+    private func exportBatch(spanList: [ReadableSpan], explicitTimeout: TimeInterval?) {
         let exportOperation = BlockOperation { [weak self] in
             self?.exportAction(spanList: spanList)
         }
         let timeoutTimer = DispatchSource.makeTimerSource(queue: DispatchQueue.global())
-        timeoutTimer.setEventHandler{ exportOperation.cancel() }
-        timeoutTimer.schedule(deadline: .now() + .milliseconds(Int(exportTimeout.toMilliseconds)), leeway: .milliseconds(1))
+        timeoutTimer.setEventHandler { exportOperation.cancel() }
+        timeoutTimer.schedule(deadline: .now() + .milliseconds(Int(explicitTimeout?.toMilliseconds ?? exportTimeout.toMilliseconds)), leeway: .milliseconds(1))
```
SergeyKanzhelev, line 138: is timeout that is higher than exportTimoue make sense?

Sources/OpenTelemetrySdk/Trace/SpanProcessor.swift
nachoBonafonte: Of course it should, added

Sources/OpenTelemetrySdk/Trace/SpanProcessors/BatchSpanProcessor.swift
nachoBonafonte: Yes, it really doesn't make much sense, it will use the lower of them

PR #276, opened 2021-11-09
Update Resource Values with Semantic Attributes

### LINKED ISSUE #274: Implement new resource values
The latest resource spec has some new attributes we'll need to capture. Enumerated here: 
https://github.com/open-telemetry/opentelemetry-swift/pull/272/files#diff-134eba6eef75f0deb4895f6658ee2f4e83d08921a0483fa14bb92a6cc84049ddR250

Sherlouk:
* Added extra parameters to resource attributes, as per the updated spec.
* Fixed bug where enum values were not public, and thus could not be used by clients.
* Addressed misuse of OS "type" which, as per spec, MUST be of type "darwin" for our clients.
* Used ResourceAttributes enum values, where possible, to avoid future bugs.
* Replaced use of hostId with the now correct deviceId.

Closes #274.

### DIFF
--- a/Sources/OpenTelemetrySdk/Resources/ResourceAttributes.swift
+++ b/Sources/OpenTelemetrySdk/Resources/ResourceAttributes.swift
@@ -936,19 +936,19 @@ public enum ResourceAttributes: String {
         /**
         Alibaba Cloud.
         */
-        static let alibabaCloud = CloudProviderValues("alibaba_cloud")
+        public static let alibabaCloud = CloudProviderValues("alibaba_cloud")
         /**
         Amazon Web Services.
         */
-        static let aws = CloudProviderValues("aws")
+        public static let aws = CloudProviderValues("aws")
         /**
         Microsoft Azure.
         */
-        static let azure = CloudProviderValues("azure")
+        public static let azure = CloudProviderValues("azure")
         /**
         Google Cloud Platform.
         */
-        static let gcp = CloudProviderValues("gcp")
+        public static let gcp = CloudProviderValues("gcp")
 
         internal let value: String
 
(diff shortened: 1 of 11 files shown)


Sources/OpenTelemetrySdk/Resources/ResourceAttributes.swift
```diff
@@ -1170,48 +1170,47 @@ public enum ResourceAttributes: String {
...
-        static let webjs = TelemetrySdkLanguageValues("webjs")
+        public static let webjs = TelemetrySdkLanguageValues("webjs")
         /**
-         swift.
-         */
-        static let swift = TelemetrySdkLanguageValues("swift")
-
+        swift.
+        */
+        public static let swift = TelemetrySdkLanguageValues("swift")
```
Sherlouk, line 1213: It's worth noting that this is something that we have added to the attributes file ourselves. If you regenerate, it will go.

Now that I'm using this value within the resource attributes, at least our tests will fail though so we should be reminded to add it back manually.

Sherlouk (commented):
This **could** be perceived as a breaking change as we are no longer sending some values which users may be relying on. It's very nuanced though.

For example, we now send `device.model.id` instead of `device.model`. Or how `os.type` has changed from "iOS" to "darwin" as per the specification.

I don't think we need to be that worried right now, but interested what others think.

Sources/OpenTelemetrySdk/Resources/ResourceAttributes.swift
nachoBonafonte: Yes, I also added it to the spec and it has been merged since then, so when the spec version is updated the value will be automatically generated

nachoBonafonte (approved):

nachoBonafonte:

PR #277, opened 2021-11-09
Improve SemanticConvention generation template

Sherlouk:
Based off of #276, will need to change the base once that work is complete. 
**Please just review the second commit for now.**

At the moment, if you run the `generate` script, it will make changes to the project which will cause compiler or test failures.

The three failures which this protects against:
* The value "int" being used as an enum value. This needs to be "Int" instead.
* The keyword "internal" being used as a case value. This should be wrapped with backticks like `internal`.
* The Swift telemetry language which we have chosen to add, outside of the specification.

While not technically a failure, this change will also revert the documentation changes made in a previous PR which changed all the `Int` values to "int" incorrectly. This fixes that too.

### DIFF
--- a/Scripts/semantic-convention/templates/SemanticAttributes.swift.j2
+++ b/Scripts/semantic-convention/templates/SemanticAttributes.swift.j2
@@ -13,6 +13,8 @@
     Bool
   {%- elif type == "number" -%}
     Int
+  {%- elif type == "int" -%}
+    Int
   {%- else -%}
     {{type}}
   {%- endif -%}
@@ -84,6 +86,11 @@ public enum {{enum}}: String {
         */
         public static let {{ member.member_id | to_camelcase }} = {{class_name}}({{ print_value(type, member.value) }})
         {%- endfor %}
+        {%- if 'TelemetrySdkLanguage' in class_name %}
+        /**
+        swift.
+        */
+        public static let swift = TelemetrySdkLanguageValues("swift"){% endif %}
 
         internal let value: String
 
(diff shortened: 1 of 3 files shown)


Sherlouk:
Added a second commit here which adds better support for enum values. Instead of just saying the type should be String, it says which enum value should be used. It will even add a clickable link to the enum - so users can more easily see the available options.

In theory, we could also list them in the inline docs - but lets iterate and see how it feels.

Scripts/semantic-convention/templates/SemanticAttributes.swift.j2
```diff
@@ -76,14 +83,21 @@ public enum {{enum}}: String {
     /**
     {{attribute.brief | to_doc_brief}}.
     */
+    /// - Tag: otel{{class_name}}
```
Sherlouk, line 86: I have absolutely no idea why we must use triple slash syntax for tags, but everything else works with `/**`.

Tags do not work unless you use triple slash syntax.
I have verified this functionality, both work side by side.

nachoBonafonte:
Everything looks really great, I only have some doubt with this:
> The Swift telemetry language which we have chosen to add, outside of the specification.

The value was added to the spec in this [PR](https://github.com/open-telemetry/opentelemetry-specification/pull/2097). So it should be included in the next version of the spec. We will probably only use the script for updating when a newer version appear. What do you think?

Sherlouk:
Noted, will remove the Swift workaround and will rebase onto main.

Sherlouk:
Should all be good now 👍

nachoBonafonte (approved):
Yes, it looks perfect now. Thanks a lot

Sherlouk:
Thanks Nacho!

PR #279, opened 2021-11-23
Make Metric and SpanData conform to Codable to support persisting them.

shelefg:
This submission is making `Metric` and `SpanData` conform to `Codable` in order to allow them to be persisted. Being able to persist signals' data is required to support use cases where signals may collected when there is no network connectivity, and so they have to be exported later - possibly after the process was terminated and relaunched - when network connectivity is back.

The core of the change is making `Metric` and `SpanData` conform to the `Codable` protocol. To that end all the types `Metric` and `SpanData` depend on were made `Codable` as well. 

Most of the types could simply have the `Codable` protocol implementation synthesized. However there were several cases where an explicit implementation was required:

1. `Metric` - explicit encoding & decoding logic was required to deduce the concrete type of `Metric.data` by `Metric.aggregationType`
2. `AttributeValue` - enum with associated values
3. `Status` - enum with associated values
4. `HistogramData` - has a tuple type member
5. Conflict with existing Encodable implementation used specifically in StdoutExporter. In this case an explicit Encodable implementation was written for `SpanExporterData`.

Tests were added to cover the encoding & decoding code, and corresponding Equatable conformance for `Metric` that was required for tests verification was added.

### DIFF
--- a/Sources/OpenTelemetryApi/Trace/Status.swift
+++ b/Sources/OpenTelemetryApi/Trace/Status.swift
@@ -60,3 +60,58 @@ extension Status: CustomStringConvertible {
         }
     }
 }
+
+// this explicit Codable implementation for Status will probably be redundant with Swift 5.5
+extension Status: Codable {
+    enum CodingKeys: String, CodingKey {
+        case ok
+        case unset
+        case error
+    }
+
+    enum EmptyCodingKeys: CodingKey {
+
+    }
+
+    enum ErrorCodingKeys: String, CodingKey {
+        case description
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        guard container.allKeys.count == 1 else {
+            let context = DecodingError.Context(
+                codingPath: container.codingPath,
+                debugDescription: "Invalid number of keys found, expected one.")
+            throw DecodingError.typeMismatch(Status.self, context)
+        }
+
+        switch container.allKeys.first.unsafelyUnwrapped {
+        case .ok:
+            _ = try container.nestedContainer(keyedBy: EmptyCodingKeys.self, forKey: .ok)
+            self = .ok
+        case .unset:
+            _ = try container.nestedContainer(keyedBy: EmptyCodingKeys.self, forKey: .unset)
+            self = .unset
+        case .error:
+            let nestedContainer = try container.nestedContainer(keyedBy: ErrorCodingKeys.self, forKey: .error)
+            self = .error(description: try nestedContainer.decode(String.self, forKey: .description))
+        }
+    }
+
+    public func encode(to encoder: Encoder) throws {
+
+        var container = encoder.container(keyedBy: CodingKeys.self)
+
+        switch self {
+        case .ok:
+            _ = container.nestedContainer(keyedBy: EmptyCodingKeys.self, forKey: .ok)
+        case .unset:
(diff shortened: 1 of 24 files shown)


Sources/OpenTelemetryApi/Trace/Status.swift
```diff
@@ -60,3 +60,58 @@ extension Status: CustomStringConvertible {
         }
     }
 }
+
+// this explicit Codable implementation for Status will probably be redundant with Swift 5.5
```
nachoBonafonte, line 64: I was thinking on keeping the default implementation for swift 5.5 with something like:

```
#if swift(>=5.5)
extension Status: Codable { }
#else 
```
and your implementation in the else, what do you think?

nachoBonafonte (commented):
I would try using the default implementation when possible for swift 5.5 (mainly for performance and maintenance reasons in the future). For the rest everything looks good to me

Sources/OpenTelemetryApi/Trace/Status.swift
shelefg: I tried this and it works. However, the encoding is different between my explicit encoding and the synthesized encoding. For example, for `AttributeValue`, the synthesized `.string("")` value will be `{"string":""}` with the explicit encoding and `{"string":{"_0": ""}}` with the synthesized encoding. 

This can cause an issue if an app has some signal data persisted pre Swift 5.5, and then it gets updated to a version built post Swift 5.5, and tries to decode the data.

I'm debating what to do:

1. Make the explicit encoding match the synthesized encoding (described [here](https://github.com/apple/swift-evolution/blob/main/proposals/0295-codable-synthesis-for-enums-with-associated-values.md))
2. Keep only the explicit encoding.
3. Make Swift 5.5 the minimal version (currently it's 5.2) and keep only the synthesized encoding.

What would you suggest?

nachoBonafonte: Yes, it was my initial concern when I proposed it, but after thinking about it I realized that using Codable does not mean that the output format is directly the codable output, it is just the in-memory representation. The real output format is the one that the user decides with the decoder implemented like JSONEncoder or NSKeyedArchiver

shelefg: My current implementation of the persistence decorator (part of the next submission) already uses JSONEncoder (it's not a user decision). That will result in differences such as the one I gave as an example. 

The persistence decorator will write the encoded data to a file (similar to how the DataDogExporter does it - it is basically copied from it), and it is possible that the data will be read from the file after an app upgrade which may include an update to Swift >= 5.5.

nachoBonafonte: Oh, sorry for my confusion. 
Then yes, there will be issues. 
Option 3 is not an option, we should not break existing version just for it.  So the options are 1, 2 or consider the breaking a minimal issue. I would prefer option 1, which means easier maintenance once written for future changes to the classes, but don’t know if the complexity would be too high. What is your preference?

shelefg: Agreed. Option 1 is pretty straight forward to implement. I pushed the changes which include tests for forward compatibility of a pre swift 5.5 encoder with a post swift 5.5 decoder.

nachoBonafonte: Thats really great, thanks a lot for the changes.

nachoBonafonte (approved):
Everything LGTM now, thanks a lot for your contribution

PR #280, opened 2021-11-24
Introduce persistence exporter decorators for MetricExporter and Span…

shelefg:
This submission builds upon #279 , and introduces a persistence decorator for signal exporters.

Most of the code in the submission was imported and adjusted from the `DataDogExporter` library. It contains:

* A persistence layer (see /Storage).
* The export worker (see /Export).
* The decorators & persistence and export configuration.

For example, decorating metric and span exporters with persistence functionality can be done as follows:

```
let metricExporter = ... // create some MetricExporter
let persistenceMetricExporter = try PersistenceMetricExporterDecorator(
                                            metricExporter: metricExporter,
                                            storageURL: metricsSubdirectoryURL,
                                            writerQueue: DispatchQueue(label: "metricWriterQueue"),
                                            readerQueue: DispatchQueue(label: "metricReaderQueue"),
                                            exportQueue: DispatchQueue(label: "metricExportQueue"),
                                            exportCondition: { return true })

let spanExporter = ... // create some SpanExporter
let persistenceTraceExporter = try PersistenceSpanExporterDecorator(
                                            spanExporter: spanExporter,
                                            storageURL: tracesSubdirectoryURL,
                                            writerQueue: DispatchQueue(label: "spanWriterQueue"),
                                            readerQueue: DispatchQueue(label: "spanWriterQueue"),
                                            exportQueue: DispatchQueue(label: "spanWriterQueue"),
                                            exportCondition: { return true })
```

The `PersistenceMetricExporterDecorator` and `PersistenceSpanExporterDecorator` will asynchronously:
1. Encode the exported `Metric`'s and `SpanData`'s objects to JSON.
2. Write the data to files in the folder specified by `storageURL` 
3. Read back these objects from the disk and forward them to be exported by the corresponding decorated exporters.


nachoBonafonte (changes requested):
Congratulations, It looks really good, the PR includes a lot of code but everything is clear and well commented.  Thanks for taking your time in making high quality code.
I am only missing a pair of things: 
- A README file at top level explaining the usage (basically the text you added to github when created the PR)
- some very simple tests for PersistenceMetricExporterDecorator and PersistenceSpanExporterDecorator. Because they are the user API and we should test that it does not break without notice.

Except for that, it is a merge for me

Sources/Exporters/Persistence/PersistenceSpanExporterDecorator.swift
```diff
@@ -65,6 +65,7 @@ public class PersistenceSpanExporterDecorator: SpanExporter {
     }
     
     public func shutdown() {
+        persistenceExporter.flush()
```
nachoBonafonte, line 68: Good catch

shelefg:
@nachoBonafonte thanks. I added a README and added a minimal happy path test for each of the decorators. In contrast to most of the other tests in the package, these tests are "end-to-end", meaning they don't mock the internal logic of the persistence exporter - they actually write the data to the disk, read it, and forward it.

I'm not sure that this meets your expectations in terms of coverage. Again, I took the DataDogExporter tests as a reference.

As a side note, I noticed something while working on the CR fixes - it looks like the FileReader both in the PersistenceExporter and in the DataDogExporter does not actually use the DispatchQueue it is given. As far as I can tell, reading is executed as part of uploading / exporting using the upload / export DispatchQueue. Am I missing something?

nachoBonafonte (approved):
LGTM

nachoBonafonte:
> @nachoBonafonte thanks. I added a README and added a minimal happy path test for each of the decorators. In contrast to most of the other tests in the package, these tests are "end-to-end", meaning they don't mock the internal logic of the persistence exporter - they actually write the data to the disk, read it, and forward it.
> 
> I'm not sure that this meets your expectations in terms of coverage. In this case (as with everything else) I used the DataDogExporter tests as a reference.

I was just reviewing it while you wrote the message. Yes, this meets my expectations, just uses the API as a user would, so it validates it does not break.

> As a side note, I noticed something while working on the CR fixes - it looks like the FileReader both in the PersistenceExporter and in the DataDogExporter does not actually use the DispatchQueue it is given. As far as I can tell, reading is executed as part of uploading / exporting using the upload / export DispatchQueue. Am I missing something?

Ok, I will review, but will create an issue to track it independently, so we dont keep this work on hold and we can merge.

Again, thanks a lot for your contribution

PR #282, opened 2021-11-29
Fix startTime of span.

nachoBonafonte:
SpanBuilderSDK's startTime was being initialised with current system time, this was forcing this value to be used as the startTime when the RecordEventsReadableSpan was initialised. This was causing two big issues:
- The start time was set when the spanBuilder was created, not when the span was really started
- If the user decided to use an external clock, it would only be used for the end time, not for the start time. Which would make the duration to be always wrong

### DIFF
--- a/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
+++ b/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
@@ -44,10 +44,9 @@ public class RecordEventsReadableSpan: ReadableSpan {
     public private(set) var clock: Clock
     /// The resource associated with this span.
     public private(set) var resource: Resource
-    /// The start time of the span.
     /// instrumentation library of the named tracer which created this span
     public private(set) var instrumentationLibraryInfo: InstrumentationLibraryInfo
-    /// The resource associated with this span.
+    /// The start time of the span.
     public private(set) var startTime: Date
     /// Set of recorded attributes. DO NOT CALL any other method that changes the ordering of events.
     private var attributes: AttributesDictionary
@@ -132,7 +131,7 @@ public class RecordEventsReadableSpan: ReadableSpan {
     ///   - attributes: the attributes set during span creation.
     ///   - links: the links set during span creation, may be truncated.
     ///   - totalRecordedLinks: the total number of links set (including dropped links).
-    ///   - startTime: the time for the new span.
+    ///   - startTime: the time for the new span, if not set it will use assigned Clock time
     public static func startSpan(context: SpanContext,
                                  name: String,
                                  instrumentationLibraryInfo: InstrumentationLibraryInfo,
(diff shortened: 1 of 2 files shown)


Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
```diff
@@ -146,7 +146,7 @@ public class RecordEventsReadableSpan: ReadableSpan {
                                  attributes: AttributesDictionary,
                                  links: [SpanData.Link],
                                  totalRecordedLinks: Int,
-                                 startTime: Date) -> RecordEventsReadableSpan
```
SergeyKanzhelev, line 149: should the comment for the `startTime` be updated to indicate the behavior when not set?

Sources/OpenTelemetrySdk/Trace/SpanBuilderSdk.swift
```diff
@@ -33,7 +33,7 @@ class SpanBuilderSdk: SpanBuilder {
 
     private var startAsActive: Bool = false
 
-    private var startTime = Date()
```
SergeyKanzhelev, line 36: where is this initialization happening now? This PR only removes initialization, not adding another one instead?

bryce-b: It looks like the constructor of `RecordEventsReadableSpan` had already taken an optional `startTime` in construction, but this behavior wasn't propagated. By making it optional in the `SpanBuilderSdk` it now is behaving as expected.  https://github.com/open-telemetry/opentelemetry-swift/pull/282/files#diff-2b0cc00ea3cbfbf944d9c31aa9fe1760f6414f213bc8bcc64a5850d6652a366fL112

nachoBonafonte: Optional types (in this case the Date? is an optional Date type) are always nil if not initialised. The value will only be set if the user explicitly calls setStartTime(). If The builder starts the span with the startTime uninitialised, the span will use the Clock assigned to it instead of this value

bryce-b: Line 112 of `RecordEventsReadableSpan`

SergeyKanzhelev: thank you for clarification!

Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
nachoBonafonte: Yes, its true. I tried to clarify it

Sources/OpenTelemetrySdk/Trace/SpanBuilderSdk.swift
nachoBonafonte: Yes, this value should only be used if the user explicitly wants to fix the start time, by default the span should use the time that the assigned Clock has

SergeyKanzhelev (approved):

PR #284, opened 2021-12-07
Implement a Stack in the TaskLocal for automatic context handling

nachoBonafonte:
Use a stack in the TaskLocal so we can push and pop context when running in async/await related code.
Added support to running async code also in Xcode 13.2 (still in beta)
Added tests for different context passing scenarios, some of them will only run in Xcode 13.2
Added a name to PropagatedSpan so we can find issues easier when running code


bryce-b:
I see there are some #ifdefs for swift 5.5.2, but the beta is only saying 5.5.0 is installed. Is there another way to install 5.5.2?

nachoBonafonte:
Xcode 13.2 beta installs swift 5.5.2, doesn’t it?

SergeyKanzhelev (approved):
mostly glanced thru, looks good.

bryce-b:
If I change the expected swift version in the package file it throws an error if I try to set 5.5.2 saying only 5.5.0 is installed (on beta). Maybe my expectation on how this works is incorrect?

nachoBonafonte:
Package is probably using swift-tools version, which can be a bit different than swift language version. Can you run the tests marked with 5.5.2 using the beta?

bryce-b:
Doesn't seem like the breakpoint will hit within the 5.5.2 ifdef using the beta. Let me try one of the beta toolchains

bryce-b:
AH, turns out there is a difference between `<` & `>`. I was misreading what was happening. Everything looks good.

bryce-b (approved):

Issue #281, opened 2021-11-25
Check DispatchQueue usage in DatadogExporter and PersistenceExporter


nachoBonafonte:
From #280 :
>it looks like the FileReader both in the PersistenceExporter and in the DataDogExporter does not actually use the DispatchQueue it is given. As far as I can tell, reading is executed as part of uploading / exporting using the upload / export DispatchQueue. Am I missing something?

creedasaurus:
Looks like this may be resolved with #288, right?

I did have a comment/question or two on code introduced in #288 however. 

1. In the Persistence exporter (non-Datadog specific), the [DataExportWorker's DispatchQueue label](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Exporters/Persistence/Export/DataExportWorker.swift#L21) is "com.otel.datadog...". Should it be something like "com.otel.persistence.dataExportWorker"?

2. Is there a reason to use the `.global(qos: .userInteractive)` DispatchQueue as the target for these FileWriters -> [DataDog](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Exporters/DatadogExporter/Persistence/FileWriter.swift#L16) and [Generic](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Exporters/Persistence/Storage/FileWriter.swift#L20)? From my limited understanding, `.userInteractive` doesn't necessarily mean it's executed on the main thread, but should be executed with the highest priority. My initial thought would be that persisting metrics/traces to disk isn't something a user would be tracking directly, so a lower qos would be more fitting. But I'm also learning more about that as I go. Is there a good reason to use that priority for writing to the files?

nachoBonafonte:
Hi Creed, 

Yes, this one should have been marked as solved with #288. I will close it after this message.

1. Yes, this is an oversight from my side, the similarities of code produced this copy/paste error. I will create a PR to fix it
2. Yes, there is a reason, and is about priority not about the thread it will run. If this code runs on an iOS device, apps are usually sent to the background and are killed by the system while in the background when the system needs the resources. If the code was run in a background queue, it would probably not be executed at all and the spans or logs would never be exported. Sending the exported data to the backend can be done with a lower priority queue, because the information is already stored in the device and wont be lost.

nachoBonafonte:
Closed by #288

PR #288, opened 2022-01-10
Simplifies Persistence and Datadog exporter queue logic

nachoBonafonte:
Simplifies Queue logic, and removes FileRead queue that was not being used.
Upload and Writer queues are internally created now, so user dont need to create them.
It makes Persistence exporter simpler to configure providing some default values.
Updated README for Persistence exporter.
Fix some formatting


bryce-b (approved):

nachoBonafonte:
Closes #281

PR #292, opened 2022-02-03
Metric refresh

bryce-b:
Initial pass at stable metric implementation
    
    based off Java implementation
    - marked current metric interface as deprecated
    - new metric implemented with prefix "Stable". This can be dropped after a deprecation period
    - stubbed out MeterProvider, Meter, and Instruments
    - added Reader & View object stubs


bryce-b:
squashed and removed the `wip` commit.

Issue #293, opened 2022-02-16
Update Metrics "labels" to use "attributes"


nachoBonafonte:
It looks like there was a breaking changed where labels has been deprecated now in [otel-proto](https://github.com/open-telemetry/opentelemetry-proto/pull/342) and attributes should be used in it's place. This is the  [otel-js fix](https://github.com/open-telemetry/opentelemetry-js/pull/2523/files)

nachoBonafonte:
Version 1.1.3 fixes this issue

PR #294, opened 2022-03-11
Update opentelemetry/proto spec

trevor-dialpad:
A couple things done here:

1. Updated the proto based on the [latest spec](https://github.com/open-telemetry/opentelemetry-proto)
2. Updated the metrics adapter to support the new spec

This should solve #293

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocol/metric/MetricsAdapter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocol/metric/MetricsAdapter.swift
@@ -54,60 +54,63 @@ struct MetricsAdapter {
                 guard let gaugeData = $0 as? SumData<Double> else {
                     break
                 }
-                var protoDataPoint = Opentelemetry_Proto_Metrics_V1_DoubleDataPoint()
+                var protoDataPoint = Opentelemetry_Proto_Metrics_V1_NumberDataPoint()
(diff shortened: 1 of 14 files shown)


trevor-dialpad:
This is a draft because the metrics I'm posting from my client aren't being successfully picked up my the collector - I'm trying to figure out _why_ that is so I can address it but I haven't found much to go off of so far. If anyone has any clue as to why these metrics won't be picked up by the opentelemetry-collector with OTLP/GRPC I'd appreciate any ideas.

trevor-dialpad:
A bit more details I have since posting this draft:

1. [This is the commit](https://github.com/open-telemetry/opentelemetry-collector/pull/4724/files) that caused metrics to stop working with the Otel collector as of [v0.44.0](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.44.0). This removed the backwards compatibility to support metrics being sent based on the older spec (using "labels", "int_histogram", etc.)

2. This exactly what I did to generate the new spec. Then I went and manually updated some of the generated proto swift functions and variables to be public in order to compile.
```
protoc --swift_out=. opentelemetry/proto/*/*/v1/*.proto
protoc --grpc-swift_out=. opentelemetry/proto/*/*/v1/*.proto
```

trevor-dialpad:
Fixed with: https://github.com/open-telemetry/opentelemetry-swift/pull/300

PR #305, opened 2022-06-16
Resolves testWhenBatchFails_thenIntervalIncreases failure.

BennettSmith:
The `testWhenBatchFails_thenIntervalIncreases` test fails when `exportExpectation.fulfill()` is called multiple times. I'm not sure why the `mockDataExporter.onExport` closure is being called more than once, but it is. Since the mock closure calls `fulfill`, the test fails.  Setting the `assertForOverFulfill` property to `false` allows the test to pass.


nachoBonafonte (approved):
Good catch, thanks a lot

nachoBonafonte:
Try updating with main branch, some flaky test should be fixed

BennettSmith:
> Try updating with main branch, some flaky test should be fixed

Will do, but I don't think that will resolve the failures shown in this PR.  Looking at the build log, all three of the failures seem to be related to a problem with the `temporaryDirectory` global variable used by the tests.  I think the failures are because the temporary directory is not set to a new value for each new test case.  You can see that some tests in separate test suites (all part of the same test bundle) have the exact same name for the temporary directory. I don't think this is what was intended.

PR #306, opened 2022-06-27
fix race condition in TracerProviderSdk.get when creating a new tracer

seemk:
When multiple threads attempt to get a tracer from tracer provider and the tracer has not yet been constructed, there's a race condition, which usually leads to `EXC_BAD_ACCESS`:

image

As it's hard to create a reliable test case for it, I didn't add a test. However it can be verified by adding the following test to `TracerSdkTests.swift`:

```swift
func testTracerAccess() {
    for i in 0...100000 {
        let name = "abc \(i)"
        for _ in 0...8 {
            let t = Thread(block: {
                let _ = self.tracerSdkFactory.get(instrumentationName: name)
            })
            t.start()
        }
    }
}
```

### DIFF
--- a/Sources/OpenTelemetrySdk/Trace/TracerProviderSdk.swift
+++ b/Sources/OpenTelemetrySdk/Trace/TracerProviderSdk.swift
@@ -7,6 +7,7 @@ import Foundation
 import OpenTelemetryApi
 
 public class TracerProviderSdk: TracerProvider {
+    private var tracerLock = pthread_rwlock_t()
     private var tracerProvider = [InstrumentationLibraryInfo: TracerSdk]()
     internal var sharedState: TracerSharedState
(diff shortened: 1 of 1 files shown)


seemk:
The error in CI looks to be the same as on main branch 🤔

nachoBonafonte (approved):
LGTM. Thanks a lot for your contribution.

nachoBonafonte:
Hi Siim,
There is no planned release, but if you need one for deployment I can just prepare one with the current main branch. Just let me know.

seemk:
A new tag would be great, thanks a lot! I believe it was causing rare issues in production

nachoBonafonte:
Its done: [1.1.4](https://github.com/open-telemetry/opentelemetry-swift/releases/tag/1.1.4 )

PR #307, opened 2022-07-01
Removes extra call to closure

BennettSmith:
Removed the line that explicitly called `completion(.allow)`.  Previously, when that was called it resulted in a message about API misuse. The message was as follows:

```
API MISUSE: dataTask:didReceiveResponse:completionHandler: completion handler called more than once completion(.allow)
```


nachoBonafonte (approved):
LGTM, looks like a debug hook during implementation, thanks a lot

BennettSmith:
Is there anything else I need to do before this can be merged?

nachoBonafonte:
Sorry, missed the update ( vacationing )

PR #318, opened 2022-08-18
Change metric internals

bryce-b:
In trying to attempt to add MetricKit values for iOS I found that the data provided is pre-aggregated. This PR allows for pre-aggregated metric data to be added to the meterSDK.


bryce-b:
new method on meterSdk isn't visible due to access visibility. This is going to require a bit more work.

PR #321, opened 2022-08-24
Add API for Pre-aggregated data

bryce-b:
These PR adds an API that allows for pre-aggregated histogram data to be recorded through the metric API. 

I'll be adding an additional API for counted metrics


nachoBonafonte (approved):
Everything looks good to me, I am just trying to understand whats is failing in the github action to be failing the build. Quite strange, it works perfectly on local both in Xcode and command line

PR #324, opened 2022-09-08
Improves testability of EnvVarResource.

BennettSmith:
This is a small addition to the EnvVarResource to allow testing with a provided environment array. It adds a static `get` method which takes a `[String: String]` array as an input parameter. The parameter has a default value of `ProcessInfo.processInfo.environment`. This makes it possible to write tests that verify that setting the `OTEL_RESOURCE_ATTRIBUTES` environment variable works as expected.  It also makes it possible for client apps that wish to construct their own Resource instance to utilize the `EnvVarResource` class.

The original `EnvVarResource.resource` static property remains, so this is not a breaking change.  

Tests of the new behavior were added too.


nachoBonafonte (approved):
All looks good to me, thanks

PR #325, opened 2022-09-09
Improve URLSessionInstrumentation loading time

nachoBonafonte:
Improve URLSessionInstrumentation loading time by injecting the methods concurrently.
As this method is usually going to happen at app start it reduces the time without hurting other processes running


bryce-b (approved):

bryce-b:
will this still block the calling thread? It might cause issues if it isn't blocking, and dependent objects are initialized while the instrumentation is occurring

nachoBonafonte:
Yes, it waits until all iterations have finished, its not done in the background.

PR #337, opened 2022-10-20
Use String key for tracerProvider dictionary

mattxw:
We encountered some crash regarding InstrumentationLibraryInfo that mutated during runtime. 

```
Crashed: 
0  libswiftCore.dylib             0x1125c _assertionFailure(_:_:flags:) + 248
1  libswiftCore.dylib             0x12f8e0 KEY_TYPE_OF_DICTIONARY_VIOLATES_HASHABLE_REQUIREMENTS(_:) + 2604
2  App                      0x1868300 specialized _NativeDictionary.mutatingFind(_:isUnique:) + 3301816
3  App                      0x18624ac TracerProviderSdk.get(instrumentationName:instrumentationVersion:) + 3277668 (<compiler-generated>:3277668)
4  App                      0x186835c protocol witness for TracerProvider.get(instrumentationName:instrumentationVersion:) in conformance TracerProviderSdk + 3301908 (<compiler-generated>:3301908)
```
We tried to make the property inside InstrumentationLibraryInfo as immutable, but the crash still happen. Our latest fix is to change dictionary key in `TracerProviderSdk.tracerProvider` into String type. So far, this crash does not happen anymore in our app with this fix.

### DIFF
--- a/Sources/OpenTelemetrySdk/Common/InstrumentationLibraryInfo.swift
+++ b/Sources/OpenTelemetrySdk/Common/InstrumentationLibraryInfo.swift
@@ -23,4 +23,9 @@ public struct InstrumentationScopeInfo: Hashable, Codable {
         self.name = name
         self.version = version
     }
+	
+	public var stringVersion: String {
+		return "\(name):\(version ?? "main")"
+	}
+	
 }
--- a/Sources/OpenTelemetrySdk/Trace/TracerProviderSdk.swift
+++ b/Sources/OpenTelemetrySdk/Trace/TracerProviderSdk.swift
@@ -8,7 +8,7 @@ import OpenTelemetryApi
 
 public class TracerProviderSdk: TracerProvider {
     private var tracerLock = pthread_rwlock_t()
-    private var tracerProvider = [InstrumentationScopeInfo: TracerSdk]()
+    private var tracerProvider = [String: TracerSdk]()
     internal var sharedState: TracerSharedState
     internal static let emptyName = "unknown"
 
@@ -45,20 +45,20 @@ public class TracerProviderSdk: TracerProvider {
             instrumentationName = TracerProviderSdk.emptyName
         }
         let instrumentationScopeInfo = InstrumentationScopeInfo(name: instrumentationName, version: instrumentationVersion ?? "")
-
+        
         if pthread_rwlock_rdlock(&tracerLock) == 0 {
-            let existingTracer = tracerProvider[instrumentationScopeInfo]
+            let existingTracer = tracerProvider[instrumentationScopeInfo.stringVersion]
             pthread_rwlock_unlock(&tracerLock)
-
+            
             if existingTracer != nil {
                 return existingTracer!
             }
         }
-
+        
         let tracer = TracerSdk(sharedState: sharedState, instrumentationScopeInfo: instrumentationScopeInfo)
-
+        
         if pthread_rwlock_wrlock(&tracerLock) == 0 {
-            tracerProvider[instrumentationScopeInfo] = tracer
+            tracerProvider[instrumentationScopeInfo.stringVersion] = tracer
             pthread_rwlock_unlock(&tracerLock)
         }
 


nachoBonafonte:
Hi @mattxw, I am not sure this is the proper fix for the issue. InstrumentationScopeInfo is an struct so it should not mutate during time and in the way it is implemented only in the initializer. 

How are you using it?, do you have a reproducible example for the issue?, could it be showing another unrelated error which is writing out of bonds memory and affecting it? Have you run your code with  Thread sanitizer or any other sanitizer?

Thanks a lot

nachoBonafonte (changes requested):
I am not sure this is the proper fix for the issue. InstrumentationScopeInfo is an struct so it should not mutate during time and in the way it is implemented only in the initializer.

How are you using it?, do you have a reproducible example for the issue?, could it be showing another unrelated error which is writing out of bonds memory and affecting it? Have you run your code with Thread sanitizer or any other sanitizer?

mattxw:
> I am not sure this is the proper fix for the issue. InstrumentationScopeInfo is an struct so it should not mutate during time and in the way it is implemented only in the initializer.
> 
> How are you using it?, do you have a reproducible example for the issue?, could it be showing another unrelated error which is writing out of bonds memory and affecting it? Have you run your code with Thread sanitizer or any other sanitizer?

Thanks for the response. We can't reproduce the crash, as we found it from Firebase Crashlytics. We are using [Splunk](https://github.com/signalfx/splunk-otel-ios/) that use OpenTelemetry library for observability.

I checked some of the crash session and there's one iPhone 8 Plus user with 400MB+ RAM free, so I think that wasn't out of bounds memory. We tried to change InstrumentationScopeInfo property to immutable (`let` instead of `var`), but the crash still happen. 

We haven't tried Thread Sanitizer though, since this solution fixed the crash in our current release.

nachoBonafonte:
Hi, InstrumentationScopeInfo is an struct and inmutable once initialised, being the properties let or var should not make a difference. 

As I said, I don't think this is the solution, and I think it's just hiding the real issue, which is probably a concurrent access to the Dictionary and that `Thread Sanitizer` should show even when not crashing.

On this [PR](https://github.com/open-telemetry/opentelemetry-swift/pull/306) a race condition was fixed that was related to the access of that dictionary. The PR was merged 27-Jun-2022, any release from the library after that [1.1.4 ](https://github.com/open-telemetry/opentelemetry-swift/releases/tag/1.1.4) and newer, should have it fixed. 

Please check with version of opentelemetry-swift is being used by the version of the Splunk you are linking, if its older than 1.1.4, then your crash is probably already fixed. If it's newer than it,then it means that there is another thread race condition that `Thread Sanitizer` should warn about and that we can fix when we know where it happens.

mattxw:
Hi, @nachoBonafonte. Thanks for the advice, I tried your suggestion to check Thread Sanitizer to reproduce the crash stack trace, but there was no warning regarding OpenTelemetry. I also tried the latest version of OpenTelemetry and it has an access race on TracerProviderSdk. 

```
==================
WARNING: ThreadSanitizer: Swift access race (pid=4090)
  Modifying access of Swift variable at 0x7b3c00001d20 by thread T16:
	#0 TracerProviderSdk.get(instrumentationName:instrumentationVersion:) <null>:2 (App:x86_64+0x1035e17dd)
	#1 protocol witness for TracerProvider.get(instrumentationName:instrumentationVersion:) in conformance TracerProviderSdk <null>:2 (App:x86_64+0x1035e3027)
	#2 buildTracer() <null>:2 (App:x86_64+0x103410b12)
    #3 static SplunkRum.log(name:type:parent:) <null>:2 (App:x86_64+0x103413136)
	#4 SplunkService.trackOperation(_:) SplunkService.swift:83 (App:x86_64+0x101c2960f)
	#5 protocol witness for SplunkServiceable.trackOperation(_:) in conformance SplunkService <compiler-generated> (App:x86_64+0x101c2c91d)
	#6 SplunkSubInterceptor.afterRpcOperation(_:) SplunkInterceptor.swift:69 (App:x86_64+0x102acb92d)
	#7 SplunkInterceptor.afterRpcOperation(_:) SplunkInterceptor.swift:46 (App:x86_64+0x102acb00a)
	#8 @objc SplunkInterceptor.afterRpcOperation(_:) <compiler-generated> (App:x86_64+0x102acb9ef)
	#9 __49-[MainInterceptor afterRpcOperation:]_block_invoke <null>:2 (App:x86_64+0x105d7bc6f)
	#10 _dispatch_client_callout <null>:2 (libdispatch.dylib:x86_64+0x4a5a)
	#11 -[MainInterceptor afterRpcOperation:] <null>:2 (App:x86_64+0x105d7ba6c)
	#12 _dispatch_client_callout <null>:2 (libdispatch.dylib:x86_64+0x4a5a)

  Previous modifying access of Swift variable at 0x7b3c00001d20 by thread T27 (mutexes: write M898603468670800912):
	#0 TracerProviderSdk.get(instrumentationName:instrumentationVersion:) <null>:2 (App:x86_64+0x1035e17dd)
	#1 protocol witness for TracerProvider.get(instrumentationName:instrumentationVersion:) in conformance TracerProviderSdk <null>:2 (App:x86_64+0x1035e3027)
	#2 buildTracer() <null>:2 (App:x86_64+0x103410b12)
	#3 startHttpSpan(request:) <null>:2 (App:x86_64+0x103423cb4)
	#4 SessionTaskObserver.observeValue(forKeyPath:of:change:context:) <null>:2 (App:x86_64+0x103425978)
	#5 @objc SessionTaskObserver.observeValue(forKeyPath:of:change:context:) <null>:2 (App:x86_64+0x1034262b4)
	#6 NSKeyValueNotifyObserver <null>:2 (Foundation:x86_64+0xb126d)
	#7 _dispatch_client_callout <null>:2 (libdispatch.dylib:x86_64+0x4a5a)
```
Regardless, I don't think it is related, as we also implemented NSRecursiveLock on the dictionary before. Do you have any other suggestion for this crash? Or maybe a suggestion to make this solution proper? Thanks before.

nachoBonafonte:
Could you provide the thread sanitizer error symbolicated? If running on debug and with debug symbols generated, Thread Sanitizer will show the file and line of the offending access. 

What are the exact versions you are using, both opentelemetry-swift and the splunk library?

nachoBonafonte:
Can you show where are the accesses? Clicking on the left view in each of the mutating accesses

mattxw:
Both thread have similar stack trace, but here's what I can show for the 2nd and 3rd stack.

Screen Shot 2022-10-25 at 00 17 41
Screen Shot 2022-10-25 at 00 17 14

nachoBonafonte:
Sorry I think I was not clear enough, I want to know where are the accesses being done in the library, so the stack line 0 of the mutating acces by thread 84 and by thread 20

mattxw:
Hey, @nachoBonafonte, sorry for being MIA. Here's the access mentioned by Xcode. It's the same part of library that I posted weeks ago.

Screenshot 2022-11-03 at 23 54 38

Screenshot 2022-11-03 at 23 54 27

nachoBonafonte:
That the offending line is the one that access the lock is very suspicious, are you sure you are not linking OpenTelemetry-swift twice? Maybe Splunk library is linking it statically inside and you are linking it again? It should show some linking warning if that is happening.

I will try to create a test that exercise this code concurrently and check if some thread sanitizer message appears.

mattxw:
> That the offending line is the one that access the lock is very suspicious, are you sure you are not linking OpenTelemetry-swift twice? Maybe Splunk library is linking it statically inside and you are linking it again? It should show some linking warning if that is happening.
> 
> I will try to create a test that exercise this code concurrently and check if some thread sanitizer message appears.

I don't think that's the case. I can only find Splunk on my linked libraries list. 🤔 

Anyway, I don't think the cause of this crash (on the first post) is access race/race condition as we also put `NSRecursiveLock` to access the dictionary before. I've tried to run thread sanitizer with our old branch and there are no warning on concurrent access anymore. Please let me know if there's something I can do for this. 🙏🏼

PR #340, opened 2022-10-21
Add Callback to Set Parent in URLSessionInstrumention

gjeltep:
This adds the ability to use a callback in `URLSessionInstrumentationConfiguration` for the purpose of setting a parent for the span produced by `URLSessionInstrumentation`. The `URLSessionInstrumentation` tooling provides excellent out-of-box support, but seems to emit only parentless spans. This addition is useful in the case of a networking call being a child of another (in- or inter-) process.

### DIFF
--- a/Sources/Instrumentation/URLSession/README.md
+++ b/Sources/Instrumentation/URLSession/README.md
@@ -19,6 +19,8 @@ This behaviour can be modified or augmented by using the optional callbacks defi
 
 `nameSpan: ((URLRequest) -> (String)?)?` - Modifies the name for the given request instead of stantard Opentelemetry name
 
+`setParent: ((URLRequest) -> (SpanContext)?)?` - Sets the Span's parent using the provided SpanContext
+
 `createdRequest: ((URLRequest, Span) -> Void)?` - Called after request is created,  it allows to add extra information to the Span
 
 `receivedResponse: ((URLResponse, DataOrFile?, Span) -> Void)?`- Called after response is received,  it allows to add extra information to the Span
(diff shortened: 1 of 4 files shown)


nachoBonafonte:
Hi @gjeltep , just an initial comment before evaluating the PR:

 The network instrumentation DO set the current active Span as the parent of the spans it creates. It follows the scheme explained [here](https://opentelemetry.io/docs/instrumentation/swift/manual/#creating-nested-spans). 

I would say that for most users of the library, the default way to create any span span should be always setting it as `the active` in the span builder, so it will automatically be the parent of any other span created in the same execution context while it is not ended. So create the span something like this: 
` let newSpan = someTracer.spanBuilder(spanName: "my span").setActive(true)`

Making all spans active by default makes the parent-child relationship to flow naturally with the code execution and with no effort from the developer (the automatic context handling works extremely well and can show child-parent relations that even the developer doesn't know). I would say that only in some very specific spans the developer should set the parent of the span specifically.

That does not mean that having a callback for this very specific situation in the network request is not useful, but I think that you will get your desired behavior just by setting all spans as active on creation. Could you try it and let me know how it works for you?

gjeltep:
Hey @nachoBonafonte - thank you for your feedback! I didn't fully appreciate the use of `setActive()`. After testing, it does provide the desired behavior in some, but not all of my in-process use cases. I'm also interested in extracting the propagated context from a network response and tagging the parent on spans for outgoing requests accordingly. Would this PR help in this situation, as well?

nachoBonafonte:
If we are adding span customization, which I think is  a nice idea, I would prefer to use a more generic approach, something that allows to change something more than the parent, what about something like:

`spanCustomization: ((URLRequest, SpanBuilder) -> Void)?

so the user can use the SpanBuilder to change not only the parent, but also add a link or set any other attribute?

gjeltep:
Excellent sounds good - I'll make those changes asap!

nachoBonafonte (approved):
It looks much better now. Thanks a lot for your contribution

PR #343, opened 2022-11-05
Add support for OTLP HTTP

### LINKED ISSUE #312: Feature Request: OTLP exporter via HTTP (not GRPC)
Would be handy for servers which can't support gRPC 🙂  we've run into huge issues internally because of this limitation with companies such as Akamai not supporting gRPC over HTTP/2 from edge to origin which is needed for collectors.

ptescher:
Would fix #312


Sources/Exporters/OpenTelemetryProtocol/metric/OltpHTTPMetricExporter.swift
```diff
@@ -0,0 +1,55 @@
...
+            .with {
+                $0.resourceMetrics = MetricsAdapter.toProtoResourceMetrics(metricDataList: metrics)
+            }
+
+        var request = URLRequest(url: endpoint)
+        request.httpMethod = "POST"
+        request.httpBody = try? body.serializedData()
+        request.setValue("application/x-protobuf", forHTTPHeaderField: "Content-Type")
+        
+        let task = URLSession.shared.dataTask(with: request) { _, response, error in
```
Sherlouk, line 41: URL session can be passed in, but is not used.

I think most clients should avoid using the shared session, in favour of a custom configuration which uses a background priority to prevent observability from slowing down the main application itself.

Comment applies to both trace and metric exporter

Sources/Exporters/OpenTelemetryProtocol/metric/OltpHTTPMetricExporter.swift
```diff
@@ -0,0 +1,55 @@
...
+        let metrics = pendingMetrics
+        pendingMetrics = []
+        let body = Opentelemetry_Proto_Collector_Metrics_V1_ExportMetricsServiceRequest
+            .with {
+                $0.resourceMetrics = MetricsAdapter.toProtoResourceMetrics(metricDataList: metrics)
+            }
+
+        var request = URLRequest(url: endpoint)
+        request.httpMethod = "POST"
+        request.httpBody = try? body.serializedData()
```
Sherlouk, line 38: Errors not being captured, should we have a callback or some way to recover? 

At least an error message with a reason why?

nachoBonafonte:
The PR should also include some tests covering the new code

nachoBonafonte:
This functionality has been added to the project in #376, which was initially based on this commit

PR #349, opened 2022-11-14
Fix ActivityContextManager  contextMap is never being cleaned up

### LINKED ISSUE #348: ActivityContextManager.swift `contextMap` is never being cleaned up
https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/OpenTelemetryApi/Context/ActivityContextManager.swift#L72

contexts are added to the context map in `setCurrentContextValue(forKey key: OpenTelemetryContextKeys, value: AnyObject) `
but never cleaned up in `removeContextValue(forKey key: OpenTelemetryContextKeys, value: AnyObject) `

nachoBonafonte:
fixes  #348

Fix also  tests that were passing because of the leak instead of properly synchronizing themAlso 

Remove SwiftMetricShim dependency from NIO, copying the Locks to the project

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -65,7 +65,6 @@ let package = Package(
                 exclude: ["README.md"]),
         .target(name: "SwiftMetricsShim",
                 dependencies: ["OpenTelemetrySdk",
-                               .product(name: "NIO", package: "swift-nio"),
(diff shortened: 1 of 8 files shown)


bryce-b (approved):

vvydier (commented):
hopefully this fixes the leak reported by Bryce

PR #363, opened 2022-12-21
Fix iOS support

lunij:
This fixes the iOS support by removing all compilation errors. This means that the Swift Package [Reachability.swift](https://github.com/ashleymills/Reachability.swift) is added as dependency because it was used, but not compiling.

To ensure proper iOS compilation and testing, GitHub workflow steps are added.


nachoBonafonte (approved):
LGTM, Thanks a lot for your contribution

PR #365, opened 2023-01-05
Fix multiple Xcode warnings

emanuelschmoczer:
This PR fixes multiple small Xcode warnings/issues currently visible in Xcodes "Issue navigator":

- Use `ClientConnection.Configuration.default` instead of deprecated `init`
- Use `NIOLock` instead of deprecated `Lock`
- Silence warning about `retain_me` being written but never read
- Change `testHistogram` to let, because it was never mutated
- Add assertion for `subtype`


nachoBonafonte (approved):
Everything LGTM, thanks a lot for your contribution

PR #376, opened 2023-02-08
Add an OTLP via HTTP exporter

### LINKED ISSUE #312: Feature Request: OTLP exporter via HTTP (not GRPC)

kevinearls:
Signed-off-by: Kevin Earls <kevinearls@mac.com>

This fixes #312 and is based on code from PR #343   I will add an example and a test for the metrics exporter, but first I have several questions:

1. Is it ok to use NIOTestUtils.NIOHTTP1TestServer?  
2. How can I convert the body it returns back to [SpanData].   (See testExport() in OtlpHttpTraceExporterTests.swift)
3. Why am I getting "Connection refused" and "The network connection was lost" (sometimes) messages when I run that test, even though the test appears to be working?
4. Is it ok that shutdown() in OtlpHttpTraceExporter is a no-op?
5. I copied HTTPClient.swift from DataDog Exporter.  Is that ok or should I move the original so it can be used by both?

Any reviews or other comments would be appreciated.  Please be nice though, don't forget that I am new to Swift.  😆


Sources/Exporters/OpenTelemetryProtocol/metric/OltpHTTPMetricExporter.swift
```diff
@@ -0,0 +1,61 @@
...
+import OpenTelemetrySdk
+import Foundation
+
+public func defaultOltpHTTPMetricsEndpoint() -> URL {
+    URL(string: "http://localhost:4318/v1/metrics")!
+}
+
+public class OtlpHttpMetricExporter: MetricExporter {
+    let endpoint: URL
+    let urlSession: URLSession // TODO do we need this?
```
nachoBonafonte, line 15: No, HTTPClient can be initialized with a given URLSession, so this is superfluous

Sources/Exporters/OpenTelemetryProtocol/metric/OltpHTTPMetricExporter.swift
```diff
@@ -0,0 +1,61 @@
...
+    private let httpClient: HTTPClient
+    var pendingMetrics: [Metric] = []
+    
+    public init(endpoint: URL = defaultOltpHTTPMetricsEndpoint(), urlSession: URLSession = URLSession.shared) {
+        self.endpoint = endpoint
+        self.httpClient = HTTPClient();
```
nachoBonafonte, line 21: This should use the provided URLSession, and remove following line

nachoBonafonte: Same for the rest of exporters

Tests/ExportersTests/OpenTelemetryProtocol/OtlpHttpMetricsExporterTest.swift
```diff
@@ -0,0 +1,116 @@
...
+
+        let result = exporter.export(metrics: metrics) { () -> Bool in
+            false
+        }
+        XCTAssertEqual(result, MetricExporterResultCode.success)
+        
+        XCTAssertNoThrow(try testServer.receiveHead())
+        XCTAssertNoThrow(try testServer.receiveBodyAndVerify() { body in
+            var contentsBuffer = ByteBuffer(buffer: body)
+            let contents = contentsBuffer.readString(length: contentsBuffer.readableBytes)!
```
kevinearls, line 91: Hi @vvydier If I remember correctly you know a bit about using protobuf.  Is there any way I can turn the ByteBuffer received here back into [Metric] (or [SpanData] in the OtlpHttpTraceExporterTests)?  Thanks.

nachoBonafonte:
> @nachoBonafonte Can you tell me why the build is now failing? Thanks.
Not sure, it looks like it fails to find NIOTestUtils, but is in theory there

nachoBonafonte:
> @nachoBonafonte Can you tell me why the build is now failing? Thanks.

Yes, with the latest change it will use package@swift-5.6.swift for newer Xcode versions, the library is missing there

Sources/Exporters/OpenTelemetryProtocol/trace/utils/HTTPClient.swift
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import Foundation
+
+/// Client for sending requests over HTTP.
+internal final class HTTPClient {
+    private let session: URLSession
+
+    convenience init() {
+        let configuration: URLSessionConfiguration = .ephemeral
```
Sherlouk, line 13: Is it worth specifying "wait for connectivity" or a session task priority (ensuring that OTLP data doesn't slow down UI updates)?

Sources/Exporters/OpenTelemetryProtocol/trace/OtlpHTttpTraceExporter.swift
```diff
@@ -0,0 +1,55 @@
...
+        let body = Opentelemetry_Proto_Collector_Trace_V1_ExportTraceServiceRequest.with {
+            $0.resourceSpans = SpanAdapter.toProtoResourceSpans(spanDataList: spans)
+        }
+        
+        var request = URLRequest(url: endpoint)
+        request.httpMethod = "POST"
+        request.httpBody = try? body.serializedData()
+        request.setValue("application/x-protobuf", forHTTPHeaderField: "Content-Type")
+                
+        httpClient.send(request: request) { result in
```
Sherlouk, line 41: Which thread does flush get called on? Given the amount of time it could take for a send to occur, are we at risk at blocking other work on same thread?

kevinearls: @Sherlouk Thanks for the comments.  I am relatively new to Swift (I do have some background in Observability) so I really don't know.  Any suggestions you have on how to do this would be welcome.

Sources/Exporters/OpenTelemetryProtocol/trace/utils/HTTPClient.swift
```diff
@@ -0,0 +1,55 @@
...
+        // and move session configuration constants to `PerformancePreset`.
+        self.init(session: URLSession(configuration: configuration))
+    }
+
+    init(session: URLSession) {
+        self.session = session
+    }
+
+    func send(request: URLRequest, completion: @escaping (Result<HTTPURLResponse, Error>) -> Void) {
+        let task = session.dataTask(with: request) { data, response, error in
```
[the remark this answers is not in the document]
nachoBonafonte: It should not, but the traces will appear. There is also a callback function in the configuration that can be used to exclude the internal URL's for reporting: `shouldInstrument`, and return `false` for the URLs we want to exclude

Sources/Exporters/OpenTelemetryProtocol/trace/utils/HTTPClient.swift
nachoBonafonte: Yes is a good idea, anyway we should also provide a constructor which accepts an existing URLSession to use

Sources/Exporters/OpenTelemetryProtocol/trace/OtlpHTttpTraceExporter.swift
nachoBonafonte: The flush method is explicitly called in the spanProcessor when called using forceFlush, and will execute on the calling thread (as documentation for Span Processor specifies), it is also called when explicitly calling shutdown().

Sources/Exporters/OpenTelemetryProtocol/trace/OtlpHTttpTraceExporter.swift
```diff
@@ -0,0 +1,55 @@
...
+
+public class OtlpHttpTraceExporter: SpanExporter {
+    let endpoint: URL
+    private let httpClient: HTTPClient
+    var pendingSpans: [SpanData] = []
+     
+    public init(endpoint: URL = defaultOltpHttpTracesEndpoint()) {
+        self.endpoint = endpoint
+        self.httpClient = HTTPClient()
+    }
```
nachoBonafonte, line 21:
```suggestion
       public init(endpoint: URL = defaultOltpHttpTracesEndpoint(), useSession: URLSession? = nil) {
        self.endpoint = endpoint
        if let providedSession = useSession {
            self.httpClient = HTTPClient(session: providedSession)
        } else {
            self.httpClient = HTTPClient()
        }
    }

Comment (nachoBonafonte) [reply Sources/Exporters/OpenTelemetryProtocol/trace/OtlpHTttpTraceExporter.swift:21]: Something like this allows reusing existing URLSession from the user. Would need changing all exporters

**Review by nachoBonafonte — CHANGES_REQUESTED** [review]
Looks good to me, just a small change and I think we can merge if others agree

File: Sources/Exporters/OpenTelemetryProtocol/trace/OtlpHTttpTraceExporter.swift (hunk 5 above)
Comment (Sherlouk) [reply Sources/Exporters/OpenTelemetryProtocol/trace/OtlpHTttpTraceExporter.swift:45]: In this scenario, Nacho, flush is called on export within these new HTTP exporters so it will be the default.

I'm just not sure from memory what the threading is like within the module, and whether exports are already pushed onto a different thread or not. I'm conscious that URLSession is prone to blocking threads is all.

File: Sources/Exporters/OpenTelemetryProtocol/trace/utils/HTTPClient.swift (hunk 6 above)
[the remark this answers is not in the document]
Comment (Sherlouk) [reply Sources/Exporters/OpenTelemetryProtocol/trace/utils/HTTPClient.swift:28]: Do we need to update documentation at a minimum to reflect this?

Or a better solution may be to add a custom header into the HTTPClient, such as `X-OpenTelemetry-SDK: swift` (random example), check for this within the URLSession instrumentation and filter it out?

We've had problems before with instrumentation instrumenting itself, so know this is easy to have accidentally happen!

File: Sources/Exporters/OpenTelemetryProtocol/logs/OtlpHttpLogExporter.swift [hunk 8]
```diff
@@ -0,0 +1,68 @@
...
+        let logRecords = pendingLogRecords
+        pendingLogRecords = []
+            
+        let body = Opentelemetry_Proto_Collector_Logs_V1_ExportLogsServiceRequest.with { request in
+            request.resourceLogs = LogRecordAdapter.toProtoResourceRecordLog(logRecordList: logRecords)
+        }
+
+        var request = URLRequest(url: endpoint)
+        request.httpMethod = "POST"
+        request.httpBody = try? body.serializedData()
```
Sherlouk, line 54: It may be best to wrap this in a do/catch and print out the error. Sending nil data up isn't going to help anybody.

Sources/Exporters/OpenTelemetryProtocol/metric/OltpHTTPMetricExporter.swift
```diff
@@ -0,0 +1,63 @@
...
+        let metrics = pendingMetrics
+        pendingMetrics = []
+        let body = Opentelemetry_Proto_Collector_Metrics_V1_ExportMetricsServiceRequest
+            .with {
+                $0.resourceMetrics = MetricsAdapter.toProtoResourceMetrics(metricDataList: metrics)
+            }
+
+        var request = URLRequest(url: endpoint)
+        request.httpMethod = "POST"
+        request.httpBody = try? body.serializedData()
```
Sherlouk, line 44: There's a lot of duplication between the three exporters - what can we do to minimise that? There's a maintenance overhead and just means any tweak has to be done 3 times inflating effort.

Sources/Exporters/OpenTelemetryProtocol/metric/OltpHTTPMetricExporter.swift
```diff
@@ -0,0 +1,63 @@
...
+        httpClient.send(request: request) { result in
+            switch result {
+            case .success(_):
+                exporterResult = MetricExporterResultCode.success
+            case .failure(let error):
+                print("ERROR: \(error)")
+                exporterResult = MetricExporterResultCode.failureNotRetryable // FIXME how do I know what type of failure?
+            }
+        }
+        return exporterResult
```
Sherlouk, line 56: I'm not sure I get what's happening here. exporterResult is set in the async block, but the value will be returned before the network request has completed?

We could get around that using a semaphore which would stop the return from being called until the network request has finished but that's not what we have currently (it may help us get rid of the 'sleep' in HTTPClient though)

Thoughts Nacho?

Sources/Exporters/OpenTelemetryProtocol/trace/utils/HTTPClient.swift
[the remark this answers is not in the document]
nachoBonafonte: > Do we need to update documentation at a minimum to reflect this?
Yes, documentation is currently very basic, what more things you think that should be clarified? I wrote the current one, and I thought it was short  but clear, thats why I'd like some feedback

> Or a better solution may be to add a custom header into the HTTPClient, such as `X-OpenTelemetry-SDK: swift` (random example), check for this within the URLSession instrumentation and filter it out?
I would prefer not to add extra headers just to identify these requests, since other internal requests might also be filtered by the user, I think clarifying how to filter is a better option

nachoBonafonte:
Just for context of this PR, the idea we talked last SIG meeting was to merge it early in a beta status, so people can test, debug and add add PR's to improve it directly. Kevin wanted to help with the project but has no real usage for this code, so just implemented without a clear scenario but knowing that many users were interested in it, once merged everyone can test in its own scenario and improve it. That doesn't mean we cannot add feedback before merging, but dont expect this code to be considered production ready even when merged

Sources/Exporters/OpenTelemetryProtocol/logs/OtlpHttpLogExporter.swift
nachoBonafonte: True. can you change it @kevinearls?

Sources/Exporters/OpenTelemetryProtocol/trace/utils/HTTPClient.swift
```diff
@@ -0,0 +1,55 @@
...
+            completion(httpClientResult(for: (data, response, error)))
+        }
+        
+        task.resume()
+        Thread.sleep(forTimeInterval: 1)  // FIXME is there a better way to handle this?
```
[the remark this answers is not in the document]
nachoBonafonte: Sorry i didn't see this in the last review. This sleep here makes no sense, this should be handled asynchronously. The wait should be done on the checking side using the completion callback

Sources/Exporters/OpenTelemetryProtocol/metric/OltpHTTPMetricExporter.swift
nachoBonafonte: Yes, the sleep in HTTPClient makes no sense, we should not need to synchronize after the sending by default, only maybe on shutdown

Sources/Exporters/OpenTelemetryProtocol/trace/OtlpHTttpTraceExporter.swift
nachoBonafonte: True, the code is not very clear,  I missed it. Using flush in the name by default looks like synchronization after sending, we should not block by default.

Sherlouk:
@nachoBonafonte - please don't take my comments as blockers!

Is it perhaps worth prefixing the public exporter classes with an underscore as an indication that it's not finalised? Or at least some documentation in the README to explain the status of individual elements?

Anything that'll make it easier for folks to pick up work without discussions being left on PRs never to be seen again 😄

Sources/Exporters/OpenTelemetryProtocol/trace/utils/HTTPClient.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    convenience init() {
+        let configuration: URLSessionConfiguration = .ephemeral
+        // NOTE: RUMM-610 Default behaviour of `.ephemeral` session is to cache requests.
+        // To not leak requests memory (including their `.httpBody` which may be significant)
+        // we explicitly opt-out from using cache. This cannot be achieved using `.requestCachePolicy`.
+        configuration.urlCache = nil
+        // TODO: RUMM-123 Optimize `URLSessionConfiguration` for good traffic performance
+        // and move session configuration constants to `PerformancePreset`.
+        self.init(session: URLSession(configuration: configuration))
```
kevinearls, line 20: @nachoBonafonte Isn't that what this does?

Sources/Exporters/OpenTelemetryProtocol/logs/OtlpHttpLogExporter.swift
kevinearls: Sure, I will work on as many of these as I can this afternoon.

Sources/Exporters/OpenTelemetryProtocol/trace/utils/HTTPClient.swift
nachoBonafonte: Sorry, @kevinearls , i dont know what you refer to

kevinearls: @nachoBonafonte This is about your comment from yesterday where you said "Yes is a good idea, anyway we should also provide a constructor which accepts an existing URLSession to use".  

I think I pointed to the wrong place though, I meant here: https://github.com/open-telemetry/opentelemetry-swift/pull/376/files#diff-2fdc8227312474239acd339b5f810132bd11efbc7c06c000a57bb24002b92a53R23

nachoBonafonte:
Hi @kevinearls, I pushed directly to your branch with the asynchronous approach for the span exporter, it only blocks in the flush method and tries to not lose spans if an error happens in the standard export. We should try to use a similar approach in the other exporters.

Sources/Exporters/OpenTelemetryProtocol/trace/utils/HTTPClient.swift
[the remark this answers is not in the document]
nachoBonafonte: I removed it in the latest change, if it fails it must be handled in other place

kevinearls:
@nachoBonafonte Thanks.  I will apply that to the other exporters shortly.

kevinearls:
@nachoBonafonte Hi.  If you have time can you take another look at this today?  Otherwise I am heading out to Barcelona tomorrow morning and won't be able to look at it again until next Monday.  Thanks.

nachoBonafonte (approved):

PR #377, opened 2023-02-09
watchOS exclusion/compatibility

lunij:
The current Swift Package does not compile with watchOS. This is because of some APIs are not available on watchOS. The dependency `Thrift-Swift` is watchOS incompatible as well and excluded from watchOS builds. To make this possible without another contribution to `Thrift-Swift`, the `swift-tools-version` is updated to `5.6`. This allows the usage of dependency conditions.

Why is it necessary to keep watchOS compilation working?
In case this Swift Package is used as a dependency in another Swift Package that supports watchOS, this package would make the integrating package to fail compilation. It does not mean that watchOS support needs to be fully implemented. It means the package needs to compile with watchOS, even if no source file is actually meant to be implemented for watchOS at all. Excluding all source files from watchOS builds using pre-processor declarations is solving this issue.

To ensure the Swift Package always compiles with watchOS, steps are added to the GitHub workflow.

Because watchOS was not compiling before this PR, I consider watchOS not being supported by this Swift Package. Due to time constraints, I disabled failing tests for watchOS.


nachoBonafonte:
LGTM, but there are a pair of things I would change:

-  Updating swift-tools means a breaking change, so it can be an issue for existing users which support old OS versions with old versions of the developer tools. You can create another versioned package leaving the original untouched, creating it with name `Package@swift-5.6.swift`, for toolchains 5.6 and higher it will take this one, for previous toolchains will use the non-versioned one.

- For the failing tests, could you use a conditional skip with availability as shown [here](https://developer.apple.com/documentation/xctest/methods_for_skipping_tests)? It helps in keeping track of the tests when run

What do you think?

nachoBonafonte (changes requested):
LGTM, but there are a pair of things I would change:

Updating swift-tools means a breaking change, so it can be an issue for existing users which support old OS versions with old versions of the developer tools. You can create another versioned package leaving the original untouched, creating it with name Package@swift-5.6.swift, for toolchains 5.6 and higher it will take this one, for previous toolchains will use the non-versioned one.

For the failing tests, could you use a conditional skip with availability as shown [here](https://developer.apple.com/documentation/xctest/methods_for_skipping_tests)? It helps in keeping track of the tests when run

What do you think?

lunij:
Sounds perfect!

nachoBonafonte (approved):
Looks great, just one nitpick, can you remove the watchOS support from the original package.swift

lunij:
watchOS support cannot be removed. The platform specification in a package manifest is just constraining the version, but not the platform itself. So each and every Swift Package is enabled to work on all platforms by default and it cannot be disabled. So for the Package.swift with `swift-tools-version` set to `5.2` to compile with watchOS, the `Thrift-Swift` dependency needs to compile with watchOS. This can be achieved by excluding source files with the pre-processor declarations only.

nachoBonafonte:
Oh, yes, you are true but is more because if it appears, it looks like it may work which is a bit misleading.

lunij:
So, I just remove the version constraint to give the impression watchOS support is not meant to be implemented? I can do that.

nachoBonafonte:
Thanks a lot, I will wait for the test to pass and will merge

nachoBonafonte:
Your commit with [Switch default manifest to be the most up-to-date one](https://github.com/open-telemetry/opentelemetry-swift/pull/377/commits/af9806436ba3a965398a69847644a7d8eaa5e3e5), I think wont work, the verision in the filename is the minimum used to open, you everything newere than 5.2 will be open with the old one

lunij:
> Your commit with [Switch default manifest to be the most up-to-date one](https://github.com/open-telemetry/opentelemetry-swift/pull/377/commits/af9806436ba3a965398a69847644a7d8eaa5e3e5), I think wont work, the verision in the filename is the minimum used to open, you everything newere than 5.2 will be open with the old one

I'm not sure about that. As far as I understand that versionized manifest feature, the manifest is chosen dependent on the current version of the Swift toolchain. It seems to not matter whether the Package.swift contains the minimum `swift-tools-version` or the maximum. Other open sourced projects mix this like you can see in the following repos:
https://github.com/Alamofire/Alamofire
https://github.com/apple/swift-nio
https://github.com/apple/swift-argument-parser

lunij:
Seeing some flaky tests to succeed and fail for watchOS, I'm asking myself whether it makes sense to remove the watchOS specification from both manifests. This PR is not really adding watchOS support. It rather makes it possible to be used in a project that needs to be compilable for watchOS.

lunij:
I played around with the package manifests using different swift-tools-versions.

[This one failed](https://github.com/lunij/codecov-demo/actions/runs/4162452119) because swift-tools-version was too high.
Then I added a [manifest for the lower swift-tools-version](https://github.com/lunij/codecov-demo/actions/runs/4162473012) needed using the suffix and it passed.

I had a look at the surprisingly new coverage report. The diff of file changes is not matching the actual file changes of the PR.

Because the CI is split into platforms, I recommend to update the repository settings for the required jobs.

In case any change is too much, please let me know and I'll extract it into another PR or drop it entirely.

nachoBonafonte:
Sorry for the delay, I think the versioning name for the package.swift file should be the opposite, the one with the lower swift-tools-version should be the non versioned one, and the one with swift-tools 5.6 should include it in the name `Package@swift-5.6.swift`, e.g [here](https://github.com/mxcl/PromiseKit)

lunij:
Can you please help me understand the reasoning behind the three examples are invalid I provided? They demonstrate the direction of manifest-version-naming does not matter. Otherwise it would mean Apple itself is doing it wrong and it would mean the most popular iOS framework Alamofire is doing it wrong.

lunij:
I'm okay with changing the direction of manifest-version-naming making the original `Package.swift` contain the lowest `swift-tools-version` if that moves the PR forward. But I need some feedback from you @nachoBonafonte @bryce-b, also regarding the CI jobs. Do you consider to switch the requirement of passing CI jobs to the three new ones or shall I revert that and create separate PRs for those kind of changes?

nachoBonafonte:
From the examples you provided [swift-argument-parser](https://github.com/apple/swift-argument-parser) uses the approach I was suggesting, I don't know why the others are like that, but in my experience this is the right format.

For the CI, yes I think your approach make more sense now, running in the three different platforms and in parallel, you can add it as part of this PR. Codecov is currently used as informative not forced, so we can fix it later.

lunij:
Yes, even at Apple they are struggling what direction of manifest-version-naming is the proper one. As all those repositories showcase, the direction does not matter. It is more a decision what original `Package.swift` manifest you wanna open with Xcode when working on the project. I prefer the newer one. Anyways, I'm reverting the commit for you.

nachoBonafonte:
Thanks a lot for the change and for your contribution. Merging now

PR #379, opened 2023-02-20
Fix a thread race condition that happens when span is ended in various threads

nachoBonafonte:
Use a concurrent dispatch queue to protect internal values, so they can be safely accessed concurrently
We use a concurrent queue to protect it with exclusion to writing when changing the status to ended (by using the barrier)


vvydier (approved):
Looks good to me. We need more use cases to test thread race conditions

PR #395, opened 2023-03-30
Eliminate build warning

kevinearls:


nachoBonafonte:
I dont think this is a proper fix, the warning is pointing to a value that is not being used with a good reason, that test is not evaluating anything, probably @bryce-b forgot adding the proper check that the test implies

bryce-b:
I've created an issue for this https://github.com/open-telemetry/opentelemetry-swift/issues/403

vvydier:
Does the tests in BatchLogRecordProcessorTests need to change? issues #403 is still open

PR #397, opened 2023-04-04
Add user agent on OTLP requests

### LINKED ISSUE #331: OTLP exporters should emit standard user agent string
As per the following change in the specification (1.14 releasing Oct 4):

> OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following:
> ```OTel OTLP Exporter Python/1.2.3```

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent

kevinearls:
This fixes #331 We now set a user agent conforming to the [OTEL spec](https://github.com/kevinearls/opentelemetry-swift/pull/new/add-user-agent) on HTTP and GRPC requests

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocol/common/Headers.swift
+++ b/Sources/Exporters/OpenTelemetryProtocol/common/Headers.swift
@@ -1,0 +1,18 @@
+//
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+// 
+
+import Foundation
+import OpenTelemetryApi
+
+public struct Headers {
+    // GetUserAgentHeader returns an OTLP header value of the form "OTel OTLP Exporter Swift/{{ .Version }}"
+    // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent
+    public static func getUserAgentHeader() -> String {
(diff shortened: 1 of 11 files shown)


Sources/Exporters/OpenTelemetryProtocol/common/Headers.swift
```diff
@@ -0,0 +1,18 @@
...
+// 
+
+import Foundation
+import OpenTelemetryApi
+
+public struct Headers {
+    // GetUserAgentHeader returns an OTLP header value of the form "OTel OTLP Exporter Swift/{{ .Version }}"
+    // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent
+    public static func getUserAgentHeader() -> String {
+        let version = OpenTelemetry.version
```
nachoBonafonte, line 13: The current string in Opentelemetry.version contains a "v" before the number, I think it should be removed before setting the value to the header

nachoBonafonte (approved):

nachoBonafonte:
Thanks a lot for the contribution

PR #398, opened 2023-04-05
Fix URLSession instrumentation in iOS 16

nachoBonafonte:
iOS 16 broke async/await URLSession instrumentation, add extra code to make it work:

Only  NSURLSessionTask.resume and URLSessionTaskDelegate.urlSession(_:task:didFinishCollecting:) are being called (contrary to what documentation says), try to move all the logic to these two methods

 Also for async/await calls without delegate set a task delegate so that the delegate method is called and we can capture it

Added many tests exercising async await network calls

### DIFF
--- a/Sources/Instrumentation/URLSession/InstrumentationUtils.swift
+++ b/Sources/Instrumentation/URLSession/InstrumentationUtils.swift
@@ -4,8 +4,10 @@
  */
 
 import Foundation
-#if !os(macOS)
+#if os(iOS) || os(tvOS)
 import UIKit
+#elseif os(watchOS)
+import WatchKit
 #endif
 
 enum InstrumentationUtils {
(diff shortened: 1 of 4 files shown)


Sources/Instrumentation/URLSession/InstrumentationUtils.swift
```diff
@@ -48,4 +51,21 @@ enum InstrumentationUtils {
...
+
+    static func usesUndocumentedAsyncAwaitMethods() -> Bool {
+#if os(macOS)
+        let os = ProcessInfo.processInfo.operatingSystemVersion
+        if os.majorVersion >= 13 {
+            return true
+        }
+#else
+        let version = UIDevice.current.systemVersion
+        if let versionNumber = Double(version),
```
Sherlouk, line 63: Would https://forums.swift.org/t/swift-concurrency-back-deployment/51908/13 be preferred to check for concurrency availability?

nachoBonafonte: It only fails in iOS 16, previous versions work with async/await perfectly

Sherlouk: wow, bizarre! good spot

vvydier (approved):

PR #399, opened 2023-04-07
Separate OTLP HTTP and GRPC Exporters

### LINKED ISSUE #389: Separate OTLP Exporter HTTP from gRPC
Currently the gRPC and HTTP exporters (using the OpenTelemetry Collector Protocol) are both in the same package. 

However, the gRPC exporter has heavy dependencies on GRPC and NIO - something which isn't true for the HTTP exporter.
In my testing, this change alone increases the compile time by many minutes in my project.

It appears we would still need the SwiftProtobuf dependency, but overall the separation here would provide an OpenTelemetry solution with very few dependencies (just Protobuf) and a much lower compile time.

Image

Zoomed out far because of just how many modules this project adds, but you can see just how many slow/red dependencies could be removed by performing this change (and us moving to HTTP rather than gRPC).

Sherlouk:
Closes #389

Created three targets: Common, Grpc, and HTTP. The latter two are also products.
The default product (same name as previous) will use Grpc to aide in backwards compatibility.

Clients using HTTP will have to change imports, but given the experimental nature of the current implementation I don't believe this to be a major problem. However, if others disagree we could make the default product include both Grpc and HTTP and then have separate GRPC and HTTP targets for those who specifically only want one.

Updated sample code which proves we're able to drop the NIO and GRPC dependencies. Not done full benchmarks, but locally the sample product does build quicker.

No other changes beyond separation and imports. Basic compile tests done, but not a full integraiton.


nachoBonafonte (approved):
Looks really good to me, thanks a lot for your contribution

nachoBonafonte:
Hi Sherlouk, after merging a previous PR (#331 ) it brought some conflicts with your code, could you resolve it? I will merge after done

Sherlouk:
Will try and find some time to rebase, little strapped for time though now as I'll be on holiday for the next week 👍

Have tested it locally though and am happy with functionality.

PR #400, opened 2023-04-10
fix: avoid ending spans multiple times

seemk:
https://github.com/open-telemetry/opentelemetry-swift/pull/379 changed the existing code such that ending a span multiple times sends it to the processor/exporter multiple times. With the previous logic this did not happen.

Added a quick bailout when ending a span and it has already ended.


nachoBonafonte (approved):
Oh, yes, good catch, the previous code was just exiting the block instead of the function, my bad

nachoBonafonte:

PR #404, opened 2023-04-19
Class simplification - 

nachoBonafonte:
Hi @bryce-b, this PR tries to simplify protocol and classes structure in the new metrics SDK branch

The current structure is quite complex and it tries to simplify a bit. Basically tries to remove all protocols that are mainly used by a single class starting with Any, and replace them by a class (which is basically renaming the Any... class by the protocol name.

I have created a PR since it looks a big change and maybe I am missing the need for those protocols in the future, so correct it or discard the change if you dont like it @bryce-b


bryce-b (approved):

bryce-b:
Good changes. I appreciate the simplification.

PR #408, opened 2023-04-28
Update Semantic Attributes

Sherlouk:
Based on Slack conversation with Nacho, this patches the current script.


Sherlouk:
Looks like CI is failing because the SDK utilises convention values which have been renamed. Just need to do a pass and rename all uses of them - if anybody gets time, feel free to take over and make those tweaks. The script itself is good here though 👍

nachoBonafonte:
I used the updated script and the template for #409. Thanks a lot for your help

PR #411, opened 2023-05-11
Add option to set the timestamp in LogRecordBuilder

emanuelschmoczer:
This PR adds the option to set the `Timestamp` for log events when using the `LogRecordBuilder`. I looked at other libraries, like the one for [Java](https://github.com/open-telemetry/opentelemetry-java/blob/47ee573f07c9025547d494608863d9314b29df4b/api/all/src/main/java/io/opentelemetry/api/logs/LogRecordBuilder.java#L30) before I started working on this PR and they allow setting both the timestamp and observed timestamp. To not change any existing functionality I went with `timestamp ?? sharedState.clock.now` in the `emit()` function, which should allow anyone who does not care about setting the timestamp manually to ignore this change.

For the tests I only explicitly covered the cases when both timestamp and observed timestamp are either set or not set. Please let me know if you prefer testing this in a different way or have any feedback for improvement.

### DIFF
--- a/Sources/OpenTelemetryApi/Logs/DefaultLogger.swift
+++ b/Sources/OpenTelemetryApi/Logs/DefaultLogger.swift
@@ -37,6 +37,9 @@ public class DefaultLogger : Logger {
     }
 
     private class NoopLogRecordBuilder : EventBuilder {
+        func setTimestamp(_ timestamp: Date) -> Self {
+            return self
+        }
 
         func setObservedTimestamp(_ observed: Date) -> Self {
             return self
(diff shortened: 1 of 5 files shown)


nachoBonafonte (approved):
LGTM, but I would like @bryce-b opinion before merging, as he is the original author of this area

bryce-b:
I believe that "observedTimestamp" is intended to satisfy the need proposed with this PR. not sure it's necessary to be able to set the timestamp specifically.

bryce-b:
If it's available in the java sdk, I don't see why we shouldn't add it here.

bryce-b (approved):

PR #423, opened 2023-06-15
Fix flaky tests for DatadogExporter

lunij:
The `EncodableValue` tests make use of a `JSONEncoder` instance without specifying any output format. Because JSON keys are unordered by default, the tests are randomly failing. The simple fix is to set the JSONEncoder's output format to `.sortedKeys`. This has no effect to the actual implementation that is being tested, but removes flakiness from the test environment.


nachoBonafonte (approved):
Good catch, thanks a lot

PR #424, opened 2023-06-15
Fix some Xcode warnings

lunij:
This PR fixes half of the Xcode warnings. The other half is due to the thrift-generated Swift 3 files.

For the `'currentRadioAccessTechnology' was deprecated in iOS 12.0` warning, I recommend to drop support for iOS 11 and iOS 12. They are old enough. iOS 12 was released in September 2018.

| macOS | iOS | watchOS |
|---|---|---|
| macOS | iOS | watchOS |

As far as I can see Jaeger is supporting the [OpenTelemetry Protocol](https://www.jaegertracing.io/docs/1.46/apis/#opentelemetry-protocol-stable) and [Protobuf via gRPC](https://www.jaegertracing.io/docs/1.46/apis/#protobuf-via-grpc-stable) ([second reference](https://www.jaegertracing.io/docs/1.46/apis/#grpcprotobuf-stable)). Do you have any plans moving away from Thrift?

### DIFF
--- a/Examples/Logs Sample/main.swift
+++ b/Examples/Logs Sample/main.swift
@@ -1,7 +1,7 @@
 //
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
-// 
+//
 
 import Foundation
 import OpenTelemetryApi
@@ -11,22 +11,26 @@ import Logging
 import GRPC
 import NIO
 
-
 func configure() {
-  let configuration = ClientConnection.Configuration.default(
-      target: .hostAndPort("localhost", 4317),
-      eventLoopGroup: MultiThreadedEventLoopGroup(numberOfThreads: 1))
+    let configuration = ClientConnection.Configuration.default(
+        target: .hostAndPort("localhost", 4317),
+        eventLoopGroup: MultiThreadedEventLoopGroup(numberOfThreads: 1)
+    )
 
-      
-  _ = OtlpLogExporter(channel: ClientConnection(configuration: configuration))
-  
-      OpenTelemetry.registerLoggerProvider(loggerProvider: LoggerProviderBuilder().with(processors: [
-        BatchLogRecordProcessor(logRecordExporter:OtlpLogExporter(channel: ClientConnection(configuration: configuration)))]).build())
-      
-  }
+    let loggerProvider = LoggerProviderBuilder()
+        .with(processors: [
+            BatchLogRecordProcessor(logRecordExporter: OtlpLogExporter(channel: ClientConnection(configuration: configuration)))
+        ])
+        .build()
 
-  configure()
-        
-  let eventProvider = OpenTelemetry.instance.loggerProvider.loggerBuilder(instrumentationScopeName: "myScope").setEventDomain("device").build()
-      
-      
+    OpenTelemetry.registerLoggerProvider(loggerProvider: loggerProvider)
+}
+
+configure()
+
+let eventProvider = OpenTelemetry
+    .instance
+    .loggerProvider
+    .loggerBuilder(instrumentationScopeName: "myScope")
+    .setEventDomain("device")
+    .build()
--- a/Examples/Logs Sample/main.swift
+++ b/Examples/Logs Sample/main.swift
@@ -18,7 +18,7 @@ func configure() {
       eventLoopGroup: MultiThreadedEventLoopGroup(numberOfThreads: 1))
 
       
-  OtlpLogExporter(channel: ClientConnection(configuration: configuration))
+  _ = OtlpLogExporter(channel: ClientConnection(configuration: configuration))
   
       OpenTelemetry.registerLoggerProvider(loggerProvider: LoggerProviderBuilder().with(processors: [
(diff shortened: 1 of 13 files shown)


Examples/Logs Sample/main.swift
```diff
@@ -18,7 +18,7 @@ func configure() {
       eventLoopGroup: MultiThreadedEventLoopGroup(numberOfThreads: 1))
 
       
-  OtlpLogExporter(channel: ClientConnection(configuration: configuration))
+  _ = OtlpLogExporter(channel: ClientConnection(configuration: configuration))
```
nachoBonafonte, line 21: I think this line should be deleted directly as the object is being recreated later in the place where it is used

Sources/Instrumentation/URLSession/URLSessionLogger.swift
```diff
@@ -147,7 +147,7 @@ class URLSessionLogger {
         }
         instrumentation.configuration.injectCustomHeaders?(&request, span)
         var instrumentedRequest = request
-        objc_setAssociatedObject(instrumentedRequest, &URLSessionInstrumentation.instrumentedKey, true, .OBJC_ASSOCIATION_COPY_NONATOMIC)
```
nachoBonafonte, line 150: did it show as a warning? what are your warning settings?

nachoBonafonte:
Yes, for Jaeger we could move away from Thrift, its being maintained in a custom fork that I modified for it to work. If it accepts OTLP natively we can remove it. Thanks for the heads up

Sources/Instrumentation/URLSession/URLSessionLogger.swift
lunij: Warning settings are the default by Xcode for Swift packages. But it might be that this warning is new, because I'm already using Xcode 15 beta.

Examples/Logs Sample/main.swift
lunij: https://github.com/open-telemetry/opentelemetry-swift/pull/424/commits/92ce53a1447e0c519c87afcb2935c3b7cb68e4b2

vvydier:
The otel collector accepts jaegar, dont think Jaegar accepts OTLP natively, as its Jaegar is not actively maintained

lunij:
> The otel collector accepts Jaegar, don't think Jaegar accepts OTLP natively, as its Jaegar is not actively maintained

I'm struggling to understand this. 

> ... otel collector accepts Jaegar ...

This sounds like Jaeger is a technology/standard to transmit data, but not a service/collector. 

> ... don't think Jaegar accepts OTLP natively ...

This sounds like you are questioning [their documentation](https://www.jaegertracing.io/docs/1.46/apis/#opentelemetry-protocol-stable), don't you? 

> ... Jaegar is not actively maintained

Where are you looking at? I'm looking at https://github.com/jaegertracing/jaeger and it looks pretty active.

When something is still confusing to me when it comes to active development then it is the thrift generator. In case Thrift needs to be supported, the thrift generator for Swift needs to be maintained. Is that open sourced somewhere? I cannot find it.

Tests/OpenTelemetrySdkTests/Logs/BatchLogRecordProcessorTests.swift
```diff
@@ -45,7 +45,7 @@ class BatchLogRecordProcessorTests : XCTestCase {
         let waitingExporter = WaitingLogRecordExporter(numberToWaitFor: maxQueueSize)
         let processor = BatchLogRecordProcessor(logRecordExporter: waitingExporter,scheduleDelay: 10000,maxQueueSize: maxQueueSize, maxExportBatchSize: maxQueueSize/2)
         let loggerProvider = LoggerProviderBuilder().with(processors: [processor]).build()
-        let logger = loggerProvider.get(instrumentationScopeName: "BatchLogRecordProcessorTest")
+        _ = loggerProvider.get(instrumentationScopeName: "BatchLogRecordProcessorTest")
```
nachoBonafonte, line 48: This is a valid warning, the test is not testing anything. The way to solve it is doing the actual test

Sources/Instrumentation/URLSession/URLSessionLogger.swift
nachoBonafonte: We should not use any non production ready Xcode version except for testing, we can expect bug and issues in it. In this case it is suspicious to me, beacuse `objc_setAssociatedObject` documentation is quite clear about the opposite

nachoBonafonte:
Yes, jaeger docs are clear, I think we could safely drop Thrift support, we will take a decision in the next SIG meeting

Tests/OpenTelemetrySdkTests/Logs/BatchLogRecordProcessorTests.swift
lunij: I saw the unfinished test, but decided to not fix it. As a compromise I replace the misleading warning with an explicit one, stating the test as unfinished.

https://github.com/open-telemetry/opentelemetry-swift/pull/424/commits/fb462d519d066841815a8dcf9d2821cd916c1911

nachoBonafonte:
Some of the warning fixes are great, but some others are based on a beta version of Xcode, please could you redo it fixing only the warnings being reported with the latest public release of Xcode

PR #425, opened 2023-06-15
Fix: Access requestMap only on the protected queue

andraskadar:
Accessing the content of the `requestMap` asynchronously on any thread can lead to concurrency issues and crashes.

These changes refactor the code without introducing changes to the behaviour to make sure that even reading of the dictionary happens on the dedicated queue.

## Motivation

We are using the automatic network tracing in our production code and after updating to the latest versions, we have noticed a growing tendency in the crashes originated from the OTel SDK. It happens rarely, but considering the size of our user base, it happens a lot and grew to become our number one cause of crashes.

Crashes started after we upgraded from 1.4.0 to 1.5.0:
Screenshot 2023-06-15 at 15 25 41
Screenshot 2023-06-15 at 15 26 47

Looking at the stack trace provided by Firebase Crashlytics (and also Xcode Organizer), the source of the crash is evident (file + line). Looking further into the code, I saw that at all other places we protect the access to this cache dictionary, hence my guess that this could only be a mistake that was introduced by accident in https://github.com/open-telemetry/opentelemetry-swift/commit/b384f0d.

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -549,21 +549,26 @@ public class URLSessionInstrumentation {
 
     private func urlSession(_ session: URLSession, task: URLSessionTask, didFinishCollecting metrics: URLSessionTaskMetrics) {
         let taskId = self.idKeyForTask(task)
-        if (self.requestMap[taskId]?.request) != nil {
-            /// Code for instrumenting colletion should be written here
-            var requestState: NetworkRequestState?
-            queue.sync {
-                requestState = requestMap[taskId]
-                if requestState != nil {
-                    requestMap[taskId] = nil
-                }
+
+        var requestState: NetworkRequestState?
+        queue.sync {
+            requestState = requestMap[taskId]
+
+            if requestState?.request != nil {
+                requestMap[taskId] = nil
             }
-            if let error = task.error {
-                let status = (task.response as? HTTPURLResponse)?.statusCode ?? 0
-                URLSessionLogger.logError(error, dataOrFile: requestState?.dataProcessed, statusCode: status, instrumentation: self, sessionTaskId: taskId)
-            } else if let response = task.response {
-                URLSessionLogger.logResponse(response, dataOrFile: requestState?.dataProcessed, instrumentation: self, sessionTaskId: taskId)
-            }
+        }
+
+        guard requestState?.request != nil else {
+            return
+        }
+
+        /// Code for instrumenting colletion should be written here
+        if let error = task.error {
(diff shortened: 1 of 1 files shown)


Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -549,22 +549,27 @@ public class URLSessionInstrumentation {
...
-                URLSessionLogger.logError(error, dataOrFile: requestState?.dataProcessed, statusCode: status, instrumentation: self, sessionTaskId: taskId)
-            } else if let response = task.response {
-                URLSessionLogger.logResponse(response, dataOrFile: requestState?.dataProcessed, instrumentation: self, sessionTaskId: taskId)
+
+        var requestState: NetworkRequestState?
+        queue.sync {
+            requestState = requestMap[taskId]
+
+            if requestState?.request != nil {
+                requestMap[taskId] = nil
```
andraskadar, line 558: Similarly to other methods, except an extra check on the `request`:
https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift#L530-L536

andraskadar:
> Just to confirm, you are not using async/await networking, and you have set a delegate, right? If not could you post an example of your network call? We think the underlying issue is not exactly a concurrency one, but we think that this callback should not be called in your situation.

We are not actively using async/await, we are generating our APIs from OpenAPI specs, using [SwagGen's 4.4.0 template](https://github.com/yonaskolb/SwagGen/tree/4.4.0/Templates/Swift) that uses Alamofire under the hood (there are example method calls on the linked page).

I do not have an exact call I can share because the crashes do not provide enough context for me to decipher. Also, if I am not mistaken, because of the nature of the method Swizzling, it could even originate from network requests that are produced by 3rd party libraries our app relies on.

> but we think that this callback should not be called in your situation

I also agree, I was surprised to see it even, we are definitely not calling it explicitly and I could find no `async` or `await` keywords even in our codebase using text search.

nachoBonafonte (approved):
LGTM. It solves the concurrent access to the map so it solves the problem. 

The root cause of this is that when a task has a delegate the same code is called twice, once in `didFinishCollecting` and once in `didCompleteWithError`. The `didFinishCollecting` was added because in iOS 16 and async network calls the `didCompleteWithError` was never called, but it ended up being called twice in cases like this. 

Your PR fixes this double call so it is great. I am planning to solve the double call, but even then, several concurrent network calls could end at the same time and your PR would also help for those cases (even when they would be much less frequent)

nachoBonafonte:
Thanks a lot for your contribution, this was a clear oversight

andraskadar:
Thank you very much! Looking forward to the next release! ❤️

PR #426, opened 2023-06-16
Update CI to run tests for tvOS

lunij:
This Swift package has tvOS support. Because of that it requires to run tests on tvOS as well to make sure nothing got broken because of some changes. This PR adds the missing tvOS job to the GitHub Workflow.


nachoBonafonte (changes requested):
LGTM, thanks a lot

nachoBonafonte (approved):

PR #427, opened 2023-06-16
Add matrix to macOS job in GitHub workflow

lunij:
This PR adds a matrix to the macOS job, enabling the testing for multiple Xcode versions, Swift versions and macOS versions.

The Swift/Xcode/macOS versions are chosen the following way: This library has two package manifests. One is stating the minimum Swift version 5.2. The other is set to 5.6. I used both of these versions and also added the latest stable Swift version 5.8 as well. The Xcode and macOS versions are chosen by the first versions that started to support the mentioned Swift versions. That can be checked [here](https://xcodereleases.com). For Swift 5.2 it should actually be Xcode 11.4 but GitHub is only supporting Xcode 11.7. Should be fine.

⚠️  Currently it showcases that it would not compile properly with older Swift versions. This needs to be either fixed or the support needs to be dropped. I'm trying my best to fix it. But the reality is that people are not using this library with the older Swift/Xcode/macOS versions. Otherwise someone would raise an issue.

URLSessionInstrumentation seems to not work properly because of macOS version availability checks at runtime that fail already at compile time. The compiler tries to type check the code but does not find the API on a too old macOS SDK version.
You can read more about that in this [Swift forum post](https://forums.swift.org/t/if-available-check-failing/64938).


lunij:
I'm still trying my best to add support for Swift 5.2. But the more I deal with that, the more I'm thinking about dropping the support. Because not even Swift 5.6 is fully supported and highly depends on newer Xcode versions, I'd like to drop the "support" for Swift 5.2. I put the word "support" into quotes because it's obviously not supported at all. 

Support for a certain Swift version on macOS means that it needs to compile on the first macOS version and the first Xcode version that are supporting the particular Swift version. The CI is proving that this is not the case.

What do you think?

PR #428, opened 2023-06-20
Remove Swift 5.2 package manifest

lunij:
#427 showcases that Swift 5.2 is actually not supported. This repository still keeps the Swift package manifest for that Swift version. The attempts to make the Swift 5.2 package manifest work failed due to the older [swift-nio version 2.39.1](https://github.com/apple/swift-nio/tree/2.39.1). That version still supports Swift 5.2 but does not contain the API that is being used in this package. Newer swift-nio versions contain the API, but do not support Swift 5.2.

The Package.resolved is git-ignored in addition because it is outdated for a long time and is actually irrelevant because it exists for the CI only. But the CI should run on the latest resolved Package.resolved to raise issues as soon as possible. In case this Swift package is used in another project as a dependency, the Package.resolved of this repository is always ignored.


nachoBonafonte:
I am really not sure about removing directly swift 5.2 tools and go directly to 5.6. This must be publicly stated in the Slack channel at least and wait for any feedback against it, and also must be talked in the SIG meeting

lunij:
Multiple issues here:

1. Not even Swift 5.6 is properly supported. So going directly from 5.2 to 5.6 is not a big deal, because both versions and all in-between are not working compiling at the moment.
2. I was looking for the Slack channel. The one I found on opentelemetry.io requires to join the "Cloud Native Computing Foundation". How to join CNCF as an individual? I'm missing a discussion place for OpenTelemetry.
3. Why does it need to be talked in the SIG meeting? All arguments are delivered already, here on GitHub. Obviously no one recognized that all Swift versions of this package from 5.2 to 5.6 are not properly working. That's the perfect time to drop them. Actually, even 5.6 could be dropped, but because this Swift version is a huge milestone (async/await, package plugins, etc.) I recommend to ensure proper support for that version.

There is one topic though, you could discuss in a meeting: The fact that you lost support for many Swift versions due to missing CI jobs that use those versions for testing. 😉 Okay, seriously now, let's not be too bureaucratic but practical instead. I'm already working on adding Swift 5.6 support back again.

nachoBonafonte:
All strategic decisions are taken in the SIG meeting which is published in the OpenTelemetry calendar and open to anyone to participate 
. The Slack channel, is also used for announces and day to day topics, doubts or community feedback, it can be found here (don't think you must be a member of the CNCF to participate): https://cloud-native.slack.com/archives/C01NCHR19SB

lunij:
Okay, I found the OpenTelemetry calendar and the Swift SIG is tomorrow. Good to know!

Joining the Slack channel looks like that for me:

|  |  |  |
|---|---|---|
| This image shows that you need to sign up for CNCF to join the Slack. | This image shows that you need to put an organization to sign-up. I do not see an option for "no organization" or skipping that step. | This is the biggest pain point. You need to pay at least $1000 to sign up. In true open source projects it should be possible to communicate without being forced to pay for the communication. |

nachoBonafonte:
Can you follow the steps here?: 
https://lists.cncf.io/g/harbor-users/message/74

lunij:
That works! Thanks for guiding me there. That was hidden.

lunij:
I don't see respect for the work I have done in several unmerged PRs. That's why I decided to discontinue the contribution on this library. Thanks!

PR #431, opened 2023-06-20
added event.data nested support

bryce-b:
This PR adds support for the `event.data` nested attributes.

### DIFF
--- a/Sources/Exporters/Jaeger/Adapter.swift
+++ b/Sources/Exporters/Jaeger/Adapter.swift
@@ -95,32 +95,35 @@ final class Adapter {
         var vBool: Bool?
         var vLong: Int64?
 
-        switch attrib {
-        case let .string(value):
-            vType = .string
-            vStr = value
-        case let .bool(value):
-            vType = .bool
-            vBool = value
-        case let .int(value):
-            vType = .long
-            vLong = Int64(value)
-        case let .double(value):
-            vType = .double
-            vDouble = value
-        case let .stringArray(value):
-            vType = .string
-            vStr = try? String(data: JSONEncoder().encode(value), encoding: .utf8)
-        case let .boolArray(value):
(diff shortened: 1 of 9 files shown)


Sources/Exporters/Jaeger/Adapter.swift
```diff
@@ -95,32 +95,35 @@ final class Adapter {
...
+        vStr = try? String(data: JSONEncoder().encode(value), encoding: .utf8)
+      case let .doubleArray(value):
+        vType = .string
+        vStr = try? String(data: JSONEncoder().encode(value), encoding: .utf8)
+      case let .set(value):
+        vType = .string
+        vStr = try? String(data: JSONEncoder().encode(value), encoding: .utf8)
+      }
         return Tag(key: key, vType: vType, vStr: vStr, vDouble: vDouble, vBool: vBool, vLong: vLong, vBinary: nil)
     }
```
lunij, line 128: Please respect the indentation being used in this project.

lunij (changes requested):

nachoBonafonte (approved):

crsantos:
Hello, 

I need help trying to understand what use cases we have for this nested attribute.

**Context:**

* I have noticed that there are platforms that do not handle this level of deep
* I have bumped into this because I was wrongly using it to surface `ios.state` as `event.data` ([see Mobile conventions](https://opentelemetry.io/docs/specs/semconv/mobile/events/#ios))
* But I was able to see the event being correctly reported because switched to [setAttributes](https://github.com/open-telemetry/opentelemetry-swift/blob/0033445b1de9516c16b75d678b0003a8b5ec1c8b/Sources/OpenTelemetrySdk/Logs/LogRecordBuilderSdk.swift#L61-L64)

I would like to understand what makes sense to be `event.data` or a regular attribute.

PR #444, opened 2023-07-24
added protobuf version as exporter user-agent version

bryce-b:


nachoBonafonte (approved):

vvydier (commented):
Do we have to update this with every release? or is there a way to pick it up from the release meta data?

bryce-b:
@vvydier only when we update the protobufs. which reminds me: I should update the docs to note this.

vvydier (approved):

PR #448, opened 2023-08-03
Fix active context is removed wrongly

nachoBonafonte:
We must only reset current context when the span being ended is the active span


vvydier (approved):
looks good!

bryce-b (approved):

nachoBonafonte:
Fixes #420

PR #450, opened 2023-08-07
added package.swift for swift 5.9 & xcode beta

bryce-b:


bryce-b:
looks like this is a duplicated PR

Issue #447, opened 2023-08-03
App crashes when URLSessionInstrumentation is used with AVAssetDownloadTask


initishbhatt:
we use `AVAssetDownloadURLSession` for downloading video assets.However, the tasks used for this `NSURLSession` subclasses as instances of `AVAssetDownloadTask`. Both of these classes declare that the properties originalRequest, currentRequest and response are unavailable.

When using the `URLSessionInstrumentation` app crashes with the following message:
 **Terminating app due to uncaught exception 'NSGenericException', reason: 'AVAssetDownloadTask does not support response property** from code (https://github.com/open-telemetry/opentelemetry-swift/blob/ff57bcc23747bc41e2ab82e57e578e80639fc0a5/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift#L538).

I'm not sure what would be the solution here. Could you please suggest some alternatives?

bryce-b:
I think it's appropriate to do a check in our instrumentation and NOOP if we detect a AVAssetDownloadTask, since it doesn't support the response attribute, it doesn't seem like this task object provides any information about the streaming content. 

It looks like you'll need to add manual instrumentation for your video streaming though.

nachoBonafonte:
Could you provide some simple code that exercise this issue?, it would help us a lot to fix it as soon as possible. 

For a quick and dirty workaround in the meantime, you can provide `delegateClassesToInstrument` to `URLSessionInstrumentationConfiguration` and only add the delegate classes that are not related to AVAssetDownloadURLSession. It looks like your code is crashing in the session delegate class code.

initishbhatt:
> Could you provide some simple code that exercise this issue?, it would help us a lot to fix it as soon as possible.
> 
> For a quick and dirty workaround in the meantime, you can provide `delegateClassesToInstrument` to `URLSessionInstrumentationConfiguration` and only add the delegate classes that are not related to AVAssetDownloadURLSession. It looks like your code is crashing in the session delegate class code.
Thanks, i'll add the example code for this

initishbhatt:
@bryce-b  not yet, just tested the main branch i am still getting  `Terminating app due to uncaught exception 'NSGenericException', reason: 'AVAssetDownloadTask does not support response property`

nachoBonafonte:
Can you point to the line/method it is crashing? Call stack would be perfect.

Thanks

initishbhatt:
```
[...]

PR #452, opened 2023-08-17
 Fix crash when auto instrumenting AVAsset download task

nachoBonafonte:
Skip tasks that are of kind of AVAssetDownloadTask. 
Most of the changes are Swift formatting changes, real fix starts in line 575


Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -573,7 +571,13 @@ public class URLSessionInstrumentation {
     }
 
     private func urlSessionTaskWillResume(_ task: URLSessionTask) {
-        let taskId = self.idKeyForTask(task)
+        // AV Asset Tasks cannot be auto instrumented, they dont include request attributes, skip them
+        if let avAssetDownloadTask = NSClassFromString("__NSCFBackgroundAVAssetDownloadTask"),
```
bryce-b, line 575: should we move this `classFromString` call  to a static const? maybe it's not very expensive, but will this be called every time this method is called? Is it necessary?

nachoBonafonte: Done!!

vvydier:
looks good as discussed on the call!

vvydier (approved):
looks good as discussed on the call

bryce-b (approved):

nachoBonafonte:
Fixes #447

Issue #449, opened 2023-08-05
forceFlush() and shutdown() operations hang, not honoring timeout when there is no network connection for whole session


dpasirst:
When attempting to either `forceFlush()` or `shutdown()` the traceProvider instance or LogRecordProcessors such that no network connection to the OpenTelemetry Collector endpoint was present from launch to exit, the respective function calls seemingly hang and do not honor the configured or specified timeouts as called for in the OpenTelemetry [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.23.0/specification/trace/sdk.md#forceflush:~:text=Shutdown,registered%20SpanProcessors.).

**Environment:**
macOS 13.4.1, OpenTelemetry endpoint is configured with https/TLS. OpenTelemetry-swift 1.5.1. Xcode building native macOS application.
Either - network is disabled on the MacBook from application launch through application exit (or network is operational but collector endpoint is unreachable from application launch through exit).

_Note:_ if the collector was reachable when the application launched and then became unreachable upon application exit, the functions appear to behave as expected.

**Simplified Example:** 
```swift
 OpenTelemetry.registerTracerProvider(tracerProvider: TracerProviderBuilder()
   .add(spanProcessor:spanProcessor)  // SimpleSpanProcessor 
   .with(resource: DefaultResources().get())
   .build())
logProcessors = [
    BatchLogRecordProcessor(
        logRecordExporter:OtlpLogExporter(channel: grpcChannel)
    )
]
OpenTelemetry.registerLoggerProvider(loggerProvider: LoggerProviderBuilder()
    .with(processors: logProcessors!)
        .with(resource: DefaultResources().get())
        .build())

...... <send some spans>
...... <send some logs>

let tracerProviderSDK = OpenTelemetry.instance.tracerProvider as? TracerProviderSdk
tracerProviderSDK?.forceFlush(timeout: 3) // <-- hangs - explicit timeout
tracerProviderSDK?.shutdown()  // <-- hangs - default timeout
logProcessors?.forEach { _ = $0.shutdown() }. // <--hangs 
```

**Actual Result:**
The application hangs and fails to exit.

**Expected result:**
[...]

Issue #457, opened 2023-09-12
Otlp http exporter is missing configuration options


bryce-b:
The Otlp Http Exporters are missing functionality for handling authentication. This is provided through an OTLPConfiguration option. This functionality is implemented in `OtlpGRPCTraceExporter`, and should be duplicated in the HTTP exporters.

PR #460, opened 2023-09-21
expanded exporter APIs with explicitTimeout on force flush and shutdown

bryce-b:
This fixes the issue where a processor is called with an explicit timeout, but the timeout isn't propogated to the exporter as it should be.


bryce-b:
This PR fixed #449 and #457/#455

nachoBonafonte (approved):

Issue #461, opened 2023-09-22
Upgrade the swift-nio-extras to 1.14 or above due to CVE security report


nahung89:
Hi OpenTelemetry folks, I hope you are doing great.

I'm experimenting with the library and recognize the latest version **1.7.0** contains the package dependency `swift-nio-extras` ver 1.10.2. This dependency has a security issue CVE-2022-3252 ([link](https://github.com/apple/swift-nio-extras/security/advisories/GHSA-773g-x274-8qmf) or [link](https://avd.aquasec.com/nvd/2022/cve-2022-3252/)).

I don't hesitate to make a PR to upgrade the version. Although I'm not quite sure how I can verify the impact. I'm seeing there are several test cases, perhaps it is the way? If so, how can I run those test cases to verify on my own?

Really appreciate for your effort for making this open sources and happy to contribute as well.

hunguyenaxon:
Close as fixed in https://github.com/open-telemetry/opentelemetry-swift/pull/466

PR #466, opened 2023-10-03
Upgrade swift-nio-extras version to fix security issue.

hunguyenaxon:
Update swift-nio-extras to latest version 1.19.1. This version fixed [CVE-2022-3252](https://github.com/advisories/GHSA-773g-x274-8qmf).

While consumers can upgrade the library directly in their project via SPM, this project should also update the Package.resolved because the security tool (e.g Trivy) just scans the dependencies' versions and raise issues accordingly.

It's a PR from #461

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -78,8 +78,8 @@
         "repositoryURL": "https://github.com/apple/swift-nio-extras.git",
         "state": {
           "branch": null,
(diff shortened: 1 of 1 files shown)


bryce-b (approved):

vvydier (approved):
LGTM

Issue #468, opened 2023-10-03
Missing the record of `DoubleHistogramMeterSdk` metric


hunguyenaxon:
In the `DoubleHistogramMeterSDK`, the storage need to save the recorded value in order to export it later. It is the same as we are doing in `LongHistogramMeterSDK`.

Current `DoubleHistogramMeterSDK` ([link](https://github.com/open-telemetry/opentelemetry-swift/blob/30c2a5af758c040fd5b0b2ce828e6d7610c6e37c/Sources/OpenTelemetrySdk/Metrics/Stable/DoubleHistogramMeterSdk.swift)):
```
public class DoubleHistogramMeterSdk : DoubleHistogram, Instrument {
    public var instrumentDescriptor: InstrumentDescriptor
    public var storage : WritableMetricStorage

    init(instrumentDescriptor : InstrumentDescriptor, storage: WritableMetricStorage) {
        self.instrumentDescriptor = instrumentDescriptor
        self.storage = storage
    }
    
    public func record(value: Double) {
        record(value: value, attributes: [String:AttributeValue]())
    }
    
    public func record(value: Double, attributes: [String : OpenTelemetryApi.AttributeValue]) {
       //TODO: implement
    }   
}
```

Reference from `LongHistogramMeterSDK` ([link](https://github.com/open-telemetry/opentelemetry-swift/blob/30c2a5af758c040fd5b0b2ce828e6d7610c6e37c/Sources/OpenTelemetrySdk/Metrics/Stable/LongHistogramMeterSdk.swift)):

```
public class LongHistogramMeterSdk: LongHistogram, Instrument {
    public var instrumentDescriptor: InstrumentDescriptor
    private var storage: WritableMetricStorage
    
    init(descriptor: InstrumentDescriptor, storage: WritableMetricStorage) {
        self.storage = storage
        self.instrumentDescriptor = descriptor
    }
    
    public func record(value: Int) {
        record(value: value, attributes: [String: AttributeValue]())
    }
    
    public func record(value: Int, attributes: [String: OpenTelemetryApi.AttributeValue]) {
        if value < 0 {
            // TODO: log error
            return
        }
        storage.recordLong(value: value, attributes: attributes)
    }
}
```

hunguyenaxon:
I fixed that in my forked repo. I'm going to make PR on that by tomorrow

PR #473, opened 2023-10-09
Allow the DoubleHistogramInstrument to record value

hunguyenaxon:
The fix is for https://github.com/open-telemetry/opentelemetry-swift/issues/468

As per in `LongHistogramMeterSdk`, we need to record the value for `DoubleHistogramInstrument` as well

### DIFF
--- a/Sources/OpenTelemetrySdk/Metrics/Stable/DoubleHistogramMeterSdk.swift
+++ b/Sources/OpenTelemetrySdk/Metrics/Stable/DoubleHistogramMeterSdk.swift
@@ -1,27 +1,27 @@
 //
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
-// 
+//
 
 import Foundation
 import OpenTelemetryApi
 
-public class DoubleHistogramMeterSdk : DoubleHistogram, Instrument {
+public class DoubleHistogramMeterSdk: DoubleHistogram, Instrument {
     public var instrumentDescriptor: InstrumentDescriptor
-    public var storage : WritableMetricStorage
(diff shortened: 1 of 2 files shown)


Sources/OpenTelemetrySdk/Metrics/Stable/DoubleHistogramMeterSdk.swift
```diff
@@ -1,26 +1,29 @@
...
     public func record(value: Double) {
-        record(value: value, attributes: [String:AttributeValue]())
+        record(value: value, attributes: [String: AttributeValue]())
     }
     
-    public func record(value: Double, attributes: [String : OpenTelemetryApi.AttributeValue]) {
-       //TODO: implement
+    public func record(value: Double, attributes: [String: OpenTelemetryApi.AttributeValue]) {
+        if value < 0 {
+            print("Histograms can only record non-negative values. Instrument \(instrumentDescriptor.name) has recorded a negative value.")
```
hunguyenaxon, line 24: The log message is following to Android version ([link](https://github.com/open-telemetry/opentelemetry-java/blob/ef3fef9f42d69243ac857c3e06dfc8b342d50039/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/SdkLongHistogram.java#L40))


```
@Override
  public void record(long value, Attributes attributes, Context context) {
    if (value < 0) {
      throttlingLogger.log(
          Level.WARNING,
          "Histograms can only record non-negative values. Instrument "
              + getDescriptor().getName()
              + " has recorded a negative value.");
      return;
    }
    storage.recordLong(value, attributes, context);
  }
```

Issue #470, opened 2023-10-03
Wrong wrapping between `DoubleHistogramMeterBuilderSdk` and `LongHistogramMeterBuilderSdk` cause missing instrumentName


hunguyenaxon:
In `DoubleHistogramMeterBuilderSdk` , there is a swap function transform to `LongHistogramMeterBuilderSdk`, but the initialization doesn't match. It causes the instrumentName become description. 

According to the `swapBuilder` function, the order need to be `instrumentName, description, unit` on both inittializations

This issue causes the LongHistorgram metric can't be sent to correctly.

```
// DoubleHistogramMeterBuilderSdk
public class DoubleHistogramMeterBuilderSdk : DoubleHistogramBuilder, InstrumentBuilder {
    ...
    init(meterProviderSharedState: inout MeterProviderSharedState, meterSharedState: inout StableMeterSharedState, name: String) {
    ...
    }

// LongHistogramMeterBuilderSdk
public class LongHistogramMeterBuilderSdk : LongHistogramBuilder, InstrumentBuilder {
    ...
    internal init(meterProviderSharedState: MeterProviderSharedState, meterSharedState: StableMeterSharedState, description: String, unit: String, instrumentName: String) {
    ...
    }

// swapBuilder
func swapBuilder<T : InstrumentBuilder>(_ builder : (MeterProviderSharedState, StableMeterSharedState, String, String, String )->T) ->T {
        return builder(meterProviderSharedState, meterSharedState, instrumentName, description, unit  )
    }
```

hunguyenaxon:
I fixed that in my forked repo. I'm going to make PR on that by tomorrow

PR #474, opened 2023-10-09
Fix the order of initialization between DoubleHistogram and LongHistogram

hunguyenaxon:
The fix is for https://github.com/open-telemetry/opentelemetry-swift/issues/470

Both kind of instrument build sdk should have the same order: `instrumentName`, `description`, `unit`. It then allows the `swapBuilder(...)` perform correctly

### DIFF
--- a/Sources/OpenTelemetrySdk/Metrics/Stable/LongHistogramMeterBuilderSdk.swift
+++ b/Sources/OpenTelemetrySdk/Metrics/Stable/LongHistogramMeterBuilderSdk.swift
@@ -1,20 +1,19 @@
 //
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
-// 
+//
 
 import Foundation
 import OpenTelemetryApi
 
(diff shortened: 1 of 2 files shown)


Sources/OpenTelemetrySdk/Metrics/Stable/LongHistogramMeterBuilderSdk.swift
```diff
@@ -1,41 +1,39 @@
...
+public class LongHistogramMeterBuilderSdk: LongHistogramBuilder, InstrumentBuilder {
     var meterProviderSharedState: MeterProviderSharedState
     
     var meterSharedState: StableMeterSharedState
     
-    var type: InstrumentType = .histogram
+    let type: InstrumentType = .histogram
     
-    var valueType: InstrumentValueType = .long
+    let valueType: InstrumentValueType = .long
```
hunguyenaxon, line 16: These properties should be constant because it ties with the class itself. 
Besides, it should apply to other properties like `instrumentName`, `description` and `unit`

vvydier (approved):

PR #479, opened 2023-10-25
Changing access control level for few methods in SpanAdapter

vj4iosdev:
### What's changed in this PR?
- Changing the access control level for the below methods from `internal` to `public` in `SpanAdapter`
1. toProtoSpanKind()
2. toProtoSpanEvent()
3. toProtoSpanLink()
4. toStatusProto()

- One minor change in `OtlpTraceJsonExporter` i,e changing the access control for the method `getExportedSpans()` to public from internal. I had contributed to this earlier but we were using this file locally. Now we are upgrading to use this directly from this repo.

### Why is this change required?
- We are leveraging OpenTelemetry extensively for both mobile and web in our app (nested as well). 
In some of the flows, the WebView generates the Spans and hands over the SpanData to mobile. In this case, the native side, instead of creating new Spans/ids, needs to override the SpanData with the received SpanData (containing already generated span ids.). So we need the above-mentioned methods to be public to do so. Hence opening the PR with the changes required.

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocolCommon/trace/SpanAdapter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolCommon/trace/SpanAdapter.swift
@@ -67,7 +67,7 @@ public struct SpanAdapter {
     return protoSpan
   }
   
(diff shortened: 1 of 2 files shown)


vj4iosdev:
Closing this PR and will open another one due to EasyCLA missing ID commit issue.

PR #480, opened 2023-10-26
Changing access control level for few methods in SpanAdapter

vj4iosdev:
### What's changed in this PR?
- Changing the access control level for the below methods from `internal` to `public` in `SpanAdapter`
1. toProtoSpanKind()
2. toProtoSpanEvent()
3. toProtoSpanLink()
4. toStatusProto()

- One more minor change done in `OtlpTraceJsonExporter` i,e changing the access control for the method `getExportedSpans()` to public from internal. I had contributed to this earlier but we were using this file locally. Now we are upgrading to use this directly from this repo.

### Why is this change required?
- We are leveraging OpenTelemetry extensively for both mobile and web in our app (nested as well). 
In some of the flows, the WebView generates the Spans and hands over the SpanData to mobile. In this case, the native side, instead of creating new Spans/ids, needs to override the SpanData with the received SpanData (containing already generated span ids.). So we need the above-mentioned methods to be public to do so. Hence opening the PR with the changes required.

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocolCommon/trace/SpanAdapter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolCommon/trace/SpanAdapter.swift
@@ -67,7 +67,7 @@ public struct SpanAdapter {
     return protoSpan
   }
   
-  static func toProtoSpanKind(kind: SpanKind) -> Opentelemetry_Proto_Trace_V1_Span.SpanKind {
+  public static func toProtoSpanKind(kind: SpanKind) -> Opentelemetry_Proto_Trace_V1_Span.SpanKind {
     switch kind {
     case .internal:
       return Opentelemetry_Proto_Trace_V1_Span.SpanKind.internal
@@ -82,7 +82,7 @@ public struct SpanAdapter {
     }
   }
   
-  static func toProtoSpanEvent(event: SpanData.Event) -> Opentelemetry_Proto_Trace_V1_Span.Event {
+  public static func toProtoSpanEvent(event: SpanData.Event) -> Opentelemetry_Proto_Trace_V1_Span.Event {
     var protoEvent = Opentelemetry_Proto_Trace_V1_Span.Event()
     protoEvent.name = event.name
     protoEvent.timeUnixNano = event.timestamp.timeIntervalSince1970.toNanoseconds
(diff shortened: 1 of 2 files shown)


vj4iosdev:
@nachoBonafonte @bryce-b @vvydier Please take a look whenever you get some time.

nachoBonafonte:
It looks good to me, I completely agree with the `getExportedSpans()` being public,  but I don't understand why you need the SpanAdapter utility functions to be public instead of modifying the `SpanData` and calling the default `toProtoSpan()`

vj4iosdev:
> It looks good to me, I completely agree with the `getExportedSpans()` being public, but I don't understand why you need the SpanAdapter utility functions to be public instead of modifying the `SpanData` and calling the default `toProtoSpan()`

Ok. So here is our use case. When we receive spanData from web FCI to mobile FCI, we are required not to generate the spanID, traceID, parentSpanID and span attributes but override these as received from the web.
We need to override `toProtoSpan(spanData: SpanData)` method in SpanAdapter in order to override the spanID, traceID, parentSpanID and span attributes but the SpanAdapter being struct we cannot mark the method  `toProtoSpan(spanData: SpanData)` as `open` to enable it for the override. 

So here is our code:

 ```
static func toProtoSpan(spanData: SpanData) -> Opentelemetry_Proto_Trace_V1_Span {
        
        if OverrideProtoSpan.containsOverridenAttributes(spanData: spanData) {
            // we override spanData here
            return OverrideProtoSpan.toProtoSpan(spanData: spanData)
        }

        // then the rest of the code as is..
}
```
And in the `OverrideProtoSpan.toProtoSpan(spanData: spanData)` we use the utility functions of `SpanAdapter` to override the traceID, spanID, parentSpanID and span attributes received from the web FCI. 
Hope this clarifies.

nachoBonafonte:
Thanks for the clarification it makes total sense now. If you think that making span adapter a class with the open method works better for you, fellow free to change it.

nachoBonafonte (approved):

vj4iosdev:
> Thanks for the clarification it makes total sense now. If you think that making span adapter a class with the open method works better for you, fellow free to change it.

Thanks for the approval. Static declarations are implicitly 'final' so may have to make changes in many places if I do that. So I will have this merged as of now and open a new one later.

PR #487, opened 2023-11-30
Fix system version  comparison in iOS and watchOS

### LINKED ISSUE #483: Is async-based method URLSession.shared.data supported?
Telemetry data is generated when using **URLSession.shared.dataTask**, but not when using async-based method **URLSession.shared.data**. Is this not supported? And if not, is it on the road map? Is there an ETA?

nachoBonafonte:
Fix system version comparison in iOS and watchOS. It was failing when version included more than two numbers.

Fixes #483


nachoBonafonte:
This solution is wrong, don't merge

Issue #495, opened 2023-12-19
Feature: Slow / Freeze render detection


Archish27:
Need similar implementation for iOS slow & freeze render detection. The same exists in [otel-android](https://github.com/open-telemetry/opentelemetry-android/tree/main/instrumentation/src/main/java/io/opentelemetry/android/instrumentation/slowrendering)

ddas09:
We need this as well for our use case. Something similar like the android one - https://github.com/open-telemetry/opentelemetry-android/tree/main/instrumentation/slowrendering

bryce-b:
Let's review if MetricKit instrumentation solves this?

PR #496, opened 2023-12-19
:sparkle: added slow frozen rendering instrumentation

Archish27:
feature : #495


bryce-b:
Can you also add a simple test for this?

bryce-b (changes requested):

Sherlouk:
Does this need adding to the Package.swift too? Looks like it's just loose in Instrumentation/ where others have distinct products.

bryce-b:
@Archish27 This is a good addition, but we have some feedback we'd like you to address.

vvydier:
@Archish27 Any update on this ?

PR #498, opened 2023-12-20
Fixed removeContextValue function and added a Test testRemoveContextValuesFromSpan for it

vvydier:
I tested this with a span and added it to 2 parent spans in different threads. Total of 5 contexts getting removed in this case.
Tested with a crude mechanism (using sleep) for now. Will improve it with a Semaphore or async/await later.

### DIFF
--- a/Tests/OpenTelemetryApiTests/Context/ActivityContextManagerTests.swift
+++ b/Tests/OpenTelemetryApiTests/Context/ActivityContextManagerTests.swift
@@ -318,11 +318,11 @@ class ActivityContextManagerTests: XCTestCase {
         XCTAssert(ActivityContextManager.instance.getCurrentContextValue(forKey: .span) === span1)
         
         //Add it to one parent in one thread
-        let parent1 = defaultTracer.spanBuilder(spanName: "parent1").startSpan()
-        ActivityContextManager.instance.setCurrentContextValue(forKey: .span, value: parent1)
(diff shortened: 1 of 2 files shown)


bryce-b (approved):

nachoBonafonte (approved):

Tests/OpenTelemetryApiTests/Context/ActivityContextManagerTests.swift
```diff
@@ -308,6 +308,45 @@ class ActivityContextManagerTests: XCTestCase {
...
+        DispatchQueue.global().async {
+            let activeSpan = ActivityContextManager.instance.getCurrentContextValue(forKey: .span)
+            XCTAssert(activeSpan === parent2)
+            ActivityContextManager.instance.setCurrentContextValue(forKey: .span, value: span1)
+            parent2.end()
+        }
+        
+        //remove all the contexts from the span and check if the Context is nil
+        sleep(1)
+        ActivityContextManager.instance.removeContextValue(forKey: .span, value: span1)
```
bryce-b, line 346: end span1 and verify it's removed from all contexts, instead of manually removing it.

vvydier: replaced the manual removing of Context with span1.end()

PR #499, opened 2023-12-20
Expands LogRecord to allow "anyValue" in body parameter.

bryce-b:


vvydier:
That looks like the right way to do it!  I created the spec-change branch after I modified the ActivityContextManager code and test case in my main, so it also shows up here. Do you think it would be ok to merge this or do we need to remove the ActivityContextManager updates from this PR?

vvydier (approved):
Looks good to me!

Sources/OpenTelemetryApi/Context/ActivityContextManager.swift
```diff
@@ -68,14 +68,12 @@ class ActivityContextManager: ContextManager {
...
-           currentValue === value
-        {
-         contextMap[activityIdent]?[key.rawValue] = nil
-          if contextMap[activityIdent]?.isEmpty ?? false {
-            contextMap[activityIdent] = nil
+        
+        for (activityKey, activity) in contextMap where value === activity[key.rawValue] {
+            contextMap[activityKey]?[key.rawValue] = nil
+          if contextMap[activityKey]?.isEmpty ?? false {
+            contextMap[activityKey] = nil
```
nachoBonafonte, line 76: These changes don't look related to this PR

Tests/OpenTelemetryApiTests/Context/ActivityContextManagerTests.swift
```diff
@@ -308,6 +308,45 @@ class ActivityContextManagerTests: XCTestCase {
         XCTAssert(OpenTelemetry.instance.contextProvider.activeSpan === nil)
     }
 
+
```
nachoBonafonte, line 311: this also looks like code not related to these changes

nachoBonafonte (changes requested):
LGTM, but there are some changes that I think are part of another PR from @vvydier

nachoBonafonte (approved):

PR #524, opened 2024-03-07
Update attribute processor for stable metrics

yangjian:
Previously AttribtueProcessor is a class and not open. It can't be customized externally. Now it's changed to procotol and its static functions are moved into SimpleAttributeProcessor (now it's public).

Also duplicated NoopAttributesProcessor is removed and more tests are added.


nachoBonafonte:
I would like to understand the use case for this change, whyy the need to inherit AttributeProcessor instead of extending it

yangjian:
There're common labels for all metrics in our iOS app. Attribute processor can be used to merge them into final labels. But one of these labels can be changed by user or server. Customized attribute processor can be more flexible to support it. Something like this:
```
class GlobalSharedLabels: AttributeProcessor {
    var attributes: ManagedAtomic<AtomicContainer<[String: AttributeValue]>>

    public updateLabels(...) {
        attributes.store(...)
    }

    public func process(incoming: [String: AttributeValue]) -> [String: AttributeValue] {
        incoming.merging(attributes.load(...), uniquingKeysWith: { $1 })
    }
}
```

nachoBonafonte (approved):
L>gym, I had some doubts at the beginning but keeping the base code as part of the protocol with an extension makes total sense. Thanks a lot for your contribution,

PR #525, opened 2024-03-11
Implement Exponential Histgoram metrics

jhoongo:
This PR implements `ExponentialHistogram` metric in Swift SDK. Most of the implementations was inspired and driven by https://github.com/open-telemetry/opentelemetry-java SDK. It has been validated with Java SDK to ensure that it generates the same results and passes the same tests for the most part.

There might be some missing area that I might have missed on this, but this should be good enough for us to starting point for supporting Exponential Histogram. 

I have excluded the implementation of Exponential Histogram on the old metric and only implemented it for Stable Metrics.

~~Note: I just notice the spacing/indentation issue on this PR, will follow up shortly on this.~~ Fixed the indentation issues and added few more testings


Sources/OpenTelemetrySdk/Metrics/Stable/Aggregation/AdaptingCircularBufferCounter.swift
```diff
@@ -0,0 +1,88 @@
...
+    func copy(with zone: NSZone? = nil) -> Any {
+        let copy = AdaptingCircularBufferCounter(maxSize: maxSize)
+        copy.startIndex = startIndex
+        copy.endIndex = endIndex
+        copy.baseIndex = baseIndex
+        copy.backing = backing.copy() as! AdaptingIntegerArray
+        return copy
+    }
+
+    private static let NULL_INDEX: Int = Int.min
```
nachoBonafonte, line 18: Please use Swift style names even  for constants (no capitals and no underscore even for constants. If you want something more visual maybe you can wrap it in an enum like:
```
 enum Index {
 case null = -1
}
```
And use it like `public private(set) var endIndex = Index.null`

Sources/OpenTelemetrySdk/Metrics/Stable/Aggregation/AdaptingCircularBufferCounter.swift
```diff
@@ -0,0 +1,88 @@
...
+    private var backing: AdaptingIntegerArray
+    private let maxSize: Int
+
+    init(maxSize: Int) {
+        self.backing = AdaptingIntegerArray(size: maxSize)
+        self.maxSize = maxSize
+    }
+
+    @discardableResult func increment(index: Int, delta: Int64) -> Bool{
+        if self.baseIndex == Int.min {
```
nachoBonafonte, line 31: Avoid using self when not needed (e.g. an init when the parameters matches the name )

Sources/OpenTelemetrySdk/Metrics/Stable/Data/Internal/Base2ExponentialHistogramIndexer.swift
```diff
@@ -1,76 +1,74 @@
...
-//    }
-//}
+
+import Foundation
+import OpenTelemetryApi
+
+public class Base2ExponentialHistogramIndexer {
+    private static var cache = [Int: Base2ExponentialHistogramIndexer]()
+    private static var cacheLock = Lock()
+    private static let LOG_BASE2_E = 1.0 / log(2)
```
nachoBonafonte, line 12: same as before with constant names

Sources/OpenTelemetrySdk/Metrics/Stable/Data/Internal/ExponentialHistogramPointData.swift
```diff
@@ -1,47 +1,38 @@
...
+    public var count: Int
+    public var zeroCount: Int64
+    public var hasMin: Bool
+    public var hasMax: Bool
+    public var min: Double
+    public var max: Double
+    public var positiveBuckets: ExponentialHistogramBuckets
+    public var negativeBuckets: ExponentialHistogramBuckets
+
+    public init(scale: Int, sum: Double, zeroCount: Int64, hasMin: Bool, hasMax: Bool, min: Double, max: Double, positiveBuckets: ExponentialHistogramBuckets, negativeBuckets: ExponentialHistogramBuckets, startEpochNanos: UInt64, epochNanos: UInt64, attributes: [String: AttributeValue], exemplars: [ExemplarData]) {
```
nachoBonafonte, line 22: use one parameter per line for these kind of long functions

nachoBonafonte (changes requested):
Really great work, just some minor stylistic issues.

jhoongo:
> Really great work, just some minor stylistic issues.

Thanks for the review @nachoBonafonte ! I've updated the PR with your feedback and rebase it to the latest main. Please let me know if you still have any further feedback or concern. Thanks,

nachoBonafonte (approved):

PR #527, opened 2024-04-01
Use set -e in shell script to detect errors

johnbley:
Noticed that `generate.sh` was not using `set -e`; let's fix that.


nachoBonafonte (approved):
LGTM, thanks.

PR #528, opened 2024-04-01
Adds a way to generate XCFramework that can be used for releases

tapthaker:
I am not aware of all the packages that should be generated for releases. We needed `OpenTelemetryApi`, but it should be easy to extend it for other packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,3 +69,8 @@ test-without-building-tvos:
 .PHONY: test-without-building-watchos
 test-without-building-watchos:
 	set -o pipefail && xcodebuild $(XCODEBUILD_OPTIONS_WATCHOS) test-without-building | xcbeautify
+
+
+.PHONY: generate-xcframework
+generate-xcframework:
+	./Scripts/generate_xcframework.sh OpenTelemetryApi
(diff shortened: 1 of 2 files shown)


heckj:
Just a head's up - if you want to use an XCFramework in an app deliverable through the App Store, Apple will require that it is both _signed_ and includes a _privacy manifest_.

nachoBonafonte:
We think it is a great start, but to be really useful for the project we would need it to allow selecting which targets of the project to include and also to be able to select the platforms/destinations to build for. Then it would be a great tool for any user. With the functionality as is the audience for this change is too narrow. Would you mind expanding it?

nachoBonafonte:
Hi @tapthaker, are you interested in improving the PR with the needs exposed?

tapthaker:
I missed your previous comment. Yes I am interested in modifying this PR. I'll get back to you

n-ravichandran:
@tapthaker Does this approach work for packages with external dependencies or resources, like the exporters that depend on `swift-protofbuf` and `swift-nio`?

bryce-b:
I'm going to close this, as it's pretty stale. If the requested changes are made in the future, please reopen!

PR #535, opened 2024-04-25
[WIP] Bridge: Logging Bridge for SwiftLog

khushijain21:
Logging Bridge for Apple's SwiftLog https://github.com/open-telemetry/opentelemetry-swift/issues/517

Some Notes:
Basic Structure - LogHandler implements
- log Level - sets log level on the handler (Default is INFO)
- Metadata - metadata on the log handler
- func log (...params) - converts logger data to OTel format and emits it

TODO: Write tests for the bridge


khushijain21:
Hello Reviewers, can you please take a look at the approach? Your feedback is appreciated. Thanks @nachoBonafonte @bryce-b @vvydier

Sources/Bridges/OTelSwiftLog/LogHandler.swift
```diff
@@ -0,0 +1,188 @@
...
+public struct InstrumentationScope{
+    public var name: String
+    public var version: String?
+    public var eventDomain: String?
+    public var schemaUrl: String?
+    public var includeTraceContext: Bool?
+    public var attributes: [String: AttributeValue]?
+}
+
+// Define a custom log handler
```
nachoBonafonte, line 19: Documentation comments are written with a triple ///, it also has a format for parameters and return values. Swift can parse this comments to create documentation

Sources/Bridges/OTelSwiftLog/LogHandler.swift
```diff
@@ -0,0 +1,188 @@
...
+struct OTelLogHandler: LogHandler {
+    private var scope: InstrumentationScope // Property to store instrumentation scope name
+    private var loggerProvider : LoggerProvider  // Property to set LoggerProvider
+    private var logger: OpenTelemetryApi.Logger 
+
+    // Define the log level for this handler
```
nachoBonafonte, line 25: Is this the default level? can it be changed temporary for a specific log? These things should be part of the commentary

Sources/Bridges/OTelSwiftLog/LogHandler.swift
```diff
@@ -0,0 +1,188 @@
...
+    private var _logLevel: Logging.Logger.Level = .info
```
nachoBonafonte, line 26: Why a private and a public that overwrites it?

Sources/Bridges/OTelSwiftLog/LogHandler.swift
```diff
@@ -0,0 +1,188 @@
...
+    public subscript(metadataKey key: String) -> Logging.Logger.Metadata.Value? {
+        get {
+            return self.metadata[key]
+        }
+        set {
+            self.metadata[key] = newValue
+        }
+    }
+    
+    // should default logger provider be a noop? or the sdk implementation?
```
nachoBonafonte, line 47: I would use the default one, if initialized is because the user wants to use it.

Sources/Bridges/OTelSwiftLog/LogHandler.swift
```diff
@@ -0,0 +1,188 @@
...
+        loggerProvider: LoggerProvider = OpenTelemetrySdk.LoggerProviderSdk()) {
+
+        self.scope = scope
+        self.loggerProvider = loggerProvider
+        let loggerBuilder = self.loggerProvider.loggerBuilder(instrumentationScopeName: scope.name)
+            .configure(with: scope)
+        self.logger = loggerBuilder.build()
+    }
+
+    public func log(level: Logging.Logger.Level,
```
nachoBonafonte, line 58: Shouldn't the level be optional, as there is a default one? Or should it be mandatory and remove the default one?

nachoBonafonte (commented):
This is looking really good, and the documentation is helping a lot. I added some comments about the API and consistency. Let me know what do you think.

Sources/Bridges/OTelSwiftLog/LogHandler.swift
khushijain21: The log level defined above is for the LogHandler whereas this is the log level with which the message was logged

```
var logger = Logger(label: "com.example.myapp")
logger.logLevel = .info  // the level of of the handler
logger.handler = otelLogHandler
logger.debug("This is a debug message")  // this log message will not be printed as .debug  is lower than .info

logger.info("Logging with additional metadata") // this will be printed
```

Sources/Bridges/OTelSwiftLog/LogHandler.swift
khushijain21: Yes the default level is .info. It can be changed like this 

```
var logger = Logger(label: "com.example.myapp")
logger.logLevel = .info  // the level of of the handler
```

Sources/Bridges/OTelSwiftLog/LogHandler.swift
khushijain21: It is so that the user can change the default log level.

Sources/Bridges/OTelSwiftLog/LogHandler.swift
nachoBonafonte: Add this information to the commentary of the parameter

Sources/Bridges/OTelSwiftLog/LogHandler.swift
nachoBonafonte: Then make it directly public

Sources/Bridges/OTelSwiftLog/LogHandler.swift
nachoBonafonte: Then this should be optional: `Logging.Logger.Level?`

nachoBonafonte (changes requested):
Just missing some tests for the code and I would consider it ready

Sources/Bridges/OTelSwiftLog/LogHandler.swift
khushijain21: even if the user does not set it explicitly, a default value is used by `Logger` to determine the logging level - here
https://github.com/apple/swift-log/blob/5d66f7ba25daf4f94100e7022febf3c75e37a6c7/Sources/Logging/Logging.swift#L128

Sources/Bridges/OTelSwiftLog/LogHandler.swift
```diff
@@ -0,0 +1,188 @@
+import Foundation
+import OpenTelemetryApi
+import OpenTelemetrySdk
+import Logging
+
+// let the bridgename be the url of the package?
+let bridgename: String = "OTelSwiftLog"
+let version: String = "1.0.0"
+
+public struct InstrumentationScope{
```
bryce-b, line 10: InstrumentationScope is a pre-existing construct in OpenTelemetry and adding this struct is confusing. The loggerBuilder provides all these options, which should be used rather than extending the builder.

Sources/Bridges/OTelSwiftLog/LogHandler.swift
```diff
@@ -0,0 +1,188 @@
...
+        .emit()
+    }
+
+
+
+}
+
+extension LoggerBuilder {
+    @discardableResult
+    func configure(with instrumentationScope: InstrumentationScope) -> LoggerBuilder {
```
bryce-b, line 99: let's not do this.

Sources/Bridges/OTelSwiftLog/LogHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+    case .dictionary(let nestedDictionary):
+        // If value is a nested dictionary, recursively convert it
+        var nestedAttributes: [String: AttributeValue] = [:]
+        for (nestedKey, nestedValue) in nestedDictionary {
+            nestedAttributes[nestedKey] = convertToAttributeValue(nestedValue)
+        }
+        return AttributeValue.set(AttributeSet(labels: nestedAttributes))
+    case .array(let nestedArray):
+        // If value is a nested array, recursively convert it
+        let nestedValues = nestedArray.map { convertToAttributeValue($0) }
```
bryce-b, line 157: `ConvertToAttributeValue` returns a `AttributeValue`, thus the `nestedValues` variable is of the type `[AttributeValue]`. All the subsequent checks are done for arrays of literals, thus this will always return nil.

Sources/Bridges/OTelSwiftLog/LogHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+        get {
+            return self.metadata[key]
+        }
+        set {
+            self.metadata[key] = newValue
+        }
+    }
+    
+    // should default logger provider be a noop? or the sdk implementation?
+    public init(scope: InstrumentationScope = InstrumentationScope(name: bridgename, version: version),
```
bryce-b, line 48: does the instrumentation scope actually need to be settable?  Is the idea here if someone wants to use this bridge but wants to give it a different instrumentation name? Is that a necessary use case?

Sources/OpenTelemetryApi/Common/AttributeValue.swift
```diff
@@ -11,10 +11,7 @@ public enum AttributeValue: Equatable, CustomStringConvertible, Hashable {
   case bool(Bool)
   case int(Int)
   case double(Double)
-  case stringArray([String])
```
bryce-b, line 14: My concern is this breaking change. Should I try to retain these attributes and have them translate to AttributeArray under the hood, or do we rev the main version indicating a breaking change? @nachoBonafonte

nachoBonafonte (approved):

khushijain21:
Sorry, was not able to work on this again. There is just one thing - the bridge may not want to depend on the OpenTelemetrySDK

bryce-b:
closing PR due to contamination of upstream repo. I'll reopen with another branch.

bryce-b:
@khushijain21 I removed the dependency on the SDK in the new PR

PR #546, opened 2024-06-08
Add Linux support via a TaskLocalContextManager

semicoleon:
This is another attempt at adding Linux support (and closing https://github.com/open-telemetry/opentelemetry-swift/issues/38) after https://github.com/open-telemetry/opentelemetry-swift/pull/438 turned out to be unworkable for merging.

I have attempted to avoid changes to the core library as much as possible, instead using an additional library `OpenTelemetryConcurrency` (better name suggestions are welcome) which uses some wrapper types, and the split out `SpanBuilderBase` and `SpanBase` protocols to discourage use of old style APIs like `setActive` and `end(time:)` which don't behave as expected when using `TaskLocal`'s to carry context around for users who have elected to use the wrapper library. 

Spliting `SpanBuilder` and `Span` into two protocols each is the most invasive change in this PR. I chose to do it that way to avoid having to maintain all of the `SpanBuilderBase` and `SpanBase` methods on a wrapper type, but if that split would be a semver hazard (or just not desirable to the team) I can switch to using a wrapper type for those like I did for `TracerProvider` and `Tracer`.

Using a wrapper library does require exposing some API's outside of `OpenTelemetryApi`. I elected to use the `package` keyword rather than make the public, but that means `OpenTelemetryConcurrency` can't be used on versions of swift < 5.9. This is also something that could be adjusted if that's not desirable.


bryce-b:
can you add a GH action that runs the build & tests on linux?

Sources/OpenTelemetryTestUtils/OpenTelemetryTestCase.swift
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+@testable import OpenTelemetryApi
+@testable import OpenTelemetrySdk
+import XCTest
+#if canImport(_Concurrency)
+@testable import OpenTelemetryConcurrency
```
bryce-b, line 10: This is not included in the Package.swift, causing build errors

semicoleon:
@bryce-b I believe this is ready again. I had to debug a bizarre issue that was appearing in the Linux job, but it seems to be resolved now

Examples/Logging Tracer/LoggingTracer.swift
```diff
@@ -69,5 +69,23 @@ class LoggingTracer: Tracer {
         func setActive(_ active: Bool) -> Self {
             return self
         }
+
+        func withActiveSpan<T>(_ operation: (any SpanBase) throws -> T) rethrows -> T {
```
bryce-b, line 73: should this be in an `#else` to the `#if canImport(_Concurrency)`? It seems like this my result in duplicate symbols when _Concurrency is imported?  Or does the fact it's async cause it not to conflict?

bryce-b (approved):

PR #551, opened 2024-06-30
Allow Prometheus exporter to be used without HTTP server

CallumTodd7:
Currently the only way to use the Prometheus exporter is via `PrometheusExporterHttpServer`. 

`PrometheusExporterHttpServer` calls `PrometheusExporterExtensions.writeMetricsCollection` to do the heavy lifting of building the response payload string. Allowing library users access to this function would allow for different/existing HTTP server frameworks to be used.


nachoBonafonte (approved):

vvydier (approved):

nachoBonafonte:
LGTM, thanks for your contribution.

PR #554, opened 2024-07-10
[WIP] Fixed concurrency crash

mamunto:
Inside `getAggregatorHandle` function we are just not doing read, there is a write as well. Here is code that's doing write operation:
```
            aggregatorHandles[processedAttributes] = newHandle
            aggregatorHandle = newHandle
```
Concurrency queue could be overkill here, better to use a serial queue with sync or async (we will need more code changes for return) is much better option.

### DIFF
--- a/Sources/OpenTelemetrySdk/Metrics/Stable/State/SynchronousMetricStorage.swift
+++ b/Sources/OpenTelemetrySdk/Metrics/Stable/State/SynchronousMetricStorage.swift
@@ -20,7 +20,7 @@ public class SynchronousMetricStorage: SynchronousMetricStorageProtocol {
     var aggregatorHandles = [[String: AttributeValue]: AggregatorHandle]()
     let attributeProcessor: AttributeProcessor
     var aggregatorHandlePool = [AggregatorHandle]()
-    private let aggregatorHandlesQueue = DispatchQueue(label: "org.opentelemetry.SynchronousMetricStorage.aggregatorHandlesQueue", attributes: .concurrent)
+    private let aggregatorHandlesQueue = DispatchQueue(label: "org.opentelemetry.SynchronousMetricStorage.aggregatorHandlesQueue")
 
     
     static func empty() -> SynchronousMetricStorageProtocol {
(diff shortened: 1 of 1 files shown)


nachoBonafonte:
Yes, you are true, it was an oversight and there is indeed a potential crash. If "get" cannot run concurrently is better to use serial queue with sync, as is basically what is doing it now using barrier in both. Would you mind doing that change instead?

mamunto:
> Yes, you are true, it was an oversight and there is indeed a potential crash. If "get" cannot run concurrently is better to use serial queue with sync, as is basically what is doing it now using barrier in both. Would you mind doing that change instead?

Yes, I can do but wanted some feedback from the team incase there is any other thoughts. Will push the change.

Sources/OpenTelemetrySdk/Metrics/Stable/State/SynchronousMetricStorage.swift
```diff
@@ -20,7 +20,7 @@ public class SynchronousMetricStorage: SynchronousMetricStorageProtocol {
     var aggregatorHandles = [[String: AttributeValue]: AggregatorHandle]()
     let attributeProcessor: AttributeProcessor
     var aggregatorHandlePool = [AggregatorHandle]()
-    private let aggregatorHandlesQueue = DispatchQueue(label: "org.opentelemetry.SynchronousMetricStorage.aggregatorHandlesQueue", attributes: .concurrent)
+    private let aggregatorHandlesQueue = DispatchQueue(label: "org.opentelemetry.SynchronousMetricStorage.aggregatorHandlesQueue")
```
mamunto, line 23: Made the queue serial

Sources/OpenTelemetrySdk/Metrics/Stable/State/SynchronousMetricStorage.swift
```diff
@@ -76,7 +76,7 @@ public class SynchronousMetricStorage: SynchronousMetricStorageProtocol {
         
         var points = [PointData]()
         
-        aggregatorHandlesQueue.sync(flags: .barrier) {
+        aggregatorHandlesQueue.sync {
```
mamunto, line 79: I didn't do any load test here. Any concern here will require to convert to async but that will trigger more API changes

nachoBonafonte (approved):
That looks perfect, thanks a lot

mamunto:
> That looks perfect, thanks a lot

Please let me know when we can get this PR merge and possibly a release

PR #579, opened 2024-07-24
Fix putHttpStatusCode(statusCode: Int, reasonPhrase: String)

HannesSteiner:
The reasonPhrase was never used.


bryce-b (approved):

bryce-b:
Looks good to me, but you'll need to set up easyCLA before we can merge it

Issue #552, opened 2024-07-08
When using network instrumentation, network background tasks failes as the reported image.


lelinolino:
This is my code where I initialize the instrumentation:
```swift
_ = URLSessionInstrumentation(configuration: URLSessionInstrumentationConfiguration( shouldInstrument: { request in
            if request.url?.absoluteString.contains( "something" ) ?? false {
                return false
            }
            return true
        }, nameSpan: { request in
            if request.url?.path().contains("something") ?? false{
                return "name"
            }
            return request.url?.path()
        }))
```

<img width="1653" alt="Screenshot 2024-07-08 alle 17 13 57" src="https://github.com/open-telemetry/opentelemetry-swift/assets/7314598/f1a39ecb-dac7-4d55-b2c2-c49ccf5d6f2b">

atreat:
Looks like using the [`URLSessionConfiguration.identifier`](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1408987-identifier) is the best way to determine if we're using a background configuration. We'll need to check this before setting the URLSessionDelegate

lelinolino:
<img width="1701" alt="Screenshot 2024-09-20 alle 00 32 06" src="https://github.com/user-attachments/assets/36944953-82bf-40b6-957a-81781f6ee448">

I tested commit **b22dd411f9381ae1ea6f3d750784b8c03a8d7810**.
Unfortunately I have the same behaviour. Attached the line of the **URLSessionInstrumentation class** where the lib fails.

Thank you so much

PR #597, opened 2024-08-22
Avoid instrumenting background async tasks

nachoBonafonte:
A background task does not support assigning a delegate and crashes when added, we avoid instrumenting them


vvydier (approved):
look good, this check will avoid crashes!

nachoBonafonte:
Fixes #552

bryce-b (approved):

PR #599, opened 2024-09-03
Add `recordException`

### LINKED ISSUE #117: add recordException() to span
should capture Error, NSError, and NSException

brunnoferreira:
This pull request aims to resolve #117.

Since Swift does not offer a superclass like [Throwable in Java](https://developer.android.com/reference/java/lang/Throwable) to group all exception types, one challenge facing these changes is a polluted API, with a long list of methods to encompass `NSException`, `NSError` and `Error`.

To work around that problem I have chosen to include a new protocol named `SpanException`, describing the basic properties required to create a meaningful exception event. That, alongside extensions that automatically conform both `NSError` (and, automatically, `Error`) and `NSException`, should cater for most use cases. For everything else, users should simply need to conform their custom types to `SpanException`.

If there is any feedback on this, please let me know – happy to update as we go.

### DIFF
--- a/Examples/Logging Tracer/LoggingSpan.swift
+++ b/Examples/Logging Tracer/LoggingSpan.swift
@@ -72,6 +72,22 @@ class LoggingSpan: Span {
         Logger.log("Span.addEvent(\(name), attributes:\(attributes), timestamp:\(timestamp))")
     }
 
+    public func recordException(_ exception: SpanException) {
+        Logger.log("Span.recordException(\(exception)")
+    }
+
+    public func recordException(_ exception: SpanException, timestamp: Date) {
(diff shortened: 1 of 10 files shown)


vvydier (approved):
This is good addition to the API to capture Exceptions. Thanks for adding the tests. Ran through them, looks good to me

brunnoferreira:
Thanks for the review, @vvydier!

I just pushed a couple of minor changes based on the results of the workflow, if you could please run it again when convenient.

One of my tests was flaky, so I got it fixed. Also turns out `NSException` is not available on Linux despite being part of Foundation, so I made sure we don't try compiling it there.

PR #601, opened 2024-09-11
Export max and min in Metric exporter for HistogramDataPoint

### LINKED ISSUE #594: Exported metric data has a count and sum but no min and max value
I wanted to post this as a discussion since I don't know for sure whether it is my own user error or an opentelemetry-swift implementation issue.

I am currently using opentelemetry-swift along with ADOT to put metrics generated by using our SDK into AWS Cloudwatch. However, I noticed that when I switched the metric calculation from average to a percentile like p90 the value goes to 0. After inspecting the logs I discovered that my metrics are being sent as follows:

```
"smithy.client.duration": {
    "Max": 0,
    "Min": 0,
    "Count": 12,
    "Sum": 1.0121349096298218
},
```

Likely, cloudwatch is using min/max in its calculation of the p90 value which is why the data always goes to 0. Using a debugger I saw that min and max were there for each count of the datapoint (equal to whatever was in sum) but I would expect that across several counts, max != min != sum and max/min never to be 0.

Is there a way to fix this so that max and min are properly propagated to the exported data?

nachoBonafonte:
Fixes #594


vvydier (approved):
looks like it was only missing in Histogram case, already in ExponentialHistogram. looks good.

PR #604, opened 2024-09-20
Fix `SpanExceptionTests` on Linux

brunnoferreira:


nachoBonafonte (approved):

brunnoferreira:
Oh wow that's a quick approval @nachoBonafonte!

Thanks for that and apologies for bothering with the flurry of changes recently! Really just wanted to correct #599 being merged with failing tests on Linux so it doesn't get in the way of future contributors.

nachoBonafonte:
No worries, when I merged it I thought it was a flaky test (we have some) after build on main I realized it was not, so fixing it asap was important. Thanks for doing it.

PR #610, opened 2024-09-30
added release workflow

bryce-b:
This workflow;
- accepts a version parameter
- commit new version to Version.swift
- creates tag as `v${new_version}`
- generates a pre-release with name `{new_version}` with auto-generated notes.


nachoBonafonte (approved):
Worked really well in muy tests, great improvement

PR #619, opened 2024-10-25
deprecated root interfaces for old metrics

bryce-b:
I chose to only deprecate the root interfaces for metrics, rather than all the various meters

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocolGrpc/metric/OtlpMetricExporter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolGrpc/metric/OtlpMetricExporter.swift
@@ -12,6 +12,7 @@ import OpenTelemetryApi
 import OpenTelemetryProtocolExporterCommon
 import OpenTelemetrySdk
 
+@available(*, deprecated, renamed: "StableOtlpMetricExporter")
 public class OtlpMetricExporter: MetricExporter {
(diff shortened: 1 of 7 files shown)


nachoBonafonte:
How about directly renaming StableMetric -> Metric and existing Metric to OldMetrics, I know it can sound painful at first but it may be quickly solved by the users of the old metrics by using something like typealias Metric = OldMetrics.

bryce-b:
I thought about this, but I'm worried that marking the old metrics deprecated and also changing their names will result in some confusion, and will result in users having their APIs change in some classes without them realizing. For example, the StableMetrics `StableMeter` will be renamed to `Meter` and the "old metrics" `Meter` will be marked as deprecated and renamed `OldMeter`. This will result in no deprecation warning for anyone using `Meter`, but it will also require immediate intervention from users when they receive the change.  I think it would be better if we warn people the change is coming with the deprecation notice initially, but not force immediate interaction with code-fixing and give users some time to switch over to the stable version, and do a name change later (which will be less confusing and easier to do with developer tools).  

@nachoBonafonte, What do you think?

nachoBonafonte:
You are right, a warning period is necessary here

nachoBonafonte (approved):

vvydier (commented):
good plan.

PR #631, opened 2024-11-06
Added data compression support for HTTP exporter

mamunto:
HTTP client didn't have any compression support. In this PR I made following changes 
* Extracted `DataCompression` from DataDogExporter and made new target 
* Introduced `gzip` and `deflate`  compression support


mamunto:
@nachoBonafonte any thought on Linux platform support? We will need a completely different cross-platform library. Let's discuss to find a solution.

Sources/Exporters/DataCompression/DataCompression.swift
```diff
@@ -28,9 +28,11 @@
 
 
 import Foundation
+
+#if canImport(Compression)
 import Compression
```
mamunto, line 33: Looks like macOS and Linus both fail with that solution. @nachoBonafonte @bryce-b let me know if there is anywhere else I need to address the platform issue. Most probably in the `Package.swift` file but not entirely sure where

Sources/Exporters/OpenTelemetryProtocolHttp/OtlpHttpExporterBase.swift
```diff
@@ -28,20 +31,44 @@ public class OtlpHttpExporterBase {
...
+            case .deflate:
+                if let data = rawData.deflate() {
+                    compressedData = data
+                    request.setValue("deflate", forHTTPHeaderField: "Content-Encoding")
+                }
+                
+            case .none:
+                break
+            }
+#endif
```
mamunto, line 62: Did same

Sources/Exporters/DataCompression/DataCompression.swift
bryce-b: I think the issue is you're importing DataCompression in the exporters when there's nothing there due to the #if

mamunto: Looks Test was a blocker. Fixed it. Thanks for the tip!

mamunto:
It would be nice if we can include this PR and [here is another](https://github.com/open-telemetry/opentelemetry-swift/pull/634) in next release. Really appreciate your time and help

nachoBonafonte (approved):

vvydier (approved):
looks good to me. All the tests passed as well

PR #632, opened 2024-11-06
Modify `toSpanData` to happen inside `attributesLock` to prevent data races

### LINKED ISSUE #630: Concurrency issues with `SpanData` modification leading to vrashes
Greetings! Our SDK uses `OpenTelemetryApi` and `OpenTelemetrySdk`, and we've encountered an issue related to crashes caused by concurrency problems, specifically when working with `SpanData`. I've only found [this documentation](https://github.com/open-telemetry/opentelemetry-specification/blob/8d043e9763aa3818b1ebb904a6c6efb5746e531b/specification/trace/api.md#concurrency) around concurrency, so i'm not pretty sure if this is or not the expected behavior. When testing directly with the `Span`, this issue does not occur (probably because adding events or attributes occur inside a `Lock`). I've added some tests to verify both cases:
```swift
class OpenTelemetryConcurrencyTests: XCTestCase {
    /// Crashes :(
    func test_spanDataShouldntCrashOnBeingModifiedFromDifferentThreads() {
        var data = SpanData(
            traceId: .random(),
            spanId: .random(),
            name: #function,
            kind: .internal,
            startTime: Date(),
            endTime: Date().addingTimeInterval(100)
        )
        
        DispatchQueue.concurrentPerform(iterations: Int.random(in: 100...150), execute: { run in
            data.settingAttributes([UUID().uuidString: .int(run)])
        })
    }
    
    /// Works!
    func test_spanShoudlntCrashOnBeingModifiedFromDifferentThreads() {
        let tracer = TracerProviderSdk()
        let builder = SpanBuilderSdk(spanName: #function, instrumentationScopeInfo: .init(), tracerSharedState: tracer.sharedState, spanLimits: tracer.getActiveSpanLimits())
        let span = builder.startSpan()

        DispatchQueue.concurrentPerform(iterations: Int.random(in: 100...150), execute: { run in
            span.setAttribute(key: UUID().uuidString, value: .int(run))
        })
    }
}
```

ArielDemarco:
# Overview
This PR addresses a potential data race issue when `toSpanData()` is called concurrently with modifications to `attributes` by wrapping both operations within the same lock. 

## Explanation
When somebody needs to call `toSpanData` during the lifetime of a `Span`, the `toSpanData()` method accesses `attributes` and `totalAttributeCount ` while another thread may be modifying them (each time an `attribute` is added/removed `totalAttributeCount` is also modified), leading to potential crashes or memory corruption.

Fixes #630

### DIFF
--- a/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
+++ b/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
@@ -192,29 +192,27 @@ public class RecordEventsReadableSpan: ReadableSpan {
     }
 
     public func toSpanData() -> SpanData {
-        eventsSyncLock.withLock {
-            attributesSyncLock.withLock {
-                return SpanData(traceId: context.traceId,
-                                spanId: context.spanId,
-                                traceFlags: context.traceFlags,
-                                traceState: context.traceState,
-                                parentSpanId: parentContext?.spanId,
-                                resource: resource,
-                                instrumentationScope: instrumentationScopeInfo,
-                                name: name,
-                                kind: kind,
-                                startTime: startTime,
-                                attributes: attributes.attributes,
-                                events: adaptEvents(),
-                                links: adaptLinks(),
-                                status: status,
-                                endTime: endTime ?? clock.now,
-                                hasRemoteParent: hasRemoteParent,
-                                hasEnded: hasEnded,
-                                totalRecordedEvents: getTotalRecordedEvents(),
-                                totalRecordedLinks: totalRecordedLinks,
-                                totalAttributeCount: totalAttributeCount)
-            }
+        attributesSyncLock.withLock {
+            return SpanData(traceId: context.traceId,
+                            spanId: context.spanId,
+                            traceFlags: context.traceFlags,
+                            traceState: context.traceState,
+                            parentSpanId: parentContext?.spanId,
+                            resource: resource,
+                            instrumentationScope: instrumentationScopeInfo,
+                            name: name,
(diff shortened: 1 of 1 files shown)


nachoBonafonte:
Sorry I just realized that it must be locked also with the other lock (the event one) as also events could be modified in while calling it ( this pattern is also used in the stop method)

ArielDemarco:
> Sorry I just realized that it must be locked also with the other lock (the event one) as also events could be modified in while calling it ( this pattern is also used in the stop method)

You're right! that lock also wraps properties that are going to be used by the `toSpanData` method so it should be wrapped by that lock too. Already made the commit

nachoBonafonte (approved):

nachoBonafonte (approved):

ArielDemarco:
Seems like adding the `eventsSyncLock` causes a deadlock (that's why tests are failing), cause inside of the `withLock` in `toSpanData` there's a call to `adaptEvents()` which also uses the `eventsSyncLock`. Seems that every other thing locked with the `eventsSyncLock` is also being called directly from the `toSpanData` (`adaptEvents`, `getTotalRecordedEvents`).
So maybe is safe to say that the best thing to do is to remove the `eventsSyncLock` usage in this method? what do you think @nachoBonafonte ?

nachoBonafonte:
Yes, if all the calls being made by the method are covered then it is not needed. It was my fault asking you to add it, I should have checked first, sorry.

ArielDemarco:
No problem! did the rollback! Hope next run works fine

nachoBonafonte:
It worked with last changes so I merged. Thanks a lot.

PR #639, opened 2024-11-20
Add support for Default Baggage Injection in URLSessionInstrumentation

batuhansk:
**Description:**
This PR introduces a `defaultBaggageProvider` callback in `URLSessionInstrumentationConfiguration`, enabling a default baggage to be applied globally to all instrumented requests. The provided baggage is combined with active baggage when available, or used standalone otherwise. The `tracePropagationHTTPHeaders` method has been optimized to ensure clear and concise baggage propagation across all requests.


Tests/InstrumentationTests/URLSessionTests/URLSessionInstrumentationTests.swift
```diff
@@ -250,7 +260,7 @@ class URLSessionInstrumentationTests: XCTestCase {
         XCTAssertTrue(URLSessionInstrumentationTests.checker.receivedErrorCalled)
     }
 
-    public func testShouldInstrumentRequest() throws {
+    public func testShouldInstrumentRequest_PropagateCombinedActiveAndDefaultBaggages() throws {
```
nachoBonafonte, line 263: Could this functionally be tested in different tests? Just move baggage evaluation to two new different tests, and keep the existing test without baggage evaluation

nachoBonafonte (changes requested):
LGTM, just a small change would be desired

Tests/InstrumentationTests/URLSessionTests/URLSessionInstrumentationTests.swift
batuhansk: Makes sense! Baggage logic is now in separate tests, and the original test stays focused on the basics.

nachoBonafonte (approved):
Thanks for updating, looks perfect for me now.

vvydier (approved):

PR #654, opened 2024-12-12
BatchedSpanProcessor does not wait for all exports to finish on forceFlush

arriIsHere:
Fixes problem where `forceFlush` doesn't wait for all exports to be fully completed before returning by adding a `flush` call on span exporters after an export call.


nachoBonafonte:
Sorry but I don't understand what you want to achieve here, because you are currently flushing for every succesful export , aren't you?

arriIsHere:
I took a look and this I might be barking up the wrong tree with this. The `OtlpHttpTraceExporter` is likely where the problem lies. I'll close this one so I can look over how to solve that.

PR #655, opened 2024-12-13
added pod trunk push to Tag-And-Release.yml

bryce-b:
This will push our podspecs to trunk when the release job runs.

We won't know if this works or not until the next release. 😬

### DIFF
--- a/.github/workflows/Tag-And-Release.yml
+++ b/.github/workflows/Tag-And-Release.yml
@@ -32,3 +32,11 @@ jobs:
           tag: ${{ steps.check-release.outputs.version }}
           prerelease: true
           generateReleaseNotes: true
+      - name: push cocoapods
+        env: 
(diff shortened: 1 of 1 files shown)


nachoBonafonte (approved):

Sherlouk:
How long do we expect this to last? 

CocoaPods have actively said they're making it read-only later this year, and actively discourage new integrations (though I appreciate this app is already on trunk).

Is it not worth encouraging folks to transition, or add this repo as a source (not using CocoaPods trunk even if we provide a spec)

bryce-b:
I believe the main goal behind this is to provide people using react-native some kind of support. I can't see supporting cocoapods beyond react-native figuring out how manage dependencies using SPM. Adding cocoapod support for the SDK and API leaves a lot to be desired anyway: you cannot use any of the instrumentation packages or even the exporters. ( OTLP over GRPC, specifically the swift-nio dependency, is the main hang up for supporting cocoapods for the exporters. ) Now that the trunk push is automated, there isn't a lot of maintenance cost, at least.

PR #657, opened 2024-12-14
enabled swiftlint & fixed issues

bryce-b:
I've added the swiftlint plugin to the 5.9 package.swift. I've also added .swiftlint and disabled several rules that are less simple to resolve, that we may revisit in the future.


bryce-b:
It looks like I'll need to review swiftlint & the github actions. I'll investigate this in the new year.

ArielDemarco:
My 2 cents on the swiftlint usage: If the primary consumer of `OpenTelemetry-Swift` also uses SwiftLint, you’re likely to encounter many integration issues if they're using SPM.
Beyond the fact that when you download a dependency with SPM, it pulls in all its dependencies (meaning `SwiftLint` would also be downloaded in this case), one of the main issues with `SwiftLint` is that, aside from being heavy as a dependency, it’s also **_not stable and introduces many breaking changes from version to version_** (you can check on [their releases](https://github.com/realm/SwiftLint/releases) they even include a `Breaking` changes section)

As a result, I believe this will cause significant integration conflicts for OpenTelemetry with many apps due to SwiftLint's instability and how unfriendly SPM is when handling heavy dependencies (Sadly, I speak from experience).

What I'd do is find other ways to enforce Swiftlint rules without exposing Swiftlint as a plugin in the `Package.swift`, either with a pre-commit hook (which could also `--fix`/`--autocorrect` issues) or other CI tooling.

bryce-b:
@ArielDemarco I really like being able to use swiftlint in xcode, and I think I've come up with a solution!
Using an env-var when loading xcode to enable it, it will not show up in the dependencies for consumers of our package, and it will not conflict with consumers that also use swiftlint.

ArielDemarco:
Looks great! I'll give it a try on a sample project but seems it'll work just fine!

Package.swift
```diff
@@ -16,7 +16,7 @@ let package = Package(
         .library(name: "ResourceExtension", type: .static, targets: ["ResourceExtension"]),
         .library(name: "URLSessionInstrumentation", type: .static, targets: ["URLSessionInstrumentation"]),
         .library(name: "SignPostIntegration", type: .static, targets: ["SignPostIntegration"]),
-        .library(name: "OpenTracingShim-experimental", type: .static, targets: ["OpenTracingShim"]),
+        .library(name: " -experimental", type: .static, targets: ["OpenTracingShim"]),
```
nachoBonafonte, line 19: This looks like an unwanted change

Package@swift-5.9.swift
```diff
@@ -33,6 +34,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.20.2"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.4.4"),
         .package(url: "https://github.com/apple/swift-metrics.git", from: "2.1.1"),
+//        .package(url: "https://github.com/SimplyDanny/SwiftLintPlugins", from: "0.57.1")
```
nachoBonafonte, line 37: This also should be removed

nachoBonafonte (changes requested):
Just a pair of details in the package files

nachoBonafonte (approved):

Issue #615, opened 2024-10-14
(Linux) Build failure against 2024-09-25 nightly toolchain -- error: cannot inherit from class 'Thread' (compiled with Swift 6.1) because it has overridable members that could not be loaded in Swift 5.10


clackary:
This one is another instance of a swiftlang compiler bug I've been running into for a while: https://github.com/swiftlang/swift/issues/76752

```
root@7481e59c0960:/project/opentelemetry-swift# /home/dev-toolchain/swift-DEVELOPMENT-SNAPSHOT-2024-09-25-a-ubuntu22.04-aarch64/usr/bin/swift build
...
/project/opentelemetry-swift/Sources/OpenTelemetrySdk/Logs/Processors/BatchLogRecordProcessor.swift:41:15: error: cannot inherit from class 'Thread' (compiled with Swift 6.1) because it has overridable members that could not be loaded in Swift 5.10
 39 | }
 40 | 
 41 | private class BatchWorker : Thread {
    |               `- error: cannot inherit from class 'Thread' (compiled with Swift 6.1) because it has overridable members that could not be loaded in Swift 5.10
 42 |   let logRecordExporter : LogRecordExporter
 43 |   let scheduleDelay : TimeInterval
```

From what I've seen in the other cases, it looks like would be feasible to workaround the issue by making a Thread property within BatchWorker, and then wrapping the `start` and `cancel` calls with matching BatchWorker funcs.

But the problem here is the override on `Thread.main()`. I don't think there's a way to do this without inheriting from Thread.

Configuration:
OS: Linux Ubuntu 22.04
Toolchain: swift-DEVELOPMENT-SNAPSHOT-2024-09-25-a-ubuntu22.04-aarch64

dayaffe:
+1 to fixing this issue as we are looking to use otel-swift on linux in aws-sdk-swift. Found this issue after running into the same error mentioned

PR #667, opened 2025-01-17
Work around open-telemetry/opentelemetry-swift#615

semicoleon:
open-telemetry/opentelemetry-swift#615 is still an issue when building for Linux, and the PR in the swift project appears to have stalled. This PR makes the subclass of `Thread` a plain class with a `Thread` property that spawns a thread with a closure instead. This appears to work fine on macOS and Linux in my limited testing so far.

I'm not sure if this PR should actually be merged, ideally the error would be resolved in Swift.

I also made a couple minor changes to support building with the [Static Linux SDK](https://www.swift.org/documentation/articles/static-linux-getting-started.html) as that made checking for Linux only compiler errors considerably easier.


nachoBonafonte (commented):
I am not sure I want the block change also on non Linux platforms, I don't feel very safe changing existing behavior in users, thinking about memory and circular dependencies here. Could it be changed to only affect Linux?

semicoleon:
@nachoBonafonte Understood! I went ahead made that change. I also added some tests to try to catch any reference cycle issues.

nachoBonafonte (approved):

bryce-b:
> /Users/runner/work/opentelemetry-swift/opentelemetry-swift/Tests/OpenTelemetrySdkTests/Logs/BatchLogRecordProcessorTests.swift:124: error: -[OpenTelemetrySdkTests.BatchLogRecordProcessorTests testShutdownNoMemoryCycle] : XCTAssertNil failed: "OpenTelemetrySdkTests.WaitingLogRecordExporter"
Test Case '-[OpenTelemetrySdkTests.BatchLogRecordProcessorTests testShutdownNoMemoryCycle]' failed (2.625 seconds).

Please take a look at this test that is failing.

PR #673, opened 2025-02-01
Add baggage span processor

martin308:
Adds a baggage span processor as per the other contrib repositories.

This processor takes items from the baggage and adds the items as attributes on spans.

Eg. https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5404


martin308:
The macOS tests look like they might be flaky but the linux test result is for my new tests will poke around and try to find the cause

nachoBonafonte (changes requested):
Could you change it to a more expressible naming, it looks very generic like this
It also would need proper documentation in the source code for the future  users, for the rest it LGTM

martin308:
From the sync, naming wise we will go `BaggagePropagationProcessor`

martin308:
Updated with the following changes:
- new name for the processor
- moved to an individual library/target for the processor

I left the contrib tests as a test target but we could break those up into individual test targets too if we think it is worth it?

nachoBonafonte (approved):
Looks good to e, I don't mind having the contribution tests under the same test target or separated

bryce-b (approved):

vvydier (approved):

PR #678, opened 2025-02-13
Add cocoapods for Exporters and Instrumentation libraries

arriIsHere:
Currently there are only two published cocoapods (the SDK and the API) for developers using cocoapods for their dependency management they would not be able to use common instrumentation and span exporters.

This PR ads several such as the HTTP exporter which is commonly used.


ArielDemarco:
Tested this by pushing all specs to a private repo and adding them to the Podfile as follows:
```ruby
...
    pod 'OpenTelemetry-Swift-Api'
    pod 'OpenTelemetry-Swift-Sdk'
    pod 'OpenTelemetry-Swift-StdoutExporter'
    pod 'OpenTelemetry-Swift-SdkResourceExtension'
    pod 'OpenTelemetry-Swift-Protocol-Exporter-Common'
    pod 'OpenTelemetry-Swift-Protocol-Exporter-Http'
    pod 'OpenTelemetry-Swift-DataCompression'
...
```
I was able to import them in code and run without any issues on `tvOS`, `watchOS`, and `iOS`.

_Note_: I haven’t actually tested the functionality of each module.

vvydier (approved):

bryce-b (approved):

PR #681, opened 2025-02-19
Reformatted Files for consistency

bryce-b:
This PR utilizes [swiftformat](https://github.com/nicklockwood/SwiftFormat) to apply formatting rules to the entire project.
`.swiftformat` configuration files is added with a few additional rules so it will play nice with swiftlint.

I chose `swiftformat` over formatting with `swiftlint` due to some issues I encountered with the indentation rules with `swiftlint`, such as errors on multi-line function definitions, and block-comments. More details on similar issues I experienced are enumerated in this issue : https://github.com/realm/SwiftLint/issues/3046.

I chose to format using 2-space indentation over 4, because it represented less changes to the project over all.

There are some additional formatting rules that are being applied, such as [redundant returns](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#redundantreturn) being removed. I don't mind this, but I want to bring this to reviewers' attention in case there is contention. [opaque generic parameters](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#opaquegenericparameters) is another I notice. Please take a close look and let me know if there are any rules being applied that you are not happy with. 

Once this PR is merged I'll add a swiftformat as either a lint requirement, or as an auto-reformatter to PRs. 

Additionally, I disabled the 'todo' rule and excluded the "Tests" folder for swiftlint.


ArielDemarco:
I really like the idea of using SwiftFormat, and I think most of its rules are ok. I thought I wouldn’t like the 2-space indentation formatting, but after reading through the PR, I got used to it and actually find it quite neat. Changes related to _redundant return_ and _opaque generic parameters_  also look good to me.

After reading the PR, the only rule that I'd exclude is the [redundantExternalParameterName](http://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#unusedarguments), as it feels like it makes the code a bit messier (at least, from the user/readers perspective).
For example, this is a change I saw (extracted from `InMemoryExporter`):
```swift
// Before swiftformat
public func flush(explicitTimeout: TimeInterval? = nil) -> SpanExporterResultCode

// after swiftformat
public func flush(explicitTimeout _: TimeInterval? = nil) -> SpanExporterResultCode
```

From the API perspective, nothing changes as a client could use it in the same way as before:
```swift
someExporter.flush(explicitTimeout: 2.0)
```

But within the codebase, this `_` makes the function signature look odd and less clear. For someone reading it, it might raise questions: Is this parameter optional and unused? Should/Will it be removed in the future? Was this intentional? IMO, keeping the original parameter name improves clarity without any downsides.

That said, I wouldn’t mind applying this rule to _closures_ instead of functions, using the `closure-only` option. In closures, replacing unused parameters with `_` makes sense and can help readability, but for function signatures, I feel it does more harm than good (_Note_: this is **_totally subjective_**, and maybe it’s just because I’m not used to it)

nachoBonafonte:
I like this PR but It is really difficult to review, would it be possible to do it in two steps?
1- All changes except any space indentation
2- Add all the space indentation
All the indentation changes are really making the diff comparison totally unusable, and it would allow focusing on the other changes. And add all the indentation changes after the first approval.

Sources/Exporters/OpenTelemetryProtocolCommon/common/EnvVarHeaders.swift
```diff
@@ -8,56 +8,56 @@ import OpenTelemetryApi
...
-    /// NOTE: Parsing does not fully match W3C Correlation-Context
-    private static func parseAttributes(rawEnvAttributes: String?) -> [(String, String)]? {
-        guard let rawEnvLabels = rawEnvAttributes else { return nil }
-
-        var labels = [(String, String)]()
-
-        rawEnvLabels.split(separator: labelListSplitter).forEach {
-            let split = $0.split(separator: labelKeyValueSplitter)
-            if split.count != 2 {
-                return
```
bryce-b, line 50: 🤔

Sources/OpenTelemetryApi/Baggage/DefaultBaggageBuilder.swift
```diff
@@ -35,13 +35,14 @@ public class DefaultBaggageBuilder: BaggageBuilder {
...
-      let entryValue = EntryValue(string: value)
+          let entryValue = EntryValue(string: value)
     else {
       return self
     }
     let entry = Entry(
       key: entryKey, value: entryValue,
-      metadata: EntryMetadata(metadata: metadata))
+      metadata: EntryMetadata(metadata: metadata)
+    )
```
bryce-b, line 45: what do we think about this formatting?

Sources/OpenTelemetryApi/Baggage/Propagation/JaegerBaggagePropagator.swift
```diff
@@ -11,46 +11,46 @@ import Foundation
...
+  public var fields: Set<String> = [baggageHeader]
 
-    public init() {}
+  public init() {}
 
-    public func inject<S>(baggage: Baggage, carrier: inout [String: String], setter: S) where S: Setter {
-        baggage.getEntries().forEach {
-            setter.set(carrier: &carrier, key: JaegerBaggagePropagator.baggagePrefix + $0.key.name, value: $0.value.string)
-        }
+  public func inject(baggage: Baggage, carrier: inout [String: String], setter: some Setter) {
```
bryce-b, line 21: I prefer this over the templating. What do you think?

bryce-b:
I've updated this pr to (mostly) only apply whitespace changes before applying additional formatting rules

nachoBonafonte:
Even hiding the whitespaces in the GitHub review it was barely scrollable, but after reviewing most of it, LGTM

PR #691, opened 2025-02-27
swiftformat applied

bryce-b:
I applied rules one or two at a time, so this pr can be reviewed per commit, to minimized the total number of files being look at while reviewing.


ArielDemarco (approved):
LGTM! It was much easier to read with each rule applied in each commit.

nachoBonafonte (approved):
Thanks for the separated commits, it made it  reviewable . I like them all, just have some doubts with Preferkeypath, but have no problem with it either

bryce-b:
Yes, I'm a little apprehensive about Preferkeypath as well. I'm not sure the benefit of using over the standard default accessors. I'm doing a little research as to why it may be preferred.

bryce-b:
I'll leave the preferKeyPath rule in for now. It seems like this particular change reduces the number of function calls by 1. We can remove this rule if we decide we doing like it.

PR #696, opened 2025-03-06
expose setUnit on meter builders

angelodipaolo:
Provides the ability to set the unit on counter and histogram builders.

### DIFF
--- a/Sources/OpenTelemetryApi/Metrics/Stable/DefaultStableMeter.swift
+++ b/Sources/OpenTelemetryApi/Metrics/Stable/DefaultStableMeter.swift
@@ -43,6 +43,10 @@ public class DefaultStableMeter: StableMeter {
       NoopLongHistogramBuilder()
     }
 
+    func setUnit(_ unit: String) -> Self {
+      self
+    }
+
     func build() -> DoubleHistogram {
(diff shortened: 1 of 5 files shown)


bryce-b:
Hi @angelodipaolo, this setter is already available on all instrument builders, but they are obscured by return type of StableMeter's instrument builders. This is a larger problem with the API that we need to solve. You can work around this by coercing the type as such: 
```
stableMeter.counterBuilder(name: "a") as LongCounterMeterBuilderSdk
.setUnit(...)
.build() 
```

bryce-b:
I've created a ticket for this issue #697

angelodipaolo:
@bryce-b Thanks for the info!

After casting to `LongCounterMeterBuilderSdk`, I still cannot access `setUnit(_)`. I see the `setUnit(_)` extension on `InstrumentBuilder` is public, but it seems because `InstrumentBuilder` is internal, I cannot access `setUnit(_)` externally.


image

bryce-b:
@angelodipaolo ah, another bug! InstrumentBuilder shouldn't be internal, or at least the extension should be public. I'll create an additional issue and make sure it's fixed in the next release.

PR #722, opened 2025-03-07
refactored InstrumentBuilder to make relevant builder function public

### LINKED ISSUE #721: InstrumentBuilder incorrectly marked as internal
https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/OpenTelemetrySdk/Metrics/Stable/InstrumentBuilder.swift

This is a little more complicated than just making this public, as some member vars are internal only.

bryce-b:
Fixes #721 
a 'virtual' protocol is not viable in this situation because the member variables and extension functions need variable visibility, both internal and public


bryce-b:
@angelodipaolo would you mind giving this PR a try?

angelodipaolo:
@bryce-b Tested and works great. Thanks for fixing this so quickly!

nachoBonafonte (approved):

PR #725, opened 2025-03-13
Remove Datadog exporter code.

nachoBonafonte:
There is no sense in keeping this code for the following reasons:
- Datadog currently supports importing OTLP natively
- Datadog also distributes a Otel compatible SDK for iOS
- It cannot be tested anymore in production by any of the maintainers
- It existed as a complex and functional example of what could be done with the SDK, but there are now several open-source implementations written on top of opentelemetry-swift


vvydier (approved):
make sense to not have a vendor specific exporter in the otel-swift repo that is not being maintained for spec changes. May we could refactor it as a generic exporter if someone wants to use it as an example to write their own

bryce-b (approved):
LGTM

Issue #735, opened 2025-03-25
Conflicting dependency: `DataCompression`


maxkomarychev:
It appears this package creates a target called `DataCompression` which conflicts with other libraries depending on this 3rdparty dependency https://github.com/mw99/DataCompression. In particular my trouble is with LaunchDarkly Swift SDK.

When installing via Xcode => Packages it shows an error:

<img width="1126" alt="Image" src="https://github.com/user-attachments/assets/3fee5cee-67c7-4b32-96ca-5576608fd70a" />

When installing via cocoapods it shows an error during build: 

<img width="1055" alt="Image" src="https://github.com/user-attachments/assets/fce7bddc-0772-4f56-afac-a7eb21cb0664" />

Could you please namespace the package to avoid conflicts with other existing packages?

Thanks!

maxkomarychev:
Please confider the fix provided here https://github.com/open-telemetry/opentelemetry-swift/pull/736!

PR #736, opened 2025-03-25
Namespace DataCompression to avoid conflicts

maxkomarychev:
If another library is using this dependency https://github.com/mw99/DataCompression it can not coexist in the same project with `optentelemetry-swift` because of the conflict in names of targets: `DataCompression`

Fixes: https://github.com/open-telemetry/opentelemetry-swift/issues/735


maxkomarychev:
Could you clarify please how does this project deal with breaking changes?

I could publish the package twice to keep the old code running but the new versions should be using new namespaces one.

Would that be ok?


p.s. 

> This is a breaking change for existing users of the opentelemetry data compression library, it would need previous compiler warning for at least one version.

To be honest this change from OTEL 1.11 to 1.12 has broken my code. So the original change can also be treated as a breaking change 🙃

maxkomarychev:
> Then we should revert the change we introduced to break it. Is this like the change you proposed?

The problematic change has been introduced in [here](https://github.com/open-telemetry/opentelemetry-swift/commit/60f92600b052cc39c82415fc05f023c0b3370b98#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR32-R115) by somebody else.

It is not a breaking change in "vacuum" in terms of existing API of OTEL-Swift but it broke the app I'm working on because I happened to have *another* popular library ([launchdarkly swift](https://github.com/launchdarkly/ios-client-sdk)) [depending on](https://github.com/launchdarkly/ios-client-sdk/blob/93963c27142f04bdebc962bbdabbc7ea0ffd63b7/Package.swift#L24) https://github.com/mw99/DataCompression

maxkomarychev:
Hi 👋 What's the path forward with this change? Although it does look like a breaking one I would speculate very few if any projects are using `DataCompression` library from this repository directly. That said I totally understand your reasoning and I'm happy to make adjustments in the PR to make it merged. What's the best way to add a warning?

Thanks!

nachoBonafonte:
Hi @maxkomarychev, we are merging your changes in a new branch just created from main: `datacompression-rename` . It will include the same code than the 1.15.0 version of the library being created today/tomorrow. As said it needs a major version to land in the official code, but we have created it in a branch so you can continue using it on your side.

nachoBonafonte:
We can update the branch with each new version, we expect to release a major version relatively soon, I cannot provide a date as depends on some metrics validation but not very far away.

Issue #744, opened 2025-04-07
URL Instrumentation prevents Session Delegates from being called in async contexts


MustafaHaddara:
### The Problem

For example, let's say we have this session delegate:

```swift
class SessionDelegate: NSObject, URLSessionDataDelegate, URLSessionTaskDelegate {
  var callCount = 0

  func urlSession(
    _ session: URLSession,
    task: URLSessionTask,
    didFinishCollecting metrics: URLSessionTaskMetrics
  ) {
    callCount += 1
    print("delegate called")
  }
}
```

and then we use this as a delegate for our session and make a synchronous request, everything is fine. For example:

```swift
let delegate = SessionDelegate()
let session = URLSession(configuration: .default, delegate: delegate, delegateQueue: nil)

let url = URL(string: "http://httpbin.org/get")!
let request = URLRequest(url: url)

let task = session.dataTask(with: request)
task.resume()
```

After this runs, `delegate.callCount` will be `1` and `delegate called` will be printed in the console. So far so good.

If we switch to making our request async and use this delegate as a delegate for the _task_, things are still fine. For example:

```swift
let session = URLSession(configuration: .default)
let url = URL(string: "http://httpbin.org/get")!
let request = URLRequest(url: url)
do {
    _ = try await session.data(for: request, delegate: delegate)
} catch {
    return
}
```

Then, again, this works the same as above.

However, if we make the request async and use this delegate as the delegate for the entire session, then things fall apart. That code looks like this:

```swift
let session = URLSession(configuration: .default, delegate: delegate, delegateQueue: nil)
let url = URL(string: "http://httpbin.org/get")!
let request = URLRequest(url: url)
do {
    _ = try await session.data(for: request)
} catch {
    return
}
```

And when this runs, `delegate.callCount` remains `0` and there is no `delegate called` message in the console. 

### Standalone Example
I've reproduced this case in the `URLSessionInstrumentationTests` and in the standalone `Network Sample`. My full code change is [here](https://github.com/open-telemetry/opentelemetry-swift/compare/main...MustafaHaddara:opentelemetry-swift:experimental-reproduce-urlsession-bug).

### Potential Solution
[...]

PR #747, opened 2025-04-10
URLInstrumentation: only set a Task Delegate if there is no Session Delegate

MustafaHaddara:
Fixes https://github.com/open-telemetry/opentelemetry-swift/issues/744

Majority of the changes are in test/example code. The actual fix is quite small, but relies on calling `task.value(forKey: "session") as? URLSession` to get the associated session on a task. 

### Intercept Task Creation
One other approach we discussed in the sig was to intercept the task creation and call `objc_setAssociatedObject` on the task, if the session had an active delegate. We already do this here: https://github.com/open-telemetry/opentelemetry-swift/blob/6a2c29d53ff0b543b551b2221538bd3d0206c6d6/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift#L192-L195

However, that blog of code isn't swizzling the `URLSession.data(for:)` or `URLSession.data(from:)` functions, which are used to create asynchronous tasks. We could start swizzling those functions as well, if we wanted to, but that felt like a riskier change. 

### Swizzle UISession
Honeycomb has [an alternate URL instrumentation implementation](https://github.com/honeycombio/honeycomb-opentelemetry-swift/blob/main/Sources/Honeycomb/Networking/URLSession.swift) that swizzles the `URLSession` and `URLSessionTask` _constructors_ directly. This lets us get a much stronger handle on the sessions and tasks in a wider variety of cases. Switching to that approach would be more costly and time consuming, but may serve us in the long run since the implementation is simpler.


ArielDemarco:
The PR looks good to me!

The only thing I’d recommend is adding a few tests to cover the different `NSURLSessionTask` types that can be created (at least the ones this instrumentation handles). I think it’s especially worth testing because the solution relies on KVO of a variable that should exist across the many different `NSURLSessionTask` subclasses, and the test will help ensure nothing breaks across OS versions as time goes by.

That said, the PR is failing due to the examples added. It seems like there's a simple fix needed to wrap some API usages with `@available`

nachoBonafonte (approved):

MustafaHaddara:
@ArielDemarco thanks for taking a look! I've fixed the tests for tvOS and watchOS, and added some tests for async upload and download tasks with a delegate.

ArielDemarco (approved):

PR #748, opened 2025-04-11
update MetricsAdapter to include exemplars

angelodipaolo:
Fixes an issue where exemplars are not being exported through `MetricsAdapter`

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocolCommon/metric/MetricsAdapter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolCommon/metric/MetricsAdapter.swift
@@ -203,6 +203,7 @@ public enum MetricsAdapter {
         protoExemplar.spanID = TraceProtoUtils.toProtoSpanId(spanId: spanContext.spanId)
         protoExemplar.traceID = TraceProtoUtils.toProtoTraceId(traceId: spanContext.traceId)
       }
+      protoPoint.exemplars.append(protoExemplar)
     }
   }
 
@@ -225,6 +226,7 @@ public enum MetricsAdapter {
         protoExemplar.spanID = TraceProtoUtils.toProtoSpanId(spanId: spanContext.spanId)
         protoExemplar.traceID = TraceProtoUtils.toProtoTraceId(traceId: spanContext.traceId)
       }
+      protoPoint.exemplars.append(protoExemplar)
     }
   }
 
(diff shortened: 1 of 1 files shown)


angelodipaolo:
After this change I'm seeing an issue where a lot of exemplars come through with no data.

image


Which results in empty objects in the export request payload:

```
"exemplars": [
  {},
  {
    "timeUnixNano": "1744374297925190912",
    "spanId": "hT/Waa+I6Ww=",
    "traceId": "A/9hQwLHuxwzaHpRsYD4TQ=="
  },
  {},
  {},
  {},
  {},
  {},
  {},
  {},
  {},
  {},
  {},
  {},
  {},
  {},
  {}
],
```

I was tempted to filter these out in `MetricsAdapter` but I'm not sure if that's the right approach since I don't see any other valid/empty checks in the adapter except for arrays. Is there a chance I have something configured incorrectly or is there a bug in how exemplars are aggregated that I should look into?

Sources/Exporters/OpenTelemetryProtocolCommon/proto/metrics.pb.swift
```diff
@@ -997,6 +997,10 @@ public struct Opentelemetry_Proto_Metrics_V1_SummaryDataPoint {
 
   public var unknownFields = SwiftProtobuf.UnknownStorage()
 
+  /// (Optional) List of exemplars collected from
+  /// measurements that were used to form the data point
+  public var exemplars: [Opentelemetry_Proto_Metrics_V1_Exemplar] = []
```
bryce-b, line 1002: we shouldn't edit this file. The summary data point does not have exemplars https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/opentelemetry/proto/metrics/v1/metrics.proto#L616

Sources/Exporters/OpenTelemetryProtocolCommon/metric/MetricsAdapter.swift
```diff
@@ -247,6 +249,7 @@ public enum MetricsAdapter {
         protoExemplar.spanID = TraceProtoUtils.toProtoSpanId(spanId: spanContext.spanId)
         protoExemplar.traceID = TraceProtoUtils.toProtoTraceId(traceId: spanContext.traceId)
       }
+      protoPoint.exemplars.append(protoExemplar)
```
bryce-b, line 252: this plus the above lines can probably be removed since exemplars don't exist for summary data points

Sources/Exporters/OpenTelemetryProtocolCommon/proto/metrics.pb.swift
angelodipaolo: Remove this change.

Sources/Exporters/OpenTelemetryProtocolCommon/metric/MetricsAdapter.swift
angelodipaolo: Thanks for the feedback! I removed this and the `pointData.exemplars.forEach` above assuming that's what you were suggesting.

bryce-b:
This looks good to me. There appears to be a bug somewhere in the creation that generates empty exemplar datapoints. That might be the cause of all the {} in your payload.

angelodipaolo:
> This looks good to me. There appears to be a bug somewhere in the creation that generates empty exemplar datapoints. That might be the cause of all the {} in your payload.

I'll do some more investigation and see if I can track down what's generating empty exemplar datapoints.

bryce-b:
I've created an issue https://github.com/open-telemetry/opentelemetry-swift/issues/749

bryce-b (approved):

PR #751, opened 2025-04-24
Add faro exporter

robert-northmind:
This PR adds a Faro Exporter which can handle logs and traces.
It forwards the telemetry to a Faro collector, e.g. Grafana Cloud, and from there you can get an overview of the telemetry data bundled into user sessions.


Sources/Exporters/Faro/Internal/FaroManagerFactory.swift
```diff
@@ -0,0 +1,36 @@
...
+    // Create and validate endpoint configuration
+    let endpointConfiguration = try FaroEndpointConfiguration.create(from: options)
+
+    let sessionManager = FaroSessionManagerFactory.getInstance()
+    let appInfo = FaroAppInfo.create(from: options)
+    let transport = FaroTransport(endpointConfiguration: endpointConfiguration, sessionManager: sessionManager)
+
+    let faroManager = FaroManager(appInfo: appInfo, transport: transport, sessionManager: sessionManager)
+    managers[options] = faroManager
+    return faroManager
```
ArielDemarco, line 34: Since this method can be invoked from multiple places concurrently, there’s a risk of race conditions when accessing or initializing the shared dictionary (aka `managers`). I’d recommend wrapping this logic in a locking queue here (or even inside `FaroSessionManagerFactory` itself) to ensure thread safety and prevent crashes or unexpected behaviors.

Sources/Exporters/Faro/Internal/DeviceAttributes/FaroPersistentDeviceIdentifierProvider.swift
```diff
@@ -0,0 +1,36 @@
...
+protocol PersistentDeviceIdentifierProviding {
+  func getIdentifier() -> String
+}
+
+protocol UserDefaultsProviding {
+  func string(forKey defaultName: String) -> String?
+  func set(_ value: Any?, forKey defaultName: String)
+}
+
+extension UserDefaults: UserDefaultsProviding {}
```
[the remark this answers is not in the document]
robert-northmind: aha, nice, thanks for the feedback. I didn't think about this.
I will see if I can get the FileManager approach working.

Sources/Exporters/Faro/Internal/FaroManagerFactory.swift
robert-northmind: Good point. I totally missed that. Will add some safety around this part 🙇

Sources/Exporters/Faro/Internal/FaroTransport.swift
```diff
@@ -0,0 +1,107 @@
...
+        return
+      }
+
+      // Check HTTP status code
+      if let httpResponse = response as? HTTPURLResponse {
+        let statusCode = httpResponse.statusCode
+        if statusCode < 200 || statusCode >= 300 {
+          let responseData = data != nil ? String(data: data!, encoding: .utf8) : nil
+          self.logger.logError("FaroTransport: HTTP error: status=\(statusCode), response=\(responseData ?? "nil")", error: nil)
+          completion(.failure(FaroTransportError.httpError(statusCode, responseData)))
```
ArielDemarco, line 89: I'd avoid using force unwrap (even though this shouldn't crash). Something like this should work
```swift
let responseData = if let data = data {
      String(data: data, encoding: .utf8)   
} else {
    "nil"
}
```
If that's the case, `response=\(responseData ?? "nil")",` could be replaced to be `response=\(responseData)"`

Sources/Exporters/Faro/Internal/Utils/FaroLogging.swift
```diff
@@ -0,0 +1,37 @@
...
+  private static var logger: FaroLogging = FaroLogger()
+
+  static func getInstance() -> FaroLogging {
+    return logger
+  }
+
+  static func setLogger(logger: FaroLogging) {
+    self.logger = logger
+  }
+}
```
ArielDemarco, line 37: FWIW, this `enum` is also vulnerable to potential race conditions. While `getInstance()` is safe on its own, concurrent calls to `setLogger()` and `getInstance()` could lead to inconsistent state and/or crashes. Based on its usages, I don't think there's potential risk _today_ so It'd be safe to merge it right now; however, just in case somebody modifies this in the future (or starts using this enum for other places), and it's not aware that this `enum` is not thread-safe, I’d recommend protecting access to the shared logger.

bryce-b:
Hi Robert, I'd like to invite you to come discuss this PR in our weekly meeting on [Thursday at 9am PDT](https://groups.google.com/a/opentelemetry.io/g/calendar-swift).  We discussed this today, and one main concern with this addition is maintenance of it. We'd like to have this contribution to the swift SDK, but we don't necessarily want to take on the maintenance of it. We think the best way of going forward is to have this exporter in a separate repository maintained by. your company, and we can provide a link to it in our readme as an available option. We can discuss this more in our weekly meeting if you'd like. Let me know what you think!

robert-northmind:
> Hi Robert, I'd like to invite you to come discuss this PR in our weekly meeting on [Thursday at 9am PDT](https://groups.google.com/a/opentelemetry.io/g/calendar-swift). We discussed this today, and one main concern with this addition is maintenance of it. We'd like to have this contribution to the swift SDK, but we don't necessarily want to take on the maintenance of it. We think the best way of going forward is to have this exporter in a separate repository maintained by. your company, and we can provide a link to it in our readme as an available option. We can discuss this more in our weekly meeting if you'd like. Let me know what you think!

Hey @bryce-b, thanks for your feedback. I fully understand and it makes sense :) 
I will move this over to a Grafana maintained repo. And it would be great if we could link it in a README from here.
And thanks for the invite for the Swift SIG meeting. I'm out on holiday next week. But I will join the week after.
Thanks again for looking this over and sharing your thoughts 👍

Sources/Exporters/Faro/Internal/FaroTransport.swift
robert-northmind: Nice @ArielDemarco 🤩  I actually didn't know about this syntax. thanks for the tip!

PR #760, opened 2025-05-05
Fix infinite recursion bug when getting tracer

angelodipaolo:
`tracerProvider.get(instrumentationName: "Foo")` results in an infinitely recursive call after recent changes in https://github.com/open-telemetry/opentelemetry-swift/commit/5839cb581053ad36659acc79302615bc2f75de98

image

### DIFF
--- a/Sources/OpenTelemetryApi/Trace/TracerProvider.swift
+++ b/Sources/OpenTelemetryApi/Trace/TracerProvider.swift
@@ -13,24 +13,40 @@ public protocol TracerProvider {
   ///   - instrumentationVersion:  The version of the instrumentation library (e.g., "semver:1.0.0"). Optional
   ///   - schemaUrl: The schema url. Optional
   ///   - attributes: attributes to be associated with span created by this tracer. Optional
-  func get(
-    instrumentationName: String,
-    instrumentationVersion: String?,
-    schemaUrl: String?,
-    attributes: [String: AttributeValue]?
-  ) -> any Tracer
+  func get(instrumentationName: String,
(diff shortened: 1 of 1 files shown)


angelodipaolo:
I'm not sure if the intention of https://github.com/open-telemetry/opentelemetry-swift/commit/5839cb581053ad36659acc79302615bc2f75de98 was to make a breaking change in the public AP but if not, it might be better to define variations of `get` that have the new parameters at the protocol level. That would prevent forcing types that conforms to `TracerProvider` to adopt the breaking API change. 

I'm happy to update this fix if that approach makes sense.

bryce-b (approved):

bryce-b:
> I'm not sure if the intention of https://github.com/open-telemetry/opentelemetry-swift/commit/5839cb581053ad36659acc79302615bc2f75de98 was to make a breaking change in the public AP but if not, it might be better to define variations of get that have the new parameters at the protocol level. That would prevent forcing types that conforms to TracerProvider to adopt the breaking API change.

> I'm happy to update this fix if that approach makes sense.

Please do, this was not meant to be a breaking change

bryce-b (changes requested):
correct breaking change as discussed

nachoBonafonte:
Hi @angelodipaolo, are you planning in update the fix soon? we are planning a release early next week and it would be a great change to include

bryce-b:
Testing this pr out, and there doesn't appear to be any breaking changes. Perhaps @angelodipaolo was referring to my original PR.

nachoBonafonte (approved):

PR #763, opened 2025-05-06
Improve W3C Baggage implementation with proper UTF-8 encoding and RFC7230 validation

### LINKED ISSUE #761: `W3CBaggagePropagator` Missing Percent Encoding for Special Characters in Values
## Issue Description
The current implementation of `W3CBaggagePropagator` in OpenTelemetry Swift does not properly handle percent encoding for special characters in baggage values, which is required by the [W3C Baggage Specification](https://www.w3.org/TR/baggage/#header-encoding).

## Current Behavior
The Swift implementation currently passes baggage values through directly without encoding:
```swift
headerContent += $0.key.name + "=" + $0.value.string
```

This means special characters in values are not properly encoded, which can lead to:
1. Interoperability issues with other OpenTelemetry implementations
2. Potential data corruption when values contain special characters
3. Non-compliance with the W3C Baggage specification

## Expected Behavior
According to the W3C Baggage specification:
1. Values MUST be percent-encoded if they contain special characters
2. The percent sign (`%`) itself MUST be percent-encoded
3. Characters outside the `baggage-octet` range MUST be percent-encoded
   - `baggage-octet` = `%x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E`

Example encodings:
Original: "DF 28" -> Encoded: "DF%2028"
Original: "Amélie" -> Encoded: "Am%C3%A9lie"
Original: "value%" -> Encoded: "value%25"

## Proposed Solution
1. Add percent encoding for values during injection:
```swift
let encodedValue = value.addingPercentEncoding(withAllowedCharacters: allowedCharacters) ?? value
headerContent += key + "=" + encodedValue
```

2. Add percent decoding during extraction:
```swift
let decodedValue = String(keyValue[1]).removingPercentEncoding
```

3. Handle UTF-8 encoding properly:
   - Ensure all strings are properly UTF-8 encoded
   - Replace invalid UTF-8 sequences with the replacement character (U+FFFD) during decoding

## Impact
This issue affects interoperability with other OpenTelemetry implementations that correctly implement the W3C Baggage specification. Any application using the Swift SDK to propagate baggage with special characters may experience data loss or corruption.

## Related Information
- [W3C Baggage Specification - Header Encoding](https://www.w3.org/TR/baggage/#header-encoding)
- Similar implementations:
  - [Java implementation](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/baggage/propagation/W3CBaggagePropagator.java) uses `URLEncoder.encode()`
  - [Go implementation](https://github.com/open-telemetry/opentelemetry-go/blob/main/baggage/baggage.go) implements proper percent encoding

## Additional Context
This implementation should align with Swift's native string handling capabilities while maintaining compliance with the W3C specification.

batuhansk:
## Description
This PR improves the W3C Baggage implementation to better align with the specification requirements, focusing on proper UTF-8 encoding and RFC7230 validation.

### Changes
- Implement proper UTF-8 percent encoding for baggage values
- Add RFC7230 token validation for baggage keys (using allowed characters: `!#$%&'*+-.^_`|~` and alphanumeric)
- Remove ASCII-only restriction from baggage values to allow any UTF-8 characters
- Maintain ASCII validation for resource attributes (separate concern from baggage)

### Key Implementation Details
* `EntryKey`: Now validates against RFC7230 token specification
* `EntryValue`: Allows any UTF-8 characters in values (will be percent-encoded during propagation)
* `W3CBaggagePropagator`: Implements proper percent encoding/decoding
* Tests: Added comprehensive W3C specification compliance tests

### Notes
- Removed previously proposed 180-byte length limit after reviewing other language implementations and considering potential data loss implications
- Maintains backward compatibility while improving spec compliance
- Added extensive test coverage for edge cases and W3C spec examples

### References
- Fixes #761
- [W3C Baggage Specification](https://www.w3.org/TR/baggage/#bib-rfc7230)

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocolCommon/metric/MetricsAdapter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolCommon/metric/MetricsAdapter.swift
@@ -196,6 +196,14 @@ public enum MetricsAdapter {
       var protoExemplar = Opentelemetry_Proto_Metrics_V1_Exemplar()
       protoExemplar.timeUnixNano = $0.epochNanos
 
+      if let doubleExemplar = $0 as? DoubleExemplarData {
+        protoExemplar.value = .asDouble(doubleExemplar.value)
+      }
+
+      if let longExemplar = $0 as? LongExemplarData {
+        protoExemplar.value = .asInt(Int64(longExemplar.value))
+      }
+
       $0.filteredAttributes.forEach {
         protoExemplar.filteredAttributes.append(CommonAdapter.toProtoAttribute(key: $0.key, attributeValue: $0.value))
       }
@@ -218,6 +226,14 @@ public enum MetricsAdapter {
     pointData.exemplars.forEach {
       var protoExemplar = Opentelemetry_Proto_Metrics_V1_Exemplar()
       protoExemplar.timeUnixNano = $0.epochNanos
+
+      if let doubleExemplar = $0 as? DoubleExemplarData {
+        protoExemplar.value = .asDouble(doubleExemplar.value)
+      }
+
+      if let longExemplar = $0 as? LongExemplarData {
+        protoExemplar.value = .asInt(Int64(longExemplar.value))
+      }
 
       $0.filteredAttributes.forEach {
         protoExemplar.filteredAttributes.append(CommonAdapter.toProtoAttribute(key: $0.key, attributeValue: $0.value))


Sources/OpenTelemetryApi/Baggage/Propagation/W3CBaggagePropagator.swift
```diff
@@ -20,54 +20,78 @@ public struct W3CBaggagePropagator: TextMapBaggagePropagator {
...
   static let headerBaggage = "baggage"
 
+  private func isValidKeyValuePair(_ keyValue: String) -> (key: String, value: String)? {
+    let parts = keyValue.split(separator: "=", maxSplits: 1)
+    guard parts.count == 2 else { return nil }
+
+    return (String(parts[0]), String(parts[1]))
+  }
+
+  private func percentEncode(_ value: String) -> String {
```
bryce-b, line 30: is this method used anywhere?

batuhansk: Good catch! It was used during the development, but no longer need this. Just removed it.

bryce-b (approved):

nachoBonafonte (approved):

PR #769, opened 2025-05-09
add podspecs for more targets

beekhc:
This PR adds podspecs for more of the targets that are already in `Package.swift`. I did my best to choose names consistent with the existing podspecs.

I tested these by making a local project using cocoapods, adding these targets as dependencies, and pointing them to the local checkout of the repo.

Note: When pointing at the released versions of the existing pods, I was able to use version 1.13, but 1.15 was not found, so I think there may be some problem with the release workflow. I tried to verify on cocoapods.org, but the repo didn't show up there at all.

### DIFF
--- a/.github/workflows/Create-Release-PR.yml
+++ b/.github/workflows/Create-Release-PR.yml
@@ -30,13 +30,17 @@ jobs:
        ' > Sources/OpenTelemetrySdk/Version.swift
     - name: update Podspec
       run: | 
-        sed -i -e  's/spec.version = ".*"/spec.version = "${{ inputs.new_version }}"/' OpenTelemetry-Swift-Api.podspec    
-        sed -i -e  's/spec.version = ".*"/spec.version = "${{ inputs.new_version }}"/' OpenTelemetry-Swift-Sdk.podspec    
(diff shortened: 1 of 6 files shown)


ArielDemarco:
> oh, it looks like cocoapods aren't getting released because `pod` isn't found in github actions:
> 
> https://github.com/open-telemetry/opentelemetry-swift/actions/runs/14344511947/job/40211452921

Mmm seems the `name: Tag & Note Release` job is running on `ubuntu-latest`. Probably this should be running on a `macos-xyz` (e.g. `macos-15`)

beekhc:
> > oh, it looks like cocoapods aren't getting released because `pod` isn't found in github actions:
> > https://github.com/open-telemetry/opentelemetry-swift/actions/runs/14344511947/job/40211452921
> 
> Mmm seems the `name: Tag & Note Release` job is running on `ubuntu-latest`. Probably this should be running on a `macos-xyz` (e.g. `macos-15`)

That sounds right to me. I could change it in this PR, although it might make sense to do it separately, since I can't test that part. Maybe someone who does releases should change it right before doing the next release, to verify it works? Just let me know which y'all would prefer.

PR #783, opened 2025-05-27
Move tracer to Configuration to support tracer config changes after swizzle

### LINKED ISSUE #756: Re-init URLSessionInstrumentation
Is there a way to denit or destroy or recreate URLSessionInstrumentation after it is initialized? I have a use-case where a user changes the environment or tenant and I would like to send their Traces to a different OTLP collector

nhart12:
Closes #756 

Instead of storing reference to current Tracer on the UrlSessionInstrumentationClass, move it to the UrlSessionInstrumentationConfiguration object. This allows runtime changes to the underlying TracerProvider after the UrlSession has already been swizzled

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -43,8 +43,6 @@ public class URLSessionInstrumentation {
   ]
   .compactMap { NSClassFromString($0) }
 
-  public private(set) var tracer: Tracer
-
   public var startedRequestSpans: [Span] {
     var spans = [Span]()
     URLSessionLogger.runningSpansQueue.sync {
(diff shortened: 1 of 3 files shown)


nachoBonafonte (changes requested):
Even if there is a dictionary, it can add an unwanted extra delay on the check for every request (specially for the first one when it does not exist), I would prefer if it is initialized at the beginning and the update is only made in a new method that can be explicitly called when needed.

nhart12:
> Even if there is a dictionary, it can add an unwanted extra delay on the check for every request (specially for the first one when it does not exist), I would prefer if it is initialized at the beginning and the update is only made in a new method that can be explicitly called when needed.

Would it make sense then to include the Tracer as part of the URLSessionInstrumentationConfiguration?

nachoBonafonte:
Yes, it can also be a good option

PR #799, opened 2025-06-09
Clean up locks in record events readable spans

naftaly:
While working within the Embrace SDK, I noticed that `RecordEventsReadableSpan` used 2 different locks and one queue for synchronization. This could easily be replaced by one Read/Write lock.

### DIFF
--- a/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
+++ b/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift
@@ -9,27 +9,40 @@ import OpenTelemetryApi
 /// Implementation for the Span class that records trace events.
 public class RecordEventsReadableSpan: ReadableSpan {
   public var isRecording = true
+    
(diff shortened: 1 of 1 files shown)


bryce-b (approved):

nachoBonafonte:
The two-locks approach and a queue was done on purpose for performance reasons in the beginning, as it can be accessed by multiple threads. We may simplify it, specially now that span events are not much used, but we should measure the effects before changing it.

naftaly:
> The two-locks approach and a queue was done on purpose for performance reasons in the beginning, as it can be accessed by multiple threads. We may simplify it, specially now that span events are not much used, but we should measure the effects before changing it.

@nachoBonafonte How would you like to measure this? How does Otel using measure changes? fwiw, the chosen lock is going to be much faster than the 2 lock + queue approach, especially that there's no read contention.

bryce-b (approved):

nachoBonafonte (approved):

Issue #794, opened 2025-06-06
Cocoapods release is broken.


beekhc:
The 1.16.0 release did not publish cocoapods. The [action logs](https://github.com/open-telemetry/opentelemetry-swift/actions/runs/15473299330/job/43562878035) show it failed with this error:

```
Validating podspec
 -> OpenTelemetry-Swift-Api
 -> OpenTelemetry-Swift-Api (1.16.0)
    - ERROR | [iOS] unknown: Encountered an unknown error (Pod::DSLError) during validation.

[!] The spec did not pass validation, due to 1 error.
```

PR #801, opened 2025-06-10
fix: replace circular in podspec with correct dependency.

beekhc:
I think this _might_ be the underlying cause of issue #794. It fixes a different problem that gets flagged with `pod spec lint`. When I modify cocoapods to print proper error messages and run `pod spec lint OpenTelemetry-Swift-Api.podspec --verbose`, it flags this same error, even though it's in a different file.

`pod spec lint` still fails, because it's loading the released versions of the specs from github instead of the local versions.

However, `pod spec lint --quick` fails before this change, and passes after.

I'm trying to find a better way to test these, but since this fix is so obvious, I don't think it should wait.

### DIFF
--- a/OpenTelemetry-Swift-Instrumentation-URLSession.podspec
+++ b/OpenTelemetry-Swift-Instrumentation-URLSession.podspec
@@ -17,7 +17,7 @@ Pod::Spec.new do |spec|
   spec.watchos.deployment_target = "6.0"
   spec.module_name = "URLSession"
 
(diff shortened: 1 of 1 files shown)


ArielDemarco:
Agree! Fix seems obvious. Also, it matches the dependency hierarchy in SPM:
```swift
...
        .target(name: "URLSessionInstrumentation",
                dependencies: ["OpenTelemetrySdk", "NetworkStatus"],
                path: "Sources/Instrumentation/URLSession",
                exclude: ["README.md"]),
...
```

PR #809, opened 2025-06-13
Remove old metrics and replace with StableMetrics

bryce-b:
This PR completely removes the old metrics and replaces the tests and various usages with StableMetrics
including: 
- Prometheus exporter 
- persistent metric exporter
- swift metrics 

Other details: 

* all files related to the original metrics have been removed (all files in `metrics/` except under `metrics/stable`) 
* The files in `metrics/stable` have been moved up a level into `metrics`
* stable metrics `PointData` child objects have been extended to support `codable` to enable their use with [Sources/Exporters/Persistence/PersistenceMetricExporterDecorator.swift](https://github.com/open-telemetry/opentelemetry-swift/pull/809/files#diff-950e679a06f673b01fb47075a4f5873fc755bcaf1fa68890032f12696d51d563)
  * if this is too noisy, I can break this out into a separate change.
  * Test exercising this change are here : 
  https://github.com/open-telemetry/opentelemetry-swift/pull/809/files#diff-a45fa4ec76dac519209de328dcdaea417eb2cf599cbc9b4d35943e5356b4c62d
   and
 https://github.com/open-telemetry/opentelemetry-swift/pull/809/files#diff-fba1e960fca3f7ad1655e7754ebc8cf1c843a72d3e26620036c551a38311de66


nachoBonafonte (approved):
Really great work, it has really take long getting here but you did it, congrats!!!
There is one thing I am not sure if should also be added or what are the drawback, just remove all the Stable from the name , eg StableMeterProviderSdk  ->  MeterProviderSdk and keep a type alias with the Stable prefix, like:

```
public typealias StableMeterProvider = MeterProvider

public protocol MeterProvider: AnyObject {
```
 and 
```
public  typealias StableMeterProviderSdk = MeterProviderSdk

public class MeterProviderSdk: MeterProvider {
```

 But it can also be done in a separate PR.

bryce-b:
@nachoBonafonte I thought that might be too noisy to include in this PR--I was trying to make it a little easier to review. Next PR will remove the term `stable` from the metrics SDK.

Note: I want to do one more release before merging this PR

PR #810, opened 2025-06-14
Safeguard parameters in BatchSpanProcessor init

krayc425:
## Summary

- Add parameter validation and fallback defaults to `BatchSpanProcessor` initializer.
- Use `assert` statements in debug builds to catch invalid input.
- Automatically fallback to documented default values if input is invalid.
- This improves the library's robustness and prevents misconfiguration from causing runtime failures, while keeping the production behavior silent and predictable.

## Details

- If `scheduleDelay` < 0, fallback to 5.
- If `exportTimeout` < 0, fallback to 30.
- If `maxQueueSize` <= 0, fallback to 2048.
- If `maxExportBatchSize` <= 0, fallback to 512.

## Motivation

This change was motivated by a real-world crash. One time I accidentally passed `maxExportBatchSize = 0` when initializing a `BatchSpanProcessor`, which caused the app to crash at runtime, which root cause is the following code in `BatchWorker`:

```swift
stride(from: 0, to: spanList.endIndex, by: maxExportBatchSize).forEach { ... }
```

If `maxExportBatchSize` is less than or equal to 0, the stride step becomes invalid, causing an immediate crash.

To prevent this, we now validate all parameters and fallback to safe defaults if any are invalid, ensuring that the stride step and other logic always receive a positive value.

## Unit tests

```
Test Case '-[OpenTelemetrySdkTests.BatchSpansProcessorTests testInitializeWithDefaultParameters]' started.
Test Case '-[OpenTelemetrySdkTests.BatchSpansProcessorTests testInitializeWithDefaultParameters]' passed (0.000 seconds).
Test Case '-[OpenTelemetrySdkTests.BatchSpansProcessorTests testInitializeWithInvalidParameters]' started.
Test Case '-[OpenTelemetrySdkTests.BatchSpansProcessorTests testInitializeWithInvalidParameters]' passed (0.000 seconds).
Test Case '-[OpenTelemetrySdkTests.BatchSpansProcessorTests testInitializeWithValidParameters]' started.
Test Case '-[OpenTelemetrySdkTests.BatchSpansProcessorTests testInitializeWithValidParameters]' passed (0.000 seconds).

Test Case '-[OpenTelemetrySdkTests.BatchLogRecordProcessorTests testInitializeWithDefaultParameters]' started.
Test Case '-[OpenTelemetrySdkTests.BatchLogRecordProcessorTests testInitializeWithDefaultParameters]' passed (0.001 seconds).
Test Case '-[OpenTelemetrySdkTests.BatchLogRecordProcessorTests testInitializeWithInvalidParameters]' started.
Swift/Integers.swift:3268: Fatal error: Not enough bits to represent the passed value
Test Case '-[OpenTelemetrySdkTests.BatchLogRecordProcessorTests testInitializeWithInvalidParameters]' passed (0.000 seconds).
Test Case '-[OpenTelemetrySdkTests.BatchLogRecordProcessorTests testInitializeWithValidParameters]' started.
Test Case '-[OpenTelemetrySdkTests.BatchLogRecordProcessorTests testInitializeWithValidParameters]' passed (0.000 seconds).
```

_No breaking changes._

### DIFF
--- a/Sources/OpenTelemetrySdk/Logs/Processors/BatchLogRecordProcessor.swift
+++ b/Sources/OpenTelemetrySdk/Logs/Processors/BatchLogRecordProcessor.swift
@@ -45,7 +45,12 @@ public class BatchLogRecordProcessor: LogRecordProcessor {
       safeMaxExportBatchSize = 512
     }
 
-    worker = BatchWorker(logRecordExporter: logRecordExporter, scheduleDelay: scheduleDelay, exportTimeout: exportTimeout, maxQueueSize: maxQueueSize, maxExportBatchSize: maxExportBatchSize, willExportCallback: willExportCallback)
+    worker = BatchWorker(logRecordExporter: logRecordExporter,
+                         scheduleDelay: safeScheduleDelay,
+                         exportTimeout: safeExportTimeout,
+                         maxQueueSize: safeMaxQueueSize,
+                         maxExportBatchSize: safeMaxExportBatchSize,
+                         willExportCallback: willExportCallback)
 
     worker.start()
   }
--- a/Sources/OpenTelemetrySdk/Logs/Processors/BatchLogRecordProcessor.swift
+++ b/Sources/OpenTelemetrySdk/Logs/Processors/BatchLogRecordProcessor.swift
@@ -9,7 +9,42 @@ import OpenTelemetryApi
 public class BatchLogRecordProcessor: LogRecordProcessor {
   fileprivate var worker: BatchWorker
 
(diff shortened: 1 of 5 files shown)


bryce-b:
@krayc425 could you add a few unit tests exercising these changes?

krayc425:
Hi @bryce-b I've updated to not use `assert` (since it'll crash the test immediately, and I don't see other usages of it in the codebase) added a unit test for it, by exposing these properties internally.

bryce-b (approved):

krayc425:
Hey @bryce-b I just noticed that this change also applies to `BatchLogRecordProcessor` so I added it there as well. Mind taking another look? 🙇

PR #818, opened 2025-06-19
Add OSSignposterIntegration to use OSSignposter API

### LINKED ISSUE #582: os_signpost implementation is deprecated, add support for OSSignposter
Your two usages of os_signpost APIs are considered 'legacy' and deprecated.

Usage:
https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Instrumentation/SignPostIntegration/SignPostIntegration.swift

References:
https://developer.apple.com/documentation/os/3019242-os_signpost
https://developer.apple.com/documentation/os/3019241-os_signpost

Both state to use the new OSSignposter type, which is available as of iOS 15.
https://developer.apple.com/documentation/os/ossignposter

krayc425:
## Summary

- This PR introduces a new `OSSignposterIntegration` span processor, which leverages the modern [OSSignposter](https://developer.apple.com/documentation/os/ossignposter) API available on iOS 15+, macOS 12+, tvOS 15+, and watchOS 8+.
- Resolves #582

## Details

- Adds `OSSignposterIntegration` as an alternative to the existing `SignPostIntegration` for platforms supporting the new API.
- This implementation provides improved integration with Apple’s latest signpost APIs and more efficient span interval handling.
- Use a dictionary to store the states and ensures thread safety with a serial queue.
- Maintains backward compatibility: users can choose the appropriate integration at runtime using `#available` checks.
- We can't remove the old `SignPostIntegration` yet since the new API is only available on iOS 15.0+, while this project's iOS minimum version is 13.0.

## Usage

```swift
if #available(iOS 15, macOS 12, tvOS 15, watchOS 8, *) {
    OpenTelemetry.instance.tracerProvider.addSpanProcessor(OSSignposterIntegration())
} else {
    OpenTelemetry.instance.tracerProvider.addSpanProcessor(SignPostIntegration())
}
```

## Screenshot

Screenshot 2025-06-22 at 16 40 26

### DIFF
--- a/Examples/OTLP Exporter/main.swift
+++ b/Examples/OTLP Exporter/main.swift
@@ -42,7 +42,10 @@
 
   let tracer = OpenTelemetry.instance.tracerProvider.get(instrumentationName: instrumentationScopeName, instrumentationVersion: instrumentationScopeVersion)
 
-  if #available(macOS 10.14, *), #available(iOS 12.0, *) {
+  if #available(macOS 12, *), #available(iOS 15.0, *) {
+    let tracerProviderSDK = OpenTelemetry.instance.tracerProvider as? TracerProviderSdk
+    tracerProviderSDK?.addSpanProcessor(OSSignposterIntegration())
+  } else {
     let tracerProviderSDK = OpenTelemetry.instance.tracerProvider as? TracerProviderSdk
     tracerProviderSDK?.addSpanProcessor(SignPostIntegration())
   }
--- a/Examples/OTLP Exporter/main.swift
+++ b/Examples/OTLP Exporter/main.swift
@@ -42,7 +42,7 @@
 
   let tracer = OpenTelemetry.instance.tracerProvider.get(instrumentationName: instrumentationScopeName, instrumentationVersion: instrumentationScopeVersion)
 
-  if #available(macOS 12, *), #available(iOS 15.0, *) {
+  if #available(iOS 15.0, macOS 12, tvOS 15.0, watchOS 8.0, *) {
     let tracerProviderSDK = OpenTelemetry.instance.tracerProvider as? TracerProviderSdk
     tracerProviderSDK?.addSpanProcessor(OSSignposterIntegration())
   } else {
--- a/Sources/Instrumentation/SignPostIntegration/OSSignposterIntegration.swift
+++ b/Sources/Instrumentation/SignPostIntegration/OSSignposterIntegration.swift
@@ -1,0 +1,41 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#if !os(watchOS)
+
+  import Foundation
+  import os
+  import OpenTelemetryApi
+  import OpenTelemetrySdk
+
+  /// A span processor that decorates spans with the origin attribute
+  @available(macOS 12, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+  public class OSSignposterIntegration: SpanProcessor {
+
+    private static var osSignpostStateKey: UInt8 = 0
+
+    public let isStartRequired = true
+    public let isEndRequired = true
+    public let osSignposter = OSSignposter(subsystem: "OpenTelemetry", category: .pointsOfInterest)
+
+    public init() {}
+
+    public func onStart(parentContext: SpanContext?, span: ReadableSpan) {
+      let signpostID = osSignposter.makeSignpostID()
+      let osSignpostState = osSignposter.beginInterval("Span", id: signpostID, "\(span.name, privacy: .public)")
+      objc_setAssociatedObject(span, &OSSignposterIntegration.osSignpostStateKey, osSignpostState, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+    }
+
+    public func onEnd(span: ReadableSpan) {
+      if let state = objc_getAssociatedObject(span, &OSSignposterIntegration.osSignpostStateKey) as? OSSignpostIntervalState {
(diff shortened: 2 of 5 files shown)


Sources/Instrumentation/SignPostIntegration/OSSignposterIntegration.swift
```diff
@@ -0,0 +1,41 @@
...
+    public let isStartRequired = true
+    public let isEndRequired = true
+    public let osSignposter = OSSignposter(subsystem: "OpenTelemetry", category: .pointsOfInterest)
+
+    public init() {}
+
+    public func onStart(parentContext: SpanContext?, span: ReadableSpan) {
+      let signpostID = osSignposter.makeSignpostID()
+      let osSignpostState = osSignposter.beginInterval("Span", id: signpostID, "\(span.name, privacy: .public)")
+      objc_setAssociatedObject(span, &OSSignposterIntegration.osSignpostStateKey, osSignpostState, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
```
naftaly, line 28: Preferably, you'd simply use a dictionary here to map spanId to state and just use .exclusive for the signpostId.

```
// as class member
var spanIdToStateMap: [String: OSSignpostIntervalState] = []

// start
let state = osSignposter.beginInterval("Span", id: .exclusive, "\(span.name, privacy: .public)")
spanIdToStateMap[span.spanId.hexString] = state

// end
if let state = spanIdToStateMap.remove(span.spanId.hexString) {
      osSignposter.endInterval("Span", state)
}
```

naftaly: By looking at the callers of SpanProcessors, it's hard to tell right away if this is expected to be thread safe or not. It might be a good idea to document that part and if it needs to be, add a lock around the resource.

naftaly: Edit: I just verified and SpanProcessors receive `onStart` and `onEnd` and undermined threads, so we definitely need some sort of lock here.

naftaly: > Actually that is the main reason why I avoided a dictionary here. Since a shared dictionary would require a lock or serial queue for thread safety, it introduces data race and can negatively affect the precision of span interval timestamps.

Understood and makes total sense. However, you don't own the span so anyone could in theory break your connection if they wanted to. On top of that, if the span is a Value type, the Swift runtime will box the value into an ObjC object and you might not get back what you expect since the span isn't really what you're dealing with anymore.

That aside, at a minimum use ATOMIC instead of NON_ATOMIC. Otherwise, I'd just simply use a dictionary and an unfair lock, you have much contention if at all, but it will save users from scary bugs that are hard to find/catch.

If worried, you can try and measure it, but I doubt there will be any if much of a difference between the two implementations outside of thread safety.

krayc425: Your points about object ownership and the possibility of the span being a value type make sense to me too, and I agree that explicit locking around a shared dictionary is perhaps the most predictable solution for thread safety.

I’ll update the implementation to use a dictionary instead 🤞 Thanks again for the additional context and feedback!

krayc425:
@naftaly I've updated to use a dictionary and queue. Could you take another 👀 ? Ty!

Examples/OTLP Exporter/main.swift
```diff
@@ -42,7 +42,10 @@
 
   let tracer = OpenTelemetry.instance.tracerProvider.get(instrumentationName: instrumentationScopeName, instrumentationVersion: instrumentationScopeVersion)
 
-  if #available(macOS 10.14, *), #available(iOS 12.0, *) {
+  if #available(macOS 12, *), #available(iOS 15.0, *) {
```
naftaly, line 45: If I'm not mistaken, you can do `#available(iOS 15.0, macOS 12, *)`

krayc425: good catch! updated

Sources/Instrumentation/SignPostIntegration/OSSignposterIntegration.swift
```diff
@@ -0,0 +1,44 @@
...
+    public let osSignposter = OSSignposter(subsystem: "OpenTelemetry", category: .pointsOfInterest)
+    public let ossignposterQueue = DispatchQueue(label: "org.opentelemetry.ossignposterIntegration")
+    private var spanIdToStateMap: [String: OSSignpostIntervalState] = [:]
+
+    public init() {}
+
+    public func onStart(parentContext: SpanContext?, span: ReadableSpan) {
+      let state = osSignposter.beginInterval("Span", id: .exclusive, "\(span.name, privacy: .public)")
+      ossignposterQueue.sync {
+        spanIdToStateMap[span.context.spanId.hexString] = state
```
naftaly, line 28: probably not gonna happen, but what if the same span starts twice. Do you think we should check for that case?

naftaly:
> @naftaly I've updated to use a dictionary and queue. Could you take another 👀 ? Ty!

Great stuff, thanks for taking care of it.

naftaly (approved):

Sources/Instrumentation/SignPostIntegration/OSSignposterIntegration.swift
krayc425: I think we could potentially update the key to `traceId-spanId` so that it's truly unique. wdyt?

krayc425:
Hi @bryce-b , could you help take a look at this one? Appreciated!

PR #821, opened 2025-06-20
chore: Make StdoutLogExporter public

limhjgrace:
## Description

This PR makes the `StdoutLogExporter` class public so that it can be used outside of this repo. Similar to how the `StdoutSpanExporter` class can be used for local debugging, applications/libraries consuming OTel SDK will also be able to use this class for local development as well.

## Testing
Using local iOS app that uses OTel SDK, confirmed that class is now accessible/usable.


bryce-b (approved):

bryce-b:
@limhjgrace can you sign the CLA linked above? I won't be able to merge this until you do.

PR #822, opened 2025-06-20
added public init to SpanData to support custom Span Processors

bryce-b:
SpanData's init is internal, which make it impossible to support Elastic's custom span processor. 

The recent changes to the tracing sdk made the 'isRecording' flag immutable, which prevents the hack I was using to update the span in my custom processor after calling `span.end()`.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/grpc/grpc-swift.git",
       "state" : {
-        "revision" : "8c5e99d0255c373e0330730d191a3423c57373fb",
-        "version" : "1.24.2"
+        "revision" : "a56a157218877ef3e9625f7e1f7b2cb7e46ead1b",
+        "version" : "1.26.1"
       }
     },
     {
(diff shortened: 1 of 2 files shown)


nachoBonafonte:
I have doubts if this is the change we want or if it would be preferable to let the span be modified in the span processor as before (or with other approach), and let the SpanData only be generated from the Span itself. This approach seems a bit hacky to me. Have you checked other implementations?

bryce-b:
The use case here is to provide a means to add additional (global) attributes in the span processor. This is to support things like `session.id`. The issue is when a span is ended [internalIsRecording](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift#L294) is set to false making all further changes to the span noops. Originally, I was able to get around this by creating an extension that temporarily sets the value to true, to allow attributes to be applied, this is no longer possible with the latest changes to the tracer. 
My work around now was going to create a mutable span that implements the `ReadableSpan` protocol, but I cannot implement the `toSpanData` method because the SpanData struct is using the default internal initializer. 

Alternatively, I could look into apply attributes on spanStart, where this contention shouldn't arise.

Issue #802, opened 2025-06-11
Internally defined DataCompression conflicts with externally defined DataCompression lib


djpearce:
Hello and happy WWDC week!

The internally scoped DataCompression library conflicts with the actual separately defined [DataCompression Library](https://github.com/mw99/DataCompression) it was inspired by.

**Problem**
My repository utilizes [LaunchDarkly](https://github.com/launchdarkly/ios-client-sdk/blob/v9/Package.swift) for Feature Flags.

LaunchDarkly's iOS SDK now references the actual DataCompression library that this repository references:

```swift
    dependencies: [
        //...
        .package(name: "DataCompression", url: "https://github.com/mw99/DataCompression", .exact("3.8.0"))
    ],
```

When attempting to update to the latest LaunchDarkly iOS SDK it now conflicts as OpenTelemetry defines its own library with the same name.

**Proposed Solution**

Rename the `DataCompression` library to something else like `DataCompressionUtils`.

or

Work within the actual `DataCompression` [repository](https://github.com/mw99/DataCompression) and update it to support (or at least compile) Linux based Swift as well and then use that directly.

nachoBonafonte:
Yes, this problem has already been reported in the past, we plan to rename it in the future, but we are waiting for a major version update that we expect will happen soon, in the meantime there is a branch you can use: `datacompression-rename` which already has the latest version released changes that includes a renamed data compression library

dibeakshay:
Hi @nachoBonafonte,

We're encountering a linker error while working with the datacompression-rename branch and integrating OpenTelemetryProtocolExporterGrpc.

The specific error is:

Undefined symbol: NIOHPACK.HPACKHeaders.init([(Swift.String, Swift.String)]) -> NIOHPACK.HPACKHeaders

It appears this issue stems from NIOHPACK not being exposed as a public product within the swift-nio-http2 package, which is a dependency of OpenTelemetryProtocolExporterGrpc.

Could you please advise if there's a recommended workaround for this?

<img width="959" alt="Image" src="https://github.com/user-attachments/assets/a347cfbc-ed35-4b2f-a447-62762d6b56b6" />

nachoBonafonte:
No sorry, no previous report about it, and that library is also not under our control. The workaround would be checking version differences (if it is also included by other package not related to OpenTelemetry) or directly use the http exporter instead of the grpc one.

amulRTR:
Hey @nachoBonafonte,

[...]

PR #838, opened 2025-06-30
Depend on the real DataCompression library

naftaly:
#802 

Depend on the real DataCopression lib.


ArielDemarco (commented):
I think this change is ok. Only thing to take into consideration is that this is a breaking change. But I understand its generating issues to customers. I think we've been discussing this in the SIG @nachoBonafonte / @bryce-b / @vvydier.

naftaly:
@nachoBonafonte @bryce-b updates made to depend on the real DataCompression lib.

PR #850, opened 2025-07-03
Rename RecordEventsReadableSpan to SpanSdk

nachoBonafonte:
Rename RecordEventsReadableSpan to SpanSdk to more clearly identity the default span type of the SDK


ArielDemarco (commented):
I'm ok with the change. Probably it might generate some kind of conflict/breaking change; but seems an easy-to-solve problem as Xcode will help with the rename. 
@bryce-b / @vvydier what do you guys think?

bryce-b (approved):

bryce-b:
since we are going to be release 2.0 soon with the changes to metrics, now is the time to get breaking changes in.

PR #852, opened 2025-07-03
removed 'stable' term from metrics, left deprecated typealiases

bryce-b:


Sources/OpenTelemetrySdk/Metrics/Data/ExemplarData.swift
```diff
@@ -94,7 +94,7 @@ public final class LongExemplarData: ExemplarData, Codable {
   }
 
   public func encode(to encoder: any Encoder) throws {
-    var container = try encoder.container(keyedBy: CodingKeys.self)
```
bryce-b, line 97: This was an unnecessary `try`

nachoBonafonte (approved):

PR #864, opened 2025-07-14
Fix race in TraceProviderSDK

naftaly:
Race condition: https://gist.github.com/naftaly/f5cda479adf1451da63a38e8672b3fb4

Screenshot 2025-07-14 at 4 58 07 PM


Sources/OpenTelemetrySdk/Trace/TracerProviderSdk.swift
```diff
@@ -158,3 +144,63 @@ public class TracerProviderSdk: TracerProvider {
...
+        }
+        return try body()
+    }
+    
+    // specialise Void return (for performance)
+    @inlinable
+    func withLockVoid(_ body: () throws -> Void) rethrows {
+        try withLock(body)
+    }
+}
```
bryce-b, line 206: looks good to me, please run this through swift-format, looks like you're using 4-space tabs, instead of 2.

bryce-b (approved):

ArielDemarco (approved):

naftaly:
Is anyone able to merge this branch, I don't have that access. Thanks!

naftaly:
@nachoBonafonte what implementation would you propose? As it was before with the r/w lock there was a race condition. As proposed, the lock will do just fine and perform very well as it is very constrained to a small area of concern which is processed quickly. I am not worried about performance by adding this lock.

naftaly:
It's true that the rw lock wasn't correctly implemented, but it doesn't need to be in a class. That being said, the race condition I'm pointing out is that the lock that was there previously held a read in order to check for the existence of the shared provider. Then it released the lock. Afterwards, did some work and took the write lock to set that new tracer. The time between the read and the write holds o the lock are the race condition I'm talking about.

So basically, this PR fixes the issue you mention and the issue I mention.

nachoBonafonte:
>The time between the read and the write holds o the lock are the race condition I'm talking about.
And what is the memory being accessed concurrently there? I would say none, and it there were something being written, it should be inside the write lock.

nachoBonafonte:
And even if we used a standard lock (which I don't think is the proper solution), we should use the one existing in the Locks.swift file I mentioned, and not reimplement it.

naftaly:
> And what is the memory being accessed concurrently there?

It's not concurrent access. The race condition in this case is that you can read and it's nil before another write happens, then the first write happens and that previous write will write a second time, overwriting the previous one leading to multiple instances of the provider with no clear owner.

> And even if we used a standard lock (which I don't think is the proper solution), we should use the one existing in the Locks.swift file I mentioned, and not reimplement it.

I wanted to, but it would create a circular dep. We could if we moved it around a bit, but I thought it would be problematic to do that in this PR, but once we fix this issue, we can definitely do it.

naftaly:
@nachoBonafonte In the spirit of making speeding things up and coming to a compromise, I've updated the lock to the class so it actually works and eliminates one of the race conditions.

Sound good to you?

PR #871, opened 2025-07-17
Fix thread race conditions in new metrics

nachoBonafonte:
Updated SDK locks to use a more similar to Apple Concurrency's Mutex approach, which allows being used in the future when we transition to Swift 6. It also allows the compiler to help with the errors as the type itself includes the lock.
Removed inout to properties that dont need it.
Updated some properties to constants when mutability is not needed
Cleaned also most warnings

### DIFF
--- a/Sources/OpenTelemetrySdk/Metrics/InstrumentBuilder.swift
+++ b/Sources/OpenTelemetrySdk/Metrics/InstrumentBuilder.swift
@@ -16,7 +16,7 @@ public class InstrumentBuilder {
   var instrumentName: String
   var explicitBucketBoundariesAdvice: [Double]?
 
-  init(meterProviderSharedState: inout MeterProviderSharedState, meterSharedState: inout MeterSharedState, type: InstrumentType, valueType: InstrumentValueType, description: String, unit: String, instrumentName: String) {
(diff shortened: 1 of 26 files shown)


ArielDemarco (approved):

Sources/OpenTelemetrySdk/Metrics/InstrumentBuilder.swift
```diff
@@ -40,8 +40,8 @@ public extension InstrumentBuilder {
     return self
   }
 
-  internal func swapBuilder<T: InstrumentBuilder>(_ builder: (inout MeterProviderSharedState, inout MeterSharedState, String, String, String) -> T) -> T {
```
bryce-b, line 43: I believe if these are not `inout` then the internal `readerStorageRegisteries` and `callbackRegistration` will not be shared and updated in the MeterProvider

nachoBonafonte: But they are classes aren't they, the instances themselves are not changing, I thought it was a remaining if they were structs during development.

bryce-b: yes, that may have been the issue.

bryce-b (approved):

PR #873, opened 2025-07-20
fix: crash when swizzle almofire uploadTask

tomerhy:
## Fix: Safe and Type-Correct Swizzling for URLSession.uploadTask(with:from:) and uploadTask(with:fromFile:)
🐛 **Problem**
The previous implementation incorrectly reused a single block (with a generic or shared signature) across multiple method swizzles:

`uploadTask(with:from:)` and `uploadTask(with:fromFile:)` have different parameter types (Data? vs. URL), but the swizzling logic used a shared block signature with AnyObject or similar abstraction.

This violated Swift’s type safety and led to unsafe pointer casting and incorrect calling conventions, which could result in crashes inside CoreFoundation / CFNetwork, including:

```
CFNetwork → CFURLConnectionSendSynchronousRequest
CFNetwork → CFHostCreateWithName
```
Additionally:

originalIMP was captured before being assigned, potentially leading to null dereferencing.

self (instrumentation) was captured strongly inside the block, causing retain cycles and access-after-deinit crashes.

✅ **Solution**
Defined distinct, type-safe blocks for each method:

One for (URLRequest, Data?)

One for (URLRequest, URL)

Performed unsafeBitCast of originalIMP outside the block to avoid closure capture issues.

Used @convention(block) and @convention(c) with correct signatures to ensure ABI safety.

Captured self weakly ([weak self]) to avoid memory leaks and retain cycles.

Fallbacks to the original IMP safely if self is no longer alive.

🔐 **Safety Improvements**
Prevents incorrect method dispatch due to mismatched block signatures.

Eliminates retain cycles by weakly capturing instrumentation logic.

Ensures valid and initialized method pointer is always used.

Compatible with SDK teardown, instrumentation lifecycle, and external wrappers like Alamofire.

🧪  **Testing**
Validated against real-world upload cases using URLSession and Alamofire (multipart form uploads).

Confirmed injected request headers and task tracking function correctly.

Crash involving CFHostCreateWithName is no longer reproducible.

No regressions in task creation or header propagation.


bryce-b:
@tomerhy pinging to bring attention to the comments

PR #877, opened 2025-07-27
Fix OtlpHttpTraceExporter request headers.

Waheiba:
`createRequest(body:endpoint:)` in OtlpHttpExporterBase already considers `envVarHeaders` and `config.headers`
Additional looping on those headers in `OtlpHttpTraceExporter` after the request is created is appending the same value again.


nachoBonafonte (approved):

Waheiba:
I updated the test for linux.
the header casing was actually different when running in a linux environment to from `testHeader` to `Testheader`.
I'm changing the filter to be all lowercased.
Looks less pretty tho but gets away with any case sensitive issues for the headers

vvydier (approved):

Issue #486, opened 2023-11-20
Splitting APIs in their own repo to allow custom implementations


ganeshnj:
## Motivation

Increasingly customers are asking for OpenTelemetry support in Datadog SDKs at the same time using other Datadog SDK features. With the current setup of the opentelemetry-swift SDK, it is impossible to take dependency on API only packages. This becomes a bigger problem for mobile developers (dd-sdk-ios users) where Xcode loads and indexes all the packages in the workspace and slows down the entire development experience.

## Proposal

Split the APIs from the implementation in opentelemetry-swift. Ideally there must exist two swift package repositories:
1. opentelemetry-swift-api (subject to change name)
2. opentelemetry-swift

I experimented the approach at https://github.com/ganeshnj/opentelemetry-swift-sdk & https://github.com/ganeshnj/opentelemetry-swift-api and it seems to work fine. I was able to take dependency on opentelemetry-swift-api and use the API without any issues.

It does comes with some challenges such as versioning, release and breaking changes (will be clearly visible) but I think it is worth it. It also goes with OpenTelemetry's [requirements](https://opentelemetry.io/docs/specs/otel/library-guidelines/#requirements).

> The OpenTelemetry API must be well-defined and clearly decoupled from the implementation. This allows end users to consume API only without also consuming the implementation (see points 2 and 3 for why it is important).

> The developers of the final application normally decide how to configure OpenTelemetry SDK and what extensions to use. They should be also free to choose to not use any OpenTelemetry implementation at all, even though the application and/or its libraries are already instrumented

I'm happy to discuss this in the next Swift otel sync.

bryce-b:
We discussed this issue during the Swift-SIG, and I will summarize our discussion: 
We believe that this issue is a matter of opinion, and doesn't provide significant value for the trade-offs, so we will not be implementing this requested change. However, I will leave this ticket open in the case there are more interested parties that would like to see this change made. Nacho also mentioned potentially using git submodules to only import the open-telemetry-api portion of the project to avoid adding all the additional dependencies in the sdk.

ganeshnj:
Thanks for quick summar @bryce-b

Using submodules looks counterintuitive from package management point of view (not the preferred choice here) but I can see a fork which only contains the APIs. We can eventually include support for other Swift/iOS package managers.

ancostas:
[...]

PR #885, opened 2025-08-08
[REPO-SPLIT-PART-1] API+SDK 

naftaly:
To reduce the clutter and the number of deps that get downloaded with SPM, we're going to split out API + SDK from the rest. This the first go at it and I'd like your opinion on the split.

- Is anything missing that we should keep in this repo?
- I left a few pieces here, mostly those without any deps, does this work for you or shall I move them somewhere else.
- I welcome any other feedback.

Decision: https://docs.google.com/document/d/1LugL8r4bAkbTxZ1Gq_6j_8SDv1LuZaNeUdR-QLN4d48/edit?disco=AAABl-_cxiU

--

A few issues that might be helped by this:
https://github.com/DataDog/opentelemetry-swift-packages/issues/31
https://github.com/open-telemetry/opentelemetry-swift/issues/486

### DIFF
--- a/Sources/OpenTelemetryApi/Context/OpenTelemetryContextProvider.swift
+++ b/Sources/OpenTelemetryApi/Context/OpenTelemetryContextProvider.swift
@@ -12,7 +12,7 @@ public enum OpenTelemetryContextKeys: String {
 }
 
 public struct OpenTelemetryContextProvider {
-  package var contextManager: ContextManager
+  public var contextManager: ContextManager
 
   /// Returns the Span from the current context
   public var activeSpan: Span? {
--- a/Sources/OpenTelemetryApi/Context/TaskLocalContextManager.swift
+++ b/Sources/OpenTelemetryApi/Context/TaskLocalContextManager.swift
@@ -16,7 +16,7 @@ import Foundation
   /// If you require dynamic context injection, you will need a custom context manager.
   @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
   public class TaskLocalContextManager: ContextManager {
-    package static let instance = TaskLocalContextManager()
+    public static let instance = TaskLocalContextManager()
 
     @TaskLocal static var context = [String: AnyObject]()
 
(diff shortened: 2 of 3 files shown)


Package.swift
```diff
@@ -17,29 +17,15 @@ let package = Package(
...
+    .executable(name: "LoggingTracer", targets: ["LoggingTracer"]),
+    .executable(name: "ConcurrencyContext", targets: ["ConcurrencyContext"])
   ],
   dependencies: [
-    .package(url: "https://github.com/apple/swift-nio.git", from: "2.83.0"),
-    .package(url: "https://github.com/grpc/grpc-swift.git", exact: "1.26.1"),
-    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.30.0"),
-    .package(url: "https://github.com/apple/swift-log.git", from: "1.6.3"),
-    .package(url: "https://github.com/apple/swift-metrics.git", from: "2.7.0"),
     .package(url: "https://github.com/apple/swift-atomics.git", from: "1.3.0"),
```
ArielDemarco, line 28: We can remove this considering you made the abstraction yourself

Sources/OpenTelemetryApi/Context/OpenTelemetryContextProvider.swift
```diff
@@ -12,7 +12,7 @@ public enum OpenTelemetryContextKeys: String {
 }
 
 public struct OpenTelemetryContextProvider {
-  package var contextManager: ContextManager
+  public var contextManager: ContextManager
```
ArielDemarco, line 15: Note: this probably means we can remove the line:
```ruby
spec.pod_target_xcconfig = { "OTHER_SWIFT_FLAGS" => "-module-name OpenTelemetryApi -package-name opentelemetry_swift" }
```
from the `OpenTelemetry-Swift-Api.podspec` file

__OTHER_REPO__/Makefile
```diff
@@ -0,0 +1,101 @@
+PROJECT_NAME="opentelemetry-swift-Package"
```
ArielDemarco, line 1: For later: we'll have to change this to match the name of the package of the new project

PR #889, opened 2025-08-14
Added tests to `uploadTask` instrumentation

ArielDemarco:
# Overview
Added some tests to prevent [this kind of issues](https://github.com/open-telemetry/opentelemetry-swift/pull/873) when using `URLSessionInstrumentation`

## Extra
-> Tests were failing due to changes in how the [macos-15 runners work](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md#installed-simulators). Basically, [runtimes were removed](https://github.com/actions/runner-images/issues/12758#issuecomment-3187480561) and the default Xcode behavior changed. Therefore, I added a step to explicitly select a specific Xcode version (in this case, Xcode 16.4) and updated the `Makefile` so that each build/test uses the runtime associated with Xcode 16.4 (iOS/tvOS 18.5, watchOS 11.5, and visionOS 2.5).

### DIFF
--- a/.github/workflows/BuildAndTest.yml
+++ b/.github/workflows/BuildAndTest.yml
@@ -31,7 +31,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - uses: maxim-lobanov/setup-xcode@60606e260d2fc5762a71e64e74b2174e8ea3c8bd # v1.6.0
       with:
-        xcode-version: '16.0'
+        xcode-version: '16.4'
     - name: Build and Test for macOS
       run: swift test --enable-code-coverage
     - name: Upload Code coverage
@@ -46,7 +46,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - uses: maxim-lobanov/setup-xcode@60606e260d2fc5762a71e64e74b2174e8ea3c8bd # v1.6.0
       with:
-        xcode-version: '16.0'
+        xcode-version: '16.4'
     - name: Install Homebrew kegs
       run: make setup-brew
     - name: Build for iOS
@@ -59,7 +59,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - uses: maxim-lobanov/setup-xcode@60606e260d2fc5762a71e64e74b2174e8ea3c8bd # v1.6.0
       with:
-        xcode-version: '16.0'
+        xcode-version: '16.4'
     - name: Install Homebrew kegs
       run: make setup-brew
     - name: Build for tvOS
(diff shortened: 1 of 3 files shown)


erik-bershel:
Action has "latest-stable" label, which might be useful in your case. It will always follow the most updated platform.

erik-bershel:
In the current implementation, no - you've already done enough. 

It's just that sooner or later you'll need to switch to more recent versions of Xcode, I suppose, which means more recent versions of simulators. You can choose a more flexible Xcode selection system and make the target immediately non-specific to the OS version - Xcode itself will choose the latest compatible one. 

For example, for one of your targets:
```bash
    -destination 'platform=visionOS Simulator,name=Apple Vision Pro' \
```

And change Xcode selector target to latest-stable:
```yml
    - uses: maxim-lobanov/setup-xcode@v1
      with:
        xcode-version: latest-stable
```

If you don't have a reference to a specific version of the platform in your project code, this should work, plus you won't have to waste effort on further updates. CI will crash when it stops being compatible with the new Xcode/Platform version, which is already a good signal for testing.

bryce-b (approved):

ArielDemarco:
> In the current implementation, no - you've already done enough.
> 
> It's just that sooner or later you'll need to switch to more recent versions of Xcode, I suppose, which means more recent versions of simulators. You can choose a more flexible Xcode selection system and make the target immediately non-specific to the OS version - Xcode itself will choose the latest compatible one.
> 
> For example, for one of your targets:
> 
> ```shell
>     -destination 'platform=visionOS Simulator,name=Apple Vision Pro' \
> ```
> 
> And change Xcode selector target to latest-stable:
> 
> ```yaml
>     - uses: maxim-lobanov/setup-xcode@v1
>       with:
>         xcode-version: latest-stable
> ```
> 
> If you don't have a reference to a specific version of the platform in your project code, this should work, plus you won't have to waste effort on further updates. CI will crash when it stops being compatible with the new Xcode/Platform version, which is already a good signal for testing.

Applied those changes to all platforms.
The only one that didn’t work was using Apple Vision Pro directly without specifying the OS version:
```
xcodebuild: error: Unable to find a device matching the provided destination specifier:
		{ platform:visionOS Simulator, OS:latest, name:Apple Vision Pro }
	The requested device could not be found because multiple devices matched the request. (
----- xcbeautify -----
    "<DVTiPhoneSimulator: 0x1118362c0> {\n\t\tSimDevice: Apple Vision Pro (0E7A1ACF-F3A7-42D5-9A1A-0A638DA0B89A, visionOS 2.5, Shutdown)\n}",
Version: 2.30.1
    "<DVTiPhoneSimulator: 0x111836810> {\n\t\tSimDevice: Apple Vision Pro (5C15B214-9389-4187-BB81-9108A140FF4E, visionOS 26.0, Shutdown)\n}"
----------------------
)
```
So I’ll just specify the OS version for that one until we find a better solution.
Thanks for all the help on this!

PR #894, opened 2025-08-19
fix: OtlpHttpLogRecordExporter request headers

Waheiba:
This addresses a similar issue as https://github.com/open-telemetry/opentelemetry-swift/pull/877
Headers values are added twice which concatenates the same value again over the existing for the same header.

`createRequest(body:endpoint:)` in `OtlpHttpExporterBase` already considers `envVarHeaders` and `config.headers`
Additional looping on those headers in `OtlpHttpTraceExporter` after the request is created is appending the same value again.

Tests are also added

Apologies for the other PR that I closed. Small accident with commit history.

### DIFF
--- a/OpenTelemetry-Swift-Api.podspec
+++ b/OpenTelemetry-Swift-Api.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
   spec.name = "OpenTelemetry-Swift-Api"
-  spec.version = "2.0.0"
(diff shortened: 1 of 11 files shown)


bryce-b:
> Apologies for the other PR that I closed. Small accident with commit history.
I hate it when that happens :) 

Thanks for the contribution @Waheiba 
Could you extend this change to the other exporters as well?

bryce-b (approved):

Issue #916, opened 2025-09-09
Crash  in  os_log


fangandyuan:
`
        if #available(iOS 14, macOS 11, tvOS 14, *) {
          os_log(.info, "failed to initialize network connection status: %@", safeClasses.count)
        }
`

fix:

os_log(.info, "failed to initialize network connection status: %d", safeClasses.count)

zhangjie-spec:
i have the same problem in InstrumentationUtils
fix:
os_log(.info, "failed to initialize network connection status: \(safeClasses.count)")

PR #935, opened 2025-09-29
bugfix  %@  to  %d

fangandyuan:
This is a compilation issue


ArielDemarco:
Just to clarifiy: seems like this is not a compilation issue at all. Probably a runtime issue? (based on [this issue](https://github.com/open-telemetry/opentelemetry-swift/issues/916) you've submitted).
Also, check if you can do the CLA Authorization; it's needed in order to merge.

fangandyuan:
> Just to clarifiy: seems like this is not a compilation issue at all. Probably a runtime issue? (based on [this issue](https://github.com/open-telemetry/opentelemetry-swift/issues/916) you've submitted). Also, check if you can do the CLA Authorization; it's needed in order to merge.

Yes, it will cause a crash on iOS 14 machines.

bryce-b (approved):

PR #963, opened 2025-10-29
feat: Add instrumentation for MetricKit.

beekhc:
This is proposed instrumentation for Apple's MetricKit. The technical details for the instrumentation are thoroughly documented in the [README](https://github.com/beekhc/opentelemetry-swift/blob/beeklimt.metrickit/Sources/Instrumentation/MetricKit/README.md) file.

I am sure some details of the proposal will be controversial. I'd encourage you to particularly pay attention to:
* the shape of the telemetry:
  * a span for `MXMetricPayload` with attributes for all reported values
  * log events for each diagnostic in `MXDiagnosticPayload`
* the timestamps:
  * the `MXMetricPayload` span is a 24-hour span over the time the data was aggregated
  * log events are recorded as they are received
* the way `MXHistogram`s are converted to approximate averages

Of course any other feedback would also be appreciated.

### DIFF
--- a/.github/workflows/BuildAndTest.yml
+++ b/.github/workflows/BuildAndTest.yml
@@ -94,7 +94,7 @@ jobs:
       run: make test-without-building-visionos
   linux:
     runs-on: ubuntu-latest
-    container: swift:6.2@sha256:471acf16d3ce9b270d2f15ea177b0e150bdf1595ae2acfcc4932d1a03f43c3fd
+    container: swift:6.2@sha256:1e73c4051f095f7f1bafbece9ca7f9c67de4c870246c20bf12a06c69c52dd827
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       - name: Build tests for Linux
--- a/Examples/OTLP Exporter/docker-compose.yaml
+++ b/Examples/OTLP Exporter/docker-compose.yaml
@@ -24,7 +24,7 @@ services:
   # Prometheus
   prometheus:
     container_name: prometheus
-    image: prom/prometheus:latest@sha256:23031bfe0e74a13004252caaa74eccd0d62b6c6e7a04711d5b8bf5b7e113adc7
+    image: prom/prometheus:latest@sha256:49214755b6153f90a597adcbff0252cc61069f8ab69ce8411285cd4a560e8038
     volumes:
       - ./prometheus.yaml:/etc/prometheus/prometheus.yml
     ports:
--- a/Examples/OTLP HTTP Exporter/docker-compose.yaml
+++ b/Examples/OTLP HTTP Exporter/docker-compose.yaml
@@ -2,7 +2,7 @@ version: "3"
 services:
   # Collector
   collector:
-    image: otel/opentelemetry-collector:latest@sha256:56951db9579bf00d3f32a4e934e19548183a86c14640798502bcd4c225976ea6
+    image: otel/opentelemetry-collector:latest@sha256:8ac5df2a931e9264667b236d65bf7591fa4ba633a7a634e6caa2f0a4fc549c07
 #    The latest image of the otel-collector may not work, so specifying the version that works with this release
 #    image: otel/opentelemetry-collector:latest
     command: ["--config=/conf/collector-config.yaml"]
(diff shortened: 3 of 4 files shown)


ArielDemarco:
> I don't know why the macOS tests are failing. `MXMetricPayload` [should be available](https://developer.apple.com/documentation/metrickit/mxmetricpayload?language=objc) in macOS v12. And it works on my machine, albeit with a different SDK version.

I downloaded all macOS SDKs from version `26.0` down to Monterey (`12.0`), and the only one where the API is defined like this:
```objc
API_AVAILABLE(ios(13.0), macos(10.15)) API_UNAVAILABLE(tvos, watchos)
@interface MXMetricPayload : NSObject <NSSecureCoding>
```
is in macOS `26.0`.

After that, in all the others, it looks like this:
```objc
API_AVAILABLE(ios(13.0)) API_UNAVAILABLE(macos, tvos, watchos)
@interface MXMetricPayload : NSObject <NSSecureCoding>
```

Maybe it's related to this thing mentioned in the [MXMetricManager documentation](https://developer.apple.com/documentation/metrickit/mxmetricmanager):
image

So, seems that unless you compile it using the latest versions of Xcode (26.0 and up), this problem will happen

Sources/Instrumentation/MetricKit/MetricKitInstrumentation.swift
```diff
@@ -0,0 +1,608 @@
...
+        logForEach(payload.hangDiagnostics, "hang") {
+            let callStackTree = $0.callStackTree
+            let stacktraceJson = String(
+                decoding: callStackTree.jsonRepresentation(),
+                as: UTF8.self
+            )
+
+            let namespacedAttrs: [String: AttributeValueConvertable] = [
+                "hang_duration": $0.hangDuration,
+                "exception.stacktrace_json": stacktraceJson
```
williazz: Can we just keep the [exception.stacktrace](https://opentelemetry.io/docs/specs/semconv/registry/attributes/exception/)` semconv? I think the different stacktrace representations are already implied, and if not then we can add MetricKit json format to the "Stacktrace representation" list

Sources/Instrumentation/MetricKit/MetricKitInstrumentation.swift
```diff
@@ -0,0 +1,608 @@
...
+                    estimatedValue * count
+                }
+            sampleCount += count
+        }
+        return estimatedSum.map { $0 / sampleCount }
+    }
+
+    @available(iOS 13.0, macOS 12.0, macCatalyst 13.1, visionOS 1.0, *)
+    public func reportMetrics(payload: MXMetricPayload) {
+        let span = getMetricKitTracer().spanBuilder(spanName: "MXMetricPayload")
```
williazz: I think there was some debate about capturing `MXMetric` as a metric, span, or log. Could you document your decision to record this as a span? 

With the way its setup now, seems that we can also extend this to support metrics.

Sources/Instrumentation/MetricKit/MetricKitInstrumentation.swift
```diff
@@ -0,0 +1,608 @@
...
+            }
+            if let terminationReason = $0.terminationReason {
+                namespacedAttrs["exception.termination_reason"] = terminationReason
+            }
+            let callStackTree = $0.callStackTree
+            let stacktraceJson = String(
+                decoding: callStackTree.jsonRepresentation(),
+                as: UTF8.self
+            )
+            namespacedAttrs["exception.stacktrace_json"] = stacktraceJson
```
williazz: Same question here about `stacktrace_json`

Sources/Instrumentation/MetricKit/MetricKitInstrumentation.swift
```diff
@@ -0,0 +1,608 @@
...
+                "total_sampled_time": $0.totalSampledTime,
+            ], [:])
+        }
+        logForEach(payload.crashDiagnostics, "crash") {
+            var namespacedAttrs: [String: AttributeValueConvertable] = [:]
+            var globalAttrs: [String: AttributeValueConvertable] = [:]
+
+            // Standard OTel exception attributes - will be populated below
+            var otelType: String?
+            var otelMessage: String?
```
williazz: How good are `signalDescription` and `machExceptionDescription` at describing crashes, and allowing them to be grouped together by common root cause?

beekhc: It varies, but probably not sufficient. It just maps the human-readable names for the signals from Apple's documentation. For example, `SIGSEGV` will have the description `"segmentation violation"`.

1. I looked into what KSCrash does with its "Crash Doctor" code to make better message names. I don't think it's feasible to adapt that code for this PR. There's a lot of complexity to it, and it depends heavily on the fields KSCrash populates. I think it would be a reasonable future change if we can get it working, though.
2. Even with KSCrash's better descriptions, I doubt it would be sufficient for crash deduplication. I dunno.

My recommendation would be for us to go with this, and see if we can improve the descriptions in the future, based on usage.

The Crash Reporter instrumentation you're working on may make users less reliant on this particular diagnostic anyway.

PR #982, opened 2025-11-21
fix: fix URLSessionInstrumentation for a missing case

beekhc:
This fixes a particular way of using `NSURLSession` that the instrumentation was missing, where a session delegate implements only `func urlSession(_:task:didFinishCollecting:)`. This was caught by some Honeycomb end-to-end tests that exercise URLSession in various ways. I suspect that this was missed because this method is in the super-protocol `NSURLSessionTaskDelegate`, rather than the `NSURLSessionDataDelegate` protocol.

I have verified that our own end-to-end tests fail before this change and pass after.

I've added a regression test to make sure this stays working.

### DIFF
--- a/.clomonitor.yml
+++ b/.clomonitor.yml
@@ -1,0 +1,4 @@
+# see https://github.com/cncf/clomonitor/blob/main/docs/checks.md#exemptions
+exemptions:
+  - check: artifacthub_badge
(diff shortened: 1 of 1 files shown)


beekhc:
Moved to PR #998

PR #983, opened 2025-11-22
maint: update URLSession instrumentation semconv

beekhc:
Updates the URLSession instrumentation to have the option to use the latest semconv.

This follows the OTel HTTP [Migration Guide](https://opentelemetry.io/docs/specs/semconv/non-normative/http-migration/) to add options for which version to use, which is also how the JS SDK is handling this.

By default, it keeps the existing attributes, so that this won't be a breaking change.


beekhc:
I guess I don't know the correct way to rebase for a fork of a repo. I will have to look into that.

Anyway, created a new PR #999

PR #988, opened 2025-11-26
feat: upgrade to swift 6.0

williazz:
Upgrade minimum to swift v6.0. 

1. I went ahead and increased our test coverage to eighty percent and added test safety to agents.md. 
2. From there I started resolving the Swift build and Swift Test issues. Then took care of all the platforms individually. Linux was the most flaky. 
3. Last, I also went ahead and took care of actual concurrency warnings. (2) only disabled warnings and allowed building with (2). 

All the unit tests pass and platform tests pass, but we need to sit with this for a bit and test it on real devices. I have four devices on hand that I can test with, and I would appreciate other people doing the same. I don't think emulator's gonna be enough.


williazz:
Finally had time to work on this. Got the build working. Now I need to rewrite this so that existing behaviors don't get weakened or eroded.

williazz:
Okay, this is ready for review. I'm going to continue testing it on real devices and sitting with it for a bit. I think we should also add some platform version testing as well. So I can look into that.

cc @bryce-b

williazz:
Verified on emulator.

Screenshot 2026-05-04 at 3 46 39 PM

Sources/Instrumentation/URLSession/URLSessionLogger.swift
```diff
@@ -43,61 +43,65 @@ class URLSessionLogger {
...
     let useStable = instrumentation.configuration.semanticConvention == .stable || instrumentation.configuration.semanticConvention == .httpDup
 
+    // Legacy (OTel v0) HTTP attribute keys. The semantic itself is deprecated
+    // by OpenTelemetry; we still emit them when callers opt into .old/.httpDup
+    // for backward compatibility. Using literal keys avoids the deprecation
+    // warning on each SemanticAttributes.* rawValue.
     let method = request.httpMethod ?? "unknown_method"
     if useOld {
-      attributes[SemanticAttributes.httpMethod.rawValue] = AttributeValue.string(method)
+      attributes["http.method"] = AttributeValue.string(method)
```
nachoBonafonte: why is not possible to use the constant, maybe we need to change how they are generated instead

nachoBonafonte (commented):
It looks very good, just a small nit from me

Sources/Instrumentation/URLSession/URLSessionLogger.swift
williazz: thanks nacho. those were for legacy attributes that are no longer in semconv. i created a separate object for it in `LegacyHTTPAttributes` and would like to move off of this in the future

nachoBonafonte (approved):

bryce-b (approved):
LGTM, thanks for adding an AGENTS.md

williazz:
Thanks team

PR #994, opened 2025-12-02
chore: add badges to readme

thompson-tomo:
This adds the ossf scorecard badge to the readme to improve the clo monitor score and make it easily accessible to those interested.

Can a maintainer register this repo on https://www.bestpractices.dev/en/projects?q=Opentelemetry & we can add that badge as part of this PR.

Also added is the fossa badges which also improves the clo monitor score.


bryce-b:
for future use : https://www.bestpractices.dev/en/projects/11537

PR #998, opened 2025-12-02
fix: fix URLSessionInstrumentation for a missing case

beekhc:
This fixes a particular way of using `NSURLSession` that the instrumentation was missing, where a session delegate implements only `func urlSession(_:task:didFinishCollecting:)`. This was caught by some Honeycomb end-to-end tests that exercise URLSession in various ways. I suspect that this was missed because this method is in the super-protocol `NSURLSessionTaskDelegate`, rather than the `NSURLSessionDataDelegate` protocol.

I have verified that our own end-to-end tests fail before this change and pass after.

I've added a regression test to make sure this stays working.

### DIFF
--- a/.github/workflows/BuildAndTest.yml
+++ b/.github/workflows/BuildAndTest.yml
@@ -14,7 +14,7 @@ jobs:
     outputs:
       should-run: ${{ steps.check.outputs.should-run }}
     steps:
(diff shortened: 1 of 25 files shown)


bryce-b (approved):

bryce-b:
@beekhc I believe codecov is mad because this branch is 257 commits behind main, which isn't a problem for merge; but the tests from that time hand less coverage. I believe this can be ignored for merge.

PR #1000, opened 2025-12-03
refactor!: record sessions as log events

williazz:
## Breaking

I refactoring session events to be recorded as log event, as opposed to regular log records. In addition, I am removing `start_time` and `end_time` from attributes, since they are not needed (discussed in https://github.com/open-telemetry/semantic-conventions/pull/2770). 

## Summary

Since they are session "events", I have refactored session to be log events. https://opentelemetry.io/docs/specs/semconv/general/session/

I have also improved the thread safety in session instrumentation by migrating from locks to queues, which I found to be much less susceptible to lock contention. This will also help with the migration in https://github.com/open-telemetry/opentelemetry-swift/pull/988

## Logs

Session Start

```json
{
              "timeUnixNano": "1764721305052582144",
              "attributes": [
                {
                  "key": "session.id",
                  "value": { "stringValue": "2840B2AE-E4ED-41E0-B09E-3300DD715F46" }
                },
                {
                  "key": "session.previous_id",
                  "value": { "stringValue": "C2A8C842-36B3-4CBB-90FE-5FDC0E6A4CF6" }
                }
              ],
              "observedTimeUnixNano": "1764721305052721920",
              "eventName": "session.start"
            }
```

Session End

```json
{
              "timeUnixNano": "1764721293493801984",
              "attributes": [
                {
                  "key": "session.id",
                  "value": { "stringValue": "C2A8C842-36B3-4CBB-90FE-5FDC0E6A4CF6" }
                },
                {
                  "key": "session.previous_id",
                  "value": { "stringValue": "3D787EDB-E3F1-4E8E-8216-912B5DDEAE9D" }
                },
                { "key": "session.duration", "value": { "doubleValue": 127912 } }
              ],
              "observedTimeUnixNano": "1764721305052654080",
              "eventName": "session.end"
            }
```

## Testing
Refactored the aws-otel-swift and AwsHackerNewsApp to use otel session instrumentation here - https://github.com/aws-observability/aws-otel-swift/compare/main...williazz:aws-otel-swift:dev/session-events-refactor

Visualization in local AwsOtelUI framework

Screenshot 2025-12-02 at 4 22 35 PM

### DIFF
--- a/Sources/Instrumentation/Sessions/SessionConstants.swift
+++ b/Sources/Instrumentation/Sessions/SessionConstants.swift
@@ -19,19 +19,6 @@ public class SessionConstants {
   public static let sessionStartEvent = "session.start"
   /// Event name for session end events
   public static let sessionEndEvent = "session.end"
-  /// Attribute name for session identifier
-  @available(*, deprecated, message: "Use SemanticConventions.Session.id instead")
-  public static let id = "session.id"
-  /// Attribute name for previous session identifier
-  @available(*, deprecated, message: "Use SemanticConventions.Session.previousId instead")
-  public static let previousId = "session.previous_id"
-
-  // MARK: - Extension Attributes
-  
-  /// Attribute name for session duration
-  public static let duration = "session.duration"
-
-  // MARK: - Internal Constants
   
   /// Notification name for session events
   public static let sessionEventNotification = "SessionEventInstrumentation.SessionEvent"
--- a/Sources/Instrumentation/Sessions/SessionEventInstrumentation.swift
+++ b/Sources/Instrumentation/Sessions/SessionEventInstrumentation.swift
@@ -130,14 +130,12 @@ public class SessionEventInstrumentation {
   /// end time, duration, and previous session ID (if available).
   /// - Parameter session: The expired session
   private static func createSessionEndEvent(session: Session) {
-    guard let endTime = session.endTime,
-    let duration = session.duration else {
+    guard let endTime = session.endTime else {
       return
     }
 
     var attributes: [String: AttributeValue] = [
-      SemanticConventions.Session.id.rawValue: AttributeValue.string(session.id),
-      SessionConstants.duration: AttributeValue.double(Double(duration.toNanoseconds))
+      SemanticConventions.Session.id.rawValue: AttributeValue.string(session.id)
     ]
 
     if let previousId = session.previousId {
--- a/Sources/Instrumentation/Sessions/README.md
+++ b/Sources/Instrumentation/Sessions/README.md
@@ -123,7 +123,6 @@ let session = Session(
 )
 
 print("Expired: \(session.isExpired())")
-print("Duration: \(session.duration ?? 0)")
 ```
 
 ## Configuration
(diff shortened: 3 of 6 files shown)


Sources/Instrumentation/Sessions/SessionConstants.swift
```diff
@@ -9,12 +9,15 @@
...
 public class SessionConstants {
   // MARK: - OpenTelemetry Semantic Conventions
   
   /// Event name for session start events
   public static let sessionStartEvent = "session.start"
-  /// Event name for session end events  
+  /// Event name for session end events
   public static let sessionEndEvent = "session.end"
   /// Attribute name for session identifier
   public static let id = "session.id"
```
beekhc: Should we use the [Semantic Conventions constants](https://github.com/open-telemetry/opentelemetry-swift-core/blob/aed2b0428a3a78d287817919e1611713e72d1177/Sources/OpenTelemetryApi/Common/SemanticAttributes/Attributes/Session_attributes.swift) for these?

Sources/Instrumentation/Sessions/SessionEventInstrumentation.swift
```diff
@@ -134,16 +129,14 @@ public class SessionEventInstrumentation {
...
-          let duration = session.duration else {
+    let duration = session.duration else {
       return
     }
 
     var attributes: [String: AttributeValue] = [
       SessionConstants.id: AttributeValue.string(session.id),
-      SessionConstants.startTime: AttributeValue.double(Double(session.startTime.timeIntervalSince1970.toNanoseconds)),
-      SessionConstants.endTime: AttributeValue.double(Double(endTime.timeIntervalSince1970.toNanoseconds)),
       SessionConstants.duration: AttributeValue.double(Double(duration.toNanoseconds))
```
beekhc: I'm surprised to see nanoseconds used as the unit here. The only semantic convention for durations I can find is for [http](https://opentelemetry.io/blog/2023/http-conventions-declared-stable/), where seconds are used for all units. If we're already making a breaking change to telemetry shape, it might be worth standardizing the units?

Or maybe we should remove this entirely, like you're doing with the other time attributes? None of them are required by the currently proposed [session semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/session/).

Sources/Instrumentation/Sessions/SessionManager.swift
```diff
@@ -38,22 +30,38 @@ public class SessionManager {
   /// - Returns: The current active session
   @discardableResult
   public func getSession() -> Session {
-    refreshSession()
+    let transition : SessionTransition? = lock.withLock {
+      if session == nil || session!.isExpired() {
+        return startSession()
```
naftaly: calling out here can be really dangerous since you hold the lock. Sometimes you can do it when you have a naming convention like `locked_startSession` so everyone knows that you're already lock in there. The biggest fear is that `startSession` calls out in some fashion. If this isn't the case then ignore.

Sources/Instrumentation/Sessions/SessionConstants.swift
williazz: i couldn't find the event names though, maybe we are missing some structs?

bryce-b: ~It should be accessible here: https://github.com/open-telemetry/opentelemetry-swift-core/blob/main/Sources/OpenTelemetryApi/Common/SemanticAttributes/Attributes/Session_attributes.swift#L23~

misunderstood the discussion; we may need to re-generate the semantic attributes with addition flags to include experimental items. I can follow up on this.

bryce-b (approved):

PR #1083, opened 2026-04-13
crash fix 

fangandyuan:


nachoBonafonte:
Please provide a description attaching the crashed stack and also the comments in the code should be in English.

bryce-b:
Fixed in #1086

Issue #636, opened 2024-11-13
NetworkStatus crash NetworkStatus.NetworkStatus.status() -> (Swift.String, Swift.String?, __C.CTCarrier?) SIGTRAP


xernaj:
Hi All,

Great work on the library so far.

We've had a bug report come in re: NetworkStatus instrumentation. NetworkStatus.NetworkStatus.status() -> (Swift.String, Swift.String?, __C.CTCarrier?) (NetworkStatus.swift:29)

Has anyone seen this before? Report excerpt below:
```Code Type:       arm64
Parent Process:   [1]

Date/Time:       2024-11-13T13:13:35.999Z
Launch Time:     2024-11-13T12:57:36Z
OS Version:      iPhone OS 17.4.1 (21E236)
Report Version:  104

Exception Type:  SIGTRAP
Exception Codes: TRAP_BRKPT at 0x19d5c506c
Crashed Thread:  28

Thread 28 Crashed:
0   CoreFoundation                       0x000000019d5c506c CF_IS_OBJC + 76
1   CoreFoundation                       0x000000019d5fd57c __CFStringIsCF + 44
2   Foundation                           0x000000019c464ad0 static (extension in Foundation):Swift.String._unconditionallyBridgeFromObjectiveC(__C.NSString?) -> Swift.String + 196
3   <appnameredacted>                    0x0000000104e20d74 NetworkStatus.NetworkStatus.status() -> (Swift.String, Swift.String?, __C.CTCarrier?) (NetworkStatus.swift:29)
4   <appnameredacted>                    0x0000000104e218a0 NetworkStatus.NetworkStatusInjector.inject(span: OpenTelemetryApi.Span) -> () (NetworkStatusInjector.swift:21)
5   <appnameredacted>                    0x0000000104e1c7a4 static URLSessionInstrumentation.URLSessionLogger.processAndLogRequest(_: Foundation.URLRequest, sessionTaskId: Swift.String, instrumentation: URLSessionInstrumentation.URLSessionInstrumentation, shouldInjectHeaders: Swift.Bool) -> Foundation.URLRequest? (URLSessionLogger.swift:92)
[...]

PR #1086, opened 2026-04-15
#636 Fix thread-safety crash in NetworkStatus.status()

0xLucasMarcal:
## Summary

`NetworkStatus.status()` reads `CTTelephonyNetworkInfo` properties directly on the caller's thread, which causes `EXC_BAD_ACCESS` crashes when CoreTelephony's internal state is mutated concurrently. This change caches the telephony state behind an `NSLock` and keeps it up to date via `CTTelephonyNetworkInfoDelegate`, so callers of `status()` never touch the thread-unsafe ObjC properties directly.

## Problem

`CTTelephonyNetworkInfo` is not thread-safe. Its Objective-C properties (`dataServiceIdentifier`, `serviceCurrentRadioAccessTechnology`, `serviceSubscriberCellularProviders`) can be freed internally by CoreTelephony on its own thread while being read from another thread. When the returned ObjC object is autoreleased after its backing memory has already been freed, the app crashes with `EXC_BAD_ACCESS` in `objc_autorelease`.

This is triggered in production because `URLSessionLogger.processAndLogRequest` calls `netstatInjector.inject(span:)` -- and therefore `NetworkStatus.status()` -- from arbitrary background threads (GCD worker threads via `NSOperationQueue`).

Example crash stack trace:

```
Thread Name: EXC_BAD_ACCESS
  libobjc.A.dylib        objc_autorelease
  CoreTelephony           -[CTTelephonyNetworkInfo dataServiceIdentifier]
  App                     NetworkStatus.status()            (NetworkStatus.swift:29)
  App                     NetworkStatusInjector.inject(span:) (NetworkStatusInjector.swift:21)
  App                     URLSessionLogger.processAndLogRequest(...)  (URLSessionLogger.swift:92)
```

Previously reported as #636 (closed without a fix).

## Solution

Instead of reading `CTTelephonyNetworkInfo` properties on demand from arbitrary threads, we:

1. **Cache the telephony state** (`dataServiceIdentifier`, `serviceCurrentRadioAccessTechnology`, `serviceSubscriberCellularProviders`) as plain Swift values at init time.

2. **Keep the cache up to date** by installing a `CTTelephonyNetworkInfoDelegate` (via a lightweight `NetworkInfoDelegateProxy`). The delegate callbacks fire on CoreTelephony's internal serial queue *after* the property mutation has completed, so reading the properties in the callback is safe.

3. **Serve only cached values** from `status()`, protected by an `NSLock`. Callers never touch the underlying `CTTelephonyNetworkInfo` properties directly, eliminating the race condition entirely.

This follows the same `NSLock`-based thread-safety pattern already used by `NetworkMonitor` in this codebase.

### Why a delegate proxy?

`CTTelephonyNetworkInfoDelegate` requires `NSObject` conformance. Making `NetworkStatus` inherit from `NSObject` would break its existing `convenience init() throws` (Swift does not allow overriding a non-throwing superclass init with a throwing one). The `NetworkInfoDelegateProxy` is a small private `NSObject` subclass that bridges the delegate callbacks to a closure, keeping `NetworkStatus` as a plain Swift class with its original public API intact.

## Changes

- **`NetworkStatus.swift`**: Add `NSLock`, cached properties, `refreshCachedState()`, `NetworkInfoDelegateProxy`. `status()` reads from cache instead of `CTTelephonyNetworkInfo` directly.
- **`NetworkStatusTests.swift`**: Add `testConcurrentStatusAccess` -- calls `status()` from 10,000 concurrent threads to verify no crashes or data races.

## Backward compatibility

- No changes to public API signatures.
- No changes to `URLSessionLogger.swift` or `NetworkStatusInjector.swift` (fix is entirely within `NetworkStatus`).
- The pre-iOS 13 fallback path in `status()` is preserved.

### DIFF
--- a/Sources/Instrumentation/NetworkStatus/NetworkStatus.swift
+++ b/Sources/Instrumentation/NetworkStatus/NetworkStatus.swift
@@ -8,9 +8,38 @@
   import Foundation
   import Network
 
+  /// Provides network connection type and carrier information for span attributes.
+  ///
+  /// ## Thread safety
+  ///
+  /// `CTTelephonyNetworkInfo` is not thread-safe: its Objective-C properties
+  /// can be mutated internally by CoreTelephony on its own thread while being
+  /// read from a caller's thread, causing `EXC_BAD_ACCESS` when the returned
+  /// object is autoreleased after its backing memory has already been freed.
+  ///
+  /// To avoid this, we eagerly cache the telephony state and keep it up to date
+  /// through `CTTelephonyNetworkInfoDelegate` callbacks (via a proxy object),
+  /// which fire *after* the internal mutation has completed. Callers of
+  /// `status()` read only the cached Swift values, protected by an `NSLock`,
+  /// and never touch the underlying `CTTelephonyNetworkInfo` properties
+  /// directly.
   public class NetworkStatus {
     public private(set) var networkInfo: CTTelephonyNetworkInfo
     public private(set) var networkMonitor: NetworkMonitorProtocol
+
+    // MARK: - Cached telephony state
+
+    /// Guards all reads and writes to the `cached*` properties below.
+    private let lock = NSLock()
+
+    private var cachedServiceId: String?
+    private var cachedRadioTech: [String: String]?
+    private var cachedProviders: [String: CTCarrier]?
+
+    /// Delegate proxy that forwards `CTTelephonyNetworkInfoDelegate`
+    /// callbacks to `refreshCachedState()`. Stored to prevent deallocation.
+    private var delegateProxy: AnyObject?
+
     public convenience init() throws {
       try self.init(with: NetworkMonitor())
     }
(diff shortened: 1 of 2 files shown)


nachoBonafonte:
Thanks for the PR, the documentation is not following the style of the rest of the project. We should document what the code does and if necessary because it is very complex "how", we should not document what the new code is fixing, e.g. the thread safety comment before the  Network status is not necessary, that documentation is part of the PR or the commit, not the code itself.

0xLucasMarcal:
> Thanks for the PR, the documentation is not following the style of the rest of the project. We should document what the code does and if necessary because it is very complex "how", we should not document what the new code is fixing, e.g. the thread safety comment before the Network status is not necessary, that documentation is part of the PR or the commit, not the code itself.

@nachoBonafonte done!

0xLucasMarcal:
ping @ArielDemarco @nachoBonafonte

bryce-b (approved):

PR #1063, opened 2026-03-11
Swift Distributed Tracing Bridge

simonbility:
👋 Hi!

We have a relatively large iOS codebase that has been using swift-distributed-tracing (using swift-otel as the "backend")

We now have some more sophisticated needs around exporting traces that are handled very nicely by this project :)  (specifically Auth and the PersistenceExporter)

Since our codebase is already based on swift-distributed-tracing our preferred way of addressing this would be to only swap out the "backend".

So i made an attempt to make opentelemetry-swift expose a (swift-distributed-tracing) `Tracer` 

I used the existing `OTelSwiftLog` as a reference

I checked and there seems to be some interest in this:
* https://github.com/open-telemetry/opentelemetry-swift/issues/516
* https://github.com/open-telemetry/opentelemetry-swift/issues/201
* https://github.com/open-telemetry/opentelemetry-swift/issues/563

Please let me know if there is interest in pursuing this further.

PR #1090, opened 2026-04-25
Update OTelLogHandler to use error parameter from swift-log

### LINKED ISSUE #1089: Support recent updates from swift-log
Recent versions of `swift-log` have changed the API for a `LogHandler`.
* [1.11.0](https://github.com/apple/swift-log/releases/tag/1.11.0) added a new `LogHandler.log()` method that takes a `LogEvent` parameter, while marking the previous methods as deprecated
* [1.12.0](https://github.com/apple/swift-log/releases/tag/1.12.0) added new log methods that allows the caller to attach an `Error`

The `OTelLogHandler` should be updated to reflect these changes, both to remove deprecation warnings, and to not drop/ignore any `Error` that is included in a log event from a user.

samuelmurray:
Support error parameter from `swift-log`

### Motivation:

Version 1.12.0 of `swift-log` added support for attach an `Error` to a log event (https://github.com/apple/swift-log/releases/tag/1.12.0). The `OTelSwiftLog` target should use this parameter (if present) and serialize it as OTel data. The best matching OTel attribute is `exception` (https://opentelemetry.io/docs/specs/semconv/registry/attributes/exception/), in which we can populate `message` and `type`. 
Closes #1089 

### Modifications:

* Update `swift-log` dependency to `1.12.0`
* Use new `log(event:)` method in `LogHandler`
* Serialize `error` parameter from logs as `exception.message` and `exception.type`

### Result:

If users of `OTelSwiftLog` attach an error to logs (e.g. `logger.warn("Something went wrong", error: error)`), the error will be serialized and its information included in the final OTel data.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -33,7 +33,7 @@ let package = Package(
     .package(url: "https://github.com/apple/swift-nio.git", from: "2.94.0"),
     .package(url: "https://github.com/grpc/grpc-swift.git", exact: "1.27.1"),
     .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.33.3"),
-    .package(url: "https://github.com/apple/swift-log.git", from: "1.10.1"),
+    .package(url: "https://github.com/apple/swift-log.git", from: "1.12.0"),
     .package(url: "https://github.com/apple/swift-metrics.git", from: "2.8.0")
   ],
   targets: [
--- a/Sources/Bridges/OTelSwiftLog/LogHandler.swift
+++ b/Sources/Bridges/OTelSwiftLog/LogHandler.swift
@@ -48,23 +48,17 @@ public struct OTelLogHandler: LogHandler {
       .build()
   }
 
-  public func log(level: Logging.Logger.Level,
-                  message: Logging.Logger.Message,
-                  metadata: Logging.Logger.Metadata?,
-                  source: String,
-                  file: String,
-                  function: String,
-                  line: UInt) {
+  public func log(event: LogEvent) {
     // This converts log atrributes to otel attributes
     var otelattributes: [String: AttributeValue] = [
-      "source": AttributeValue.string(source),
-      "file": AttributeValue.string(file),
-      "function": AttributeValue.string(function),
-      "line": AttributeValue.int(Int(line))
+      "source": AttributeValue.string(event.source),
+      "file": AttributeValue.string(event.file),
+      "function": AttributeValue.string(event.function),
+      "line": AttributeValue.int(Int(event.line))
     ]
 
     // Convert metadata from the method parameter to AttributeValue and assign it to otelattributes
-    if let metadata {
+    if let metadata = event.metadata {
       let methodMetadata = convertMetadata(metadata)
       otelattributes.merge(methodMetadata) { _, new in new }
     }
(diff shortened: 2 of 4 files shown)


samuelmurray:
I'm creating this as a draft-PR as I have some questions regarding the error serialization. For spans, with `recordException`, you rely on the bridging from `Error` to `NSError` (in the [API](https://github.com/open-telemetry/opentelemetry-swift-core/blob/main/Sources/OpenTelemetryApi/Trace/SpanException.swift#L15-L27) and the [SDK](https://github.com/open-telemetry/opentelemetry-swift-core/blob/main/Sources/OpenTelemetrySdk/Trace/SpanSdk.swift#L360-L375)), which also looks to be used in the [upcoming bridge](https://github.com/open-telemetry/opentelemetry-swift/pull/1063/changes#diff-e221c1692709fa5ca1bc0c9603868b8c4b54801275269f2954c13e24322a6b5eR218-R228) to `swift-distributed-tracing`. 

However, I think this gives some undesirable results, as neither "exception.message" nor "exception.type" give reasonable output, as shown in the tests. From the [OTel documentation](https://opentelemetry.io/docs/specs/semconv/registry/attributes/exception/), `exception.type` should be

> The type of the exception (its fully-qualified class name, if applicable). 

I don't think `NSError.code` can be said to yield this, especially not when bridging from `Error`. In `swift-otel`, we look to use `String(reflecting: type(of: error))` for the `type` and the string representation (`"\(error)"`) for `message` (https://github.com/swift-otel/swift-otel/pull/417). I would suggest using the same in this project - but also change the span implementation to use that, if nothing else in the bridge to `swift-distributed-tracing`. Though, as noted in the PR in `swift-otel`, span events are being deprecated (https://opentelemetry.io/blog/2026/deprecating-span-events/), so perhaps it's ok to leave that implementation untouched and use a more sensible one for logs.

bryce-b (approved):

bryce-b:
> I don't think NSError.code can be said to yield this, especially not when bridging from Error. In swift-otel, we look to use String(reflecting: type(of: error)) for the type and the string representation ("\(error)") for message (https://github.com/swift-otel/swift-otel/pull/417). I would suggest using the same in this project - but also change the span implementation to use that, if nothing else in the bridge to swift-distributed-tracing. Though, as noted in the PR in swift-otel, span events are being deprecated (https://opentelemetry.io/blog/2026/deprecating-span-events/), so perhaps it's ok to leave that implementation untouched and use a more sensible one for logs.

I agree with this. Go ahead and add that change. Let's not worry about the difference with the span implementation since we are actively removing the span-events.

bryce-b:
@samuelmurray we are waiting for the conflicts to be resolved and then we can merge this pr.

samuelmurray:
Sorry for not getting back to you, I've had a lot going on recently. It will be a couple of weeks before I can get it done; if someone else wants to finish the work before that, I'm of course ok with that!

samuelmurray:
> @samuelmurray we are waiting for the conflicts to be resolved and then we can merge this pr.

Updated the PR now, marking it as non-draft.

samuelmurray:
@bryce-b Oh, I hadn't noticed that you adjusted for the `event`-method in another branch, so actually this PR only adds the error parameter. Fixed the conflicts now, and moved the test to use the same structure as your other attributes-tests.

bryce-b:
@samuelmurray this needs another rebase, sorry about that.

samuelmurray:
@bryce-b should be fixed now

vvydier (approved):

PR #1097, opened 2026-05-13
fix(URLSessionInstrumentation): dedupe spans on watchOS data(for:) path

ypopovych:
URLSession instrumentation creates more than one span per request on watchOS.  URLSession on watchOS does public API calls instead of internal ones, so current instrumentation called multiple times for the same request and creates orphaned(normal requests) and finished(requests with redirects) spans.

`URLSession.data(for:)` on watchOS spawns an inner `__NSCFLocalDataTask` via the public `session.dataTask(with:URLRequest)` factory after our resume swizzle has already injected a `traceparent` on the outer task's `currentRequest`. That re-entry would start a second span and orphan the first (started but never .end()'d).

Added check to skip the duplicate by matching the request's traceparent against the running-spans map.

Also added a defensive early-return in `urlSessionTaskWillResume` when a span for the task id already exists — this also covers the `URLSessionTask.resume` / `__NSCFURLSessionTask.resume` super-class chain and redirect-triggered re-resumes.

Tested fix on all supported platforms. All `URLSessionInstrumentationTests` pass.

### DIFF
--- a/Sources/Instrumentation/URLSession/InstrumentationUtils.swift
+++ b/Sources/Instrumentation/URLSession/InstrumentationUtils.swift
@@ -8,7 +8,7 @@ import Foundation
   import os.log
 #endif
 
-#if os(iOS) || os(tvOS)
+#if os(iOS) || os(tvOS) || os(visionOS) 
 import UIKit
 #elseif os(watchOS)
 import WatchKit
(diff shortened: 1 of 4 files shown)


nachoBonafonte (commented):
I am concerned that this approach can bring extra

nachoBonafonte:
I am concerned that this approach can bring extra operations to son platforms, specially with the queue sync. If it is only happening in watchOS, wouldn't it be better to run only conditionally, I am concerned on penalizing other platforms without this difference.

ypopovych:
@nachoBonafonte I will check but seems as there are two bugs, and one of them can be triggered on all platforms and one is watchOS unique. Will do a more debugging and will add ifdefs for the watchOS unique.

ypopovych:
@nachoBonafonte seems as both can be safely guarded as watchOS only. Updated.

nachoBonafonte (approved):

PR #1107, opened 2026-05-28
Add configurable session lifecycle rules

robert-northmind:
## Summary

I think there might be use cases when you want to control the session lifecycle a bit more, and for example be able to tell it to always create a new session on clean app start.

I am not sure if this is the right approach to just open a PR, or if you would like to have an issue for this first.
Maybe you already discussed this topic and decided against it. 

Anyhow, it was not so much work to add this support, so I thought I would just open up a PR directly.

This PR does:

- Add `maxLifetime` and `restorePersistedSession` to `SessionConfig` and `SessionConfigBuilder` while preserving existing defaults.
- Expire sessions by either inactivity timeout or optional maximum lifetime.
- Support clean-start session behavior with `restorePersistedSession: false`, while still linking the persisted session as `session.previous_id` and emitting an inferred `session.end` before the new `session.start`.
- Persist `maxLifetime`, handle old persisted sessions where it is absent, and update Sessions README docs.

## Context

The OpenTelemetry session semantic convention defines the session attributes and lifecycle events, but leaves lifecycle policy to instrumentation. It says sessions may end due to inactivity or timeout, and allows a new `session.start` to include `session.previous_id` when continuing from a prior session: https://opentelemetry.io/docs/specs/semconv/general/session/

The Android OpenTelemetry instrumentation exposes a session lifecycle configuration for inactivity timeout and maximum lifetime (`backgroundInactivityTimeout`, `maxLifetime`): https://opentelemetry.io/docs/platforms/client-apps/android/
This change brings Swift closer to that configurable session-SDK behavior without changing default behavior for existing users.

## Compatibility

Defaults preserve current behavior:

- `sessionTimeout` remains 30 minutes.
- `maxLifetime` defaults to `nil`, so no absolute lifetime cap is applied unless configured.
- `restorePersistedSession` defaults to `true`, so persisted sessions are resumed as current by default.

Persisted sessions created by older versions do not contain `maxLifetime`; those load as `nil`.

## Validation

- `swift build --build-tests`
- `swift test --filter SessionTests`

### DIFF
--- a/Sources/Instrumentation/Sessions/README.md
+++ b/Sources/Instrumentation/Sessions/README.md
@@ -39,8 +39,10 @@ let builder = LoggerProviderBuilder()
 **Custom Configuration**:
 
 ```swift
-let sessionConfig = SessionConfigBuilder()
+let sessionConfig = SessionConfig.builder()
     .with(sessionTimeout: 45 * 60) // 45 minutes
+    .with(maxLifetime: 4 * 60 * 60)
+    .with(restorePersistedSession: false)
     .build()
 let sessionManager = SessionManager(configuration: sessionConfig)
 SessionManagerProvider.register(sessionManager: sessionManager)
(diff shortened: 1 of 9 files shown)


robert-northmind:
I looked into the `Build and Test` failure. The failing jobs appear unrelated to this PR’s session lifecycle changes.

This PR only changes `Sources/Instrumentation/Sessions`, the Sessions README, and Sessions tests. The failed jobs are failing while compiling `Sources/Exporters/Persistence/PersistenceLogExporterDecorator.swift`.

The run resolved `opentelemetry-swift-core` to `2.5.0`, which was released after this PR was opened. With that version, `LogRecordExporter` is `Sendable`, and the existing `PersistenceLogExporterDecorator` implementation fails Swift concurrency checks:

```text
non-final class 'PersistenceLogExporterDecorator' cannot conform to the 'Sendable' protocol
stored property 'persistenceExporter' ... has non-Sendable type
```

I was able to reproduce this locally: swift test passes with opentelemetry-swift-core 2.4.1, but fails before tests run after resolving core 2.5.0.
So I think this is a base/dependency compatibility issue rather than a regression introduced by this PR.

williazz:
Hi @robert-northmind , can we try to get to the bottom of this approval wflow failure? Even if it's unrelated to your changes

williazz:
Nice changes though. Both maxLifeTime and restorePersistedSession are super useful

robert-northmind:
> Hi @robert-northmind , can we try to get to the bottom of this approval wflow failure? Even if it's unrelated to your changes

Hi @williazz , thanks for the feedback. and sure thing. I'm out on vacation for 1 more week. But when I am back home I can look into this.

robert-northmind:
> Hi @robert-northmind , can we try to get to the bottom of this approval wflow failure? Even if it's unrelated to your changes

Hey again, so I noticed that there are actually 2 open PRs which solve these ci pipeline errors.

Cause: CI now resolves opentelemetry-swift-core to 2.5.0, where LogRecordExporter became Sendable. Existing non-final conformers (e.g. PersistenceLogExporterDecorator, the gRPC OtlpLogExporter) no longer compile. This hits main and every open PR, not just this one.

Here are the PRs which address this:
	•	#1120 – Fix Concurrency Issues
	•	its core dependency opentelemetry-swift-core#87

Once those merge and a new core release is out, CI here should go green again with no further changes. I’ll rebase then. The session lifecycle changes are unaffected and ready for review.

bryce-b (approved):

PR #1115, opened 2026-06-09
URLSessionInstrumentation - swizzle to always add URLSessionDataDelegate hooks

simonbility:
I encountered some IMO surprising behaviour of `URLSessionInstrumentation`

It swizzles the delegate classes to install its own hooks.
But this as far as i can tell only works when these conditions are met:

In order for the swizzling to kick in at all i must implement one of these.

```
    let selectors = [
      #selector(URLSessionDataDelegate.urlSession(_:dataTask:didReceive:)),
      #selector(
        URLSessionDataDelegate.urlSession(
          _:dataTask:didReceive:completionHandler:)),
      #selector(
        URLSessionDataDelegate.urlSession(_:task:didCompleteWithError:)),
      #selector(
        URLSessionDataDelegate.urlSession(_:dataTask:didBecome:)
          as (URLSessionDataDelegate) -> (
            (URLSession, URLSessionDataTask, URLSessionDownloadTask) -> Void
          )?),
      #selector(
        URLSessionDataDelegate.urlSession(_:dataTask:didBecome:)
          as (URLSessionDataDelegate) -> (
            (URLSession, URLSessionDataTask, URLSessionStreamTask) -> Void
          )?),
      #selector(
        URLSessionTaskDelegate.urlSession(_:task:didFinishCollecting:))
    ]
```

Additionally if i dont implement one of them the hooks relying on this wont fire.

For instance `shouldRecordPayload` will only fire if the following method is implemented.

```
urlSession(
  _ session: URLSession, dataTask: URLSessionDataTask,
  didReceive response: URLResponse,
  completionHandler: @escaping (URLSession.ResponseDisposition) -> Void
)
```

IMO This behaviour is a bit unintuitive.

I adapted the implementation so it searches for implementors of URLSessionTaskDelegate rather than just classes that have a matching method.

Please let me know what you think,
* Maybe this could also be solved by explicitly documenting these conditions?
* Or maybe some explicit way to opt in to the new behaviour, to avoid any unexpected behaviour changes.

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -492,6 +492,13 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
     let selector = #selector(
       URLSessionDataDelegate.urlSession(_:dataTask:didReceive:))
     guard let original = class_getInstanceMethod(cls, selector) else {
+      let block:
+        @convention(block) (Any, URLSession, URLSessionDataTask, Data) -> Void = { _, session, dataTask, data in
+          self.urlSession(session, dataTask: dataTask, didReceive: data)
+        }
+      let imp = imp_implementationWithBlock(
+        unsafeBitCast(block, to: AnyObject.self))
+      class_addMethod(cls, selector, imp, "v@:@@@")
       return
     }
     var originalIMP: IMP?
@@ -517,6 +524,16 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
       URLSessionDataDelegate.urlSession(
         _:dataTask:didReceive:completionHandler:))
     guard let original = class_getInstanceMethod(cls, selector) else {
+      let block:
+        @convention(block) (Any, URLSession, URLSessionDataTask, URLResponse,
+                            @escaping (URLSession.ResponseDisposition) -> Void) -> Void = { _, session, dataTask, response, completion in
+          self.urlSession(session, dataTask: dataTask, didReceive: response,
+                          completionHandler: completion)
+          completion(.allow)
+        }
+      let imp = imp_implementationWithBlock(
+        unsafeBitCast(block, to: AnyObject.self))
+      class_addMethod(cls, selector, imp, "v@:@@@@")
       return
     }
     var originalIMP: IMP?
@@ -544,6 +561,13 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
     let selector = #selector(
       URLSessionDataDelegate.urlSession(_:task:didCompleteWithError:))
     guard let original = class_getInstanceMethod(cls, selector) else {
+      let block:
+        @convention(block) (Any, URLSession, URLSessionTask, Error?) -> Void = { _, session, task, error in
+          self.urlSession(session, task: task, didCompleteWithError: error)
+        }
+      let imp = imp_implementationWithBlock(
+        unsafeBitCast(block, to: AnyObject.self))
+      class_addMethod(cls, selector, imp, "v@:@@@")
       return
     }
     var originalIMP: IMP?
@@ -576,7 +600,7 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
         }
       let imp = imp_implementationWithBlock(
         unsafeBitCast(block, to: AnyObject.self))
-      class_addMethod(cls, selector, imp, "@@@")
+      class_addMethod(cls, selector, imp, "v@:@@@")
       return
     }
     var originalIMP: IMP?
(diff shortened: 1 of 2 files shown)


ArielDemarco (changes requested):
Hey @simonbility! Thanks for the contribution and for digging into this!

The `class_addMethod` additions seems that are the right fix for the problem you mention.

However, the change to the class detection mechanism introduces a regression (aka switching from selector scan to use `class_conformsToProtocol`). Many apps (and in particular, many frameworks) implement `URLSessionDelegate` methods without declaring formal protocol conformance. `URLSession` itself dispatches based on `respondsToSelector:`, not protocol conformance (that's why it handles those cases correctly). The original selector-based scan matches that behavior; `class_conformsToProtocol` does not, will silently skip those classes and will miss those requests.

What I'd do is to keep the original selector scan to find candidate classes, and apply your `class_addMethod` additions on top of it.

simonbility:
@ArielDemarco thanks for checking it out!

I kind of assumed the backward compatibility aspect would need some changes.

I will switch back to the seletcor based discovery.

But i would propose one change.

If i explicitly pass in `delegateClassesToInstrument` it should unconditionally swizzle them.

this was what originally confused me.

I created a empty delegate class and passed that in, and nothing happened. (Since it fails the selector based check)

I did this to be a bit more selective with whats instrumented, and also avoid the performance hit of scanning all classes (just assumed but havent measured it 😅 )

Do you have any concerns with this approach?

ArielDemarco:
Fair! the selector check is a heuristic to find relevant delegates when scanning all ObjC classes. When `delegateClassesToInstrument` is explicitly provided, probably the caller already knows which classes they want instrumented, so the check is unnecessary and works against them. So, I agree on the approach if:
- `delegateClassesToInstrument == nil` keep doing the selector scan
- `delegateClassesToInstrument != nil` then inject unconditionally into the provided classes

This should cover both your original use case and the general case without breaking anything

simonbility:
@ArielDemarco I made the changes as discussed!

Tests/InstrumentationTests/URLSessionTests/URLSessionInstrumentationTests.swift
```diff
@@ -990,6 +1018,69 @@ class URLSessionInstrumentationTests: XCTestCase {
...
+    let session = URLSession(configuration: URLSessionConfiguration.default, delegate: delegate, delegateQueue: nil)
+
+    let task = session.dataTask(with: request)
+    task.resume()
+
+    wait(timeout: 5) {
+      delegate.didReceiveResponseCalled
+    }
+    wait(timeout: 5) {
+      URLSessionInstrumentationTests.instrumentation.startedRequestSpans.isEmpty
```
ArielDemarco: Shouldnt this be `URLSessionInstrumentationTests.instrumentation.startedRequestSpans.!isEmpty`?

Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -517,6 +534,16 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
       URLSessionDataDelegate.urlSession(
         _:dataTask:didReceive:completionHandler:))
     guard let original = class_getInstanceMethod(cls, selector) else {
+      let block:
+        @convention(block) (Any, URLSession, URLSessionDataTask, URLResponse,
+                            @escaping (URLSession.ResponseDisposition) -> Void) -> Void = { _, session, dataTask, response, completion in
+          self.urlSession(session, dataTask: dataTask, didReceive: response,
+                          completionHandler: completion)
+          completion(.allow)
+        }
```
ArielDemarco, line 543: The internal `self.urlSession(...)` handler already calls completion (in `completionHandler`). The second `completion(.allow`) is unnecessary.
```suggestion
      let block:
        @convention(block) (Any, URLSession, URLSessionDataTask, URLResponse,
                            @escaping (URLSession.ResponseDisposition) -> Void) -> Void = { _, session, dataTask, response, completion in
          self.urlSession(session, dataTask: dataTask, didReceive: response,
                          completionHandler: completion)
        }
```

simonbility: From what i can tell its not actually calling it.

https://github.com/open-telemetry/opentelemetry-swift/blob/291fe3fff413ae9277ac36aec9fd9b51c1caa7e0/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift#L672-L691


In the previous implementation the original implementation that was being swapped/wrapped would do the calling.

But here its adding the implementation so thats no longer happening.

I think this is actually needed.

Tests/InstrumentationTests/URLSessionTests/URLSessionInstrumentationTests.swift
simonbility: TBH i havent actually been able to run these tests 😅 

There are some preexisting build failures that i prevented this.


I actually created #1120 in a effort to address these.

bryce-b:
Hi @simonbility, I understand that it's a bit confusing why the swizzling only works when certain methods exist in the user-implemented delegate, but this was done intentionally because the behavior of the NSURLSession can change depending on which methods are implemented, if we were to automatically add methods to the delegate that nuanced behavior will be changed, so a sort of 'leave-no-trace' mentality was adopted here. URLSessions can also be used in situations where a delegate is partially implemented without callbacks such as `urlSession(_:dataTask:didReceive:)`, where the handler is used by `URLSession.dataTask(with:completionHandler:)` instead, and that needs to be taken into account in the instrumentation as well. 

I'm not sure that these changes ought to be adopted, and we'll need some extensive testing with the interaction described above. We can discuss this more during our Thursday meeting.

simonbility:
Thanks for letting me know!

I have managed to work around this by just adding empty implementations of this in my own delegate classes, which i explicitly pass to the instrumentations constructor.


This works well for our case.

PR #1141, opened 2026-07-20
fix(otlp-http): bound flush semaphore wait in metric and log exporters

leedh2004:
Hi maintainers,

Follow-up to #1005: `OtlpHttpMetricExporter.flush()` and `OtlpHttpLogExporter.flush()` still block on a bare `semaphore.wait()`, so a request that never completes hangs the calling thread forever.

## The problem

- `flush()` in both the metric and log exporters waits on its completion semaphore with no timeout
- If the HTTP completion handler never fires, the calling thread is parked for good
- On iOS this can be fatal: a common integration flushes telemetry from a lifecycle callback on the main thread when the app goes to the background. If the request stalls there, the scene-update watchdog kills the app after 10 seconds (`0x8BADF00D`) — and the default `OtlpConfiguration.timeout` is also 10 seconds, so the request timeout can never fire first

The spec is explicit about this:

> Export() MUST NOT block indefinitely, there MUST be a reasonable upper limit after which the call must time out with an error result.

This is the same reason the semaphore timeout was requested in review on #1005, which fixed it for `OtlpHttpTraceExporter`. The metric and log exporters just weren't covered by that change.

## What I changed

- Both `flush()` implementations now bound the wait with the already-computed request timeout and return `.failure` when it expires — the same pattern the trace exporter uses since #1005:

```swift
let waitResult = semaphore.wait(timeout: .now() + timeout)
if waitResult == .timedOut {
  exporterMetrics?.addFailed(value: sentCount)
  return .failure
}
```

- Pending records stay in the queue on timeout, so the next export retries them instead of dropping data
- `export()` is left alone on purpose: the metric and log `export()` implementations don't wait on the semaphore at all today, so they don't have this defect. Making them synchronous like #1005 did for traces would be a bigger behavioral change and feels like a separate discussion.

## Tests

Added two tests that drive each exporter with an `HTTPClient` stub that fails the first request (to fill the pending queue) and then never invokes its completion handler. Before this change they hang; with it they fail at the configured timeout:

```
testLogFlushTimesOutWhenResponseNeverArrives     passed (0.504 seconds)
testMetricFlushTimesOutWhenResponseNeverArrives  passed (0.501 seconds)
```

Full OTLP exporter test target: 161 tests, 0 failures. `swift build --build-tests` clean.

Let me know if you want any changes. Thanks!

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocolHttp/logs/OtlpHttpLogExporter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolHttp/logs/OtlpHttpLogExporter.swift
@@ -107,7 +107,8 @@ public class OtlpHttpLogExporter: OtlpHttpExporterBase, LogRecordExporter, @unch
         }
       let semaphore = DispatchSemaphore(value: 0)
       var request = createRequest(body: body, endpoint: endpoint)
-      request.timeoutInterval = min(explicitTimeout ?? TimeInterval.greatestFiniteMagnitude, config.timeout)
+      let timeout = min(explicitTimeout ?? TimeInterval.greatestFiniteMagnitude, config.timeout)
+      request.timeoutInterval = timeout
       if let headers = envVarHeaders {
         headers.forEach { key, value in
           request.addValue(value, forHTTPHeaderField: key)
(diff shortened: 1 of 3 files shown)


bryce-b (approved):
lgtm
Thanks!

PR #1143, opened 2026-07-23
fix(exporter): log OTLP trace export failures

### LINKED ISSUE #334: Log errors when OTLPTraceExporter fails to send spans
When debugging https://github.com/open-telemetry/opentelemetry-swift/issues/333, I had a hard time figuring out where the problem was (it was a mismatch of protobuf definitions between the Swift SDK and the OTEL collector) because the SDK didn't report any type of failure, even though the protobuf definitions were mismatched with what was expected on the server (requests were completed successfully, but no spans were actually received by the collector).

Ideally, I would have liked a way to activate diagnostic logging in the SDK, and to then see error messages about the nature of the failure to send spans.

I do see this affordance, though I'm not sure it's the right way to get logs:

```swift 
var logger = Logger(label: "OTLPTraceExporter Logs", factory: StreamLogHandler.standardOutput)
logger.logLevel = .debug

// ...

let otlpTraceExporter = OtlpTraceExporter(channel: channel, logger: logger)
```

If that had printed useful error message, it would have been quite a bit easier to troubleshoot the problem.

aranhave:
- report trace export failures through the SDK feedback handler
- report partial-success responses when the collector rejects spans or returns a warning
- include the underlying gRPC error and partial-success details in the diagnostic message
- keep partial-success responses non-retryable, as required by OTLP

On Apple platforms, diagnostics use the default `os_log` feedback handler. Callers can replace it with `OpenTelemetry.registerFeedbackHandler`.

Fixes #334.

#### Example output

With a simulated unavailable collector:

```text
OTLP trace export failed: unavailable (14): collector unavailable
```

#### Scope

This PR only updates the gRPC trace exporter. Metric and log gRPC exporters can be handled separately.

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
@@ -55,7 +55,11 @@ public final class OtlpTraceExporter: SpanExporter, @unchecked Sendable {
 
     do {
       // wait() on the response to stop the program from exiting before the response is received.
-      _ = try export.response.wait()
+      let response = try export.response.wait()
+      if response.hasPartialSuccess {
+        logPartialSuccess(response.partialSuccess, logger: perCallOptions.logger)
+      }
+      // OTLP partial-success responses must not be retried.
       return .success
     } catch {
       perCallOptions.logger.error("OTLP trace export failed", error: error)
@@ -69,5 +73,29 @@ public final class OtlpTraceExporter: SpanExporter, @unchecked Sendable {
 
   public func shutdown(explicitTimeout: TimeInterval? = nil) {
     _ = channel.close()
+  }
+
+  private func logPartialSuccess(
+    _ partialSuccess: Opentelemetry_Proto_Collector_Trace_V1_ExportTracePartialSuccess,
+    logger: Logging.Logger
+  ) {
+    let rejectedSpans = partialSuccess.rejectedSpans
+    let errorMessage = partialSuccess.errorMessage
+    guard rejectedSpans != 0 || !errorMessage.isEmpty else {
+      return
+    }
+
+    var metadata: Logging.Logger.Metadata = [
+      "rejected_spans": .stringConvertible(rejectedSpans)
+    ]
+    if !errorMessage.isEmpty {
+      metadata["error_message"] = .string(errorMessage)
+    }
+
+    if rejectedSpans != 0 {
+      logger.error("OTLP trace export partially succeeded", metadata: metadata)
+    } else {
+      logger.warning("OTLP trace export succeeded with a warning", metadata: metadata)
+    }
   }
 }
--- a/Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
@@ -58,6 +58,7 @@ public final class OtlpTraceExporter: SpanExporter, @unchecked Sendable {
       _ = try export.response.wait()
       return .success
     } catch {
+      perCallOptions.logger.error("OTLP trace export failed", error: error)
       return .failure
     }
   }
--- a/Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
@@ -57,12 +57,12 @@ public final class OtlpTraceExporter: SpanExporter, @unchecked Sendable {
       // wait() on the response to stop the program from exiting before the response is received.
       let response = try export.response.wait()
       if response.hasPartialSuccess {
-        logPartialSuccess(response.partialSuccess, logger: perCallOptions.logger)
+        reportPartialSuccess(response.partialSuccess)
       }
       // OTLP partial-success responses must not be retried.
       return .success
     } catch {
-      perCallOptions.logger.error("OTLP trace export failed", error: error)
+      OpenTelemetry.instance.feedbackHandler?("OTLP trace export failed: \(error)")
       return .failure
     }
   }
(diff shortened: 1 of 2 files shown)


vvydier (approved):

aranhave:
Thanks for the review! 👍  I think CI is still waiting for someone with write access to approve the fork workflows.

Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
```diff
@@ -55,9 +55,14 @@ public final class OtlpTraceExporter: SpanExporter, @unchecked Sendable {
 
     do {
       // wait() on the response to stop the program from exiting before the response is received.
-      _ = try export.response.wait()
+      let response = try export.response.wait()
+      if response.hasPartialSuccess {
+        logPartialSuccess(response.partialSuccess, logger: perCallOptions.logger)
```
robert-northmind: One question on the usage of `perCallOptions.logger` here. What do you think of instead using the existing self-diagnostics channel for this? There is a `OpenTelemetry.instance.feedbackHandler` which was recently introduced. 
And in the sibling  HTTP trace exporter, it is used for exactly this purpose.

https://github.com/open-telemetry/opentelemetry-swift/blob/42f5c65851c0107d52ead2a7b98567eb68669dc9/Sources/Exporters/OpenTelemetryProtocolHttp/trace/OtlpHttpTraceExporter.swift#L88

Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
```diff
@@ -55,9 +55,14 @@ public final class OtlpTraceExporter: SpanExporter, @unchecked Sendable {
...
+      }
+      // OTLP partial-success responses must not be retried.
```
robert-northmind, line 62: Another question, not exactly relating to this change. but it applies to this part of the code. Maybe something we can think of to add in the future:

Should we also report metrics here? The `OtlpHttpTraceExporter` also reports exporterMetrics here.

Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
```diff
@@ -69,4 +74,28 @@ public final class OtlpTraceExporter: SpanExporter, @unchecked Sendable {
...
+
+    var metadata: Logging.Logger.Metadata = [
+      "rejected_spans": .stringConvertible(rejectedSpans)
+    ]
+    if !errorMessage.isEmpty {
+      metadata["error_message"] = .string(errorMessage)
+    }
+
+    if rejectedSpans != 0 {
+      logger.error("OTLP trace export partially succeeded", metadata: metadata)
```
robert-northmind: A small question on the severity here. So we log an error. But the method calling this returns a success. Should we maybe treat rejectedSpans as a `warning` instead of an `error`?

robert-northmind (commented):
Hey, looks good. I just left some questions. I am not that familiar yet with this codebase. So I might be wrong.

Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
aranhave: I’d keep the exporter logger here. #334 asks for failures through the logger passed to `OtlpTraceExporter`, and this keeps the gRPC error and partial-success details structured. The feedback handler is global and string-only, so aligning those paths feels like a separate change.

Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
aranhave: Yes, matching the HTTP exporter would be useful, but it needs the meter-provider initializer and separate success/failure accounting. I’d keep that as a follow-up so this PR stays focused on diagnostics.

Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
aranhave: `success` only prevents retrying, as required by OTLP. Rejected spans are still data loss, so I’d keep those at error. warning is for a fully accepted response with only a suggestion. The C++ and PHP exporters make the same distinction.

Sources/Exporters/OpenTelemetryProtocolGrpc/trace/OtlpTraceExporter.swift
robert-northmind: I see. And good point on the structured logs. The `OpenTelemetry.instance.feedbackHandler` does not have that today, so maybe something we should add in the future.

I just think that the CallOptions.logger isn't really an OTLP diagnostics channel, it might be more a gRPC-layer one. The default label `io.grpc` hints at that. 

And I think that the #334 issue creator is not really saying that they want to be able to use the passed in logger to the `OtlpTraceExporter`. I think they are just asking if that is the correct API to use. What they seem to be asking for is a way to get OTLP diagnostics out. 
And going down the road they ask: 

```swift
var logger = Logger(label: "OTLPTraceExporter Logs", factory: StreamLogHandler.standardOutput)
logger.logLevel = .debug
// ...
let otlpTraceExporter = OtlpTraceExporter(channel: channel, logger: logger)
```
That replaces grpc-swift's per-call logger, so grpc-swift's own RPC messages now get emitted under `OTLPTraceExporter Logs`.

It also seems that the feedbackHandler is what the rest of the SDK already uses for these purposes. All three HTTP OTLP exporters, `OtlpTraceJsonExporter`, `URLSessionInstrumentation`, and `MetricStorageRegistry` in core use it. And it defaults to os_log (on apple platforms), so it produces output without configuration.

aranhave: That makes sense. I was focused on keeping the extra details structured, but I agree the existing logger is really part of gRPC’s call configuration. I’ve moved the exporter diagnostics to `feedbackHandler` to match the rest of the SDK, and left richer structured diagnostics as a follow-up. Thanks for digging into this 🙂

robert-northmind (approved):

nachoBonafonte (approved):

aranhave:
@vvydier @nachoBonafonte @bryce-b Could someone with write access rerun all jobs on this PR? visionOS failed before checkout with a GitHub Service Unavailable error and macOS/watchOS were cancelled. The tests that ran passed, so rerunning the existing workflow should preserve the current approvals. 🙇 Please feel free to merge as I do not have access to merge this. thanks!

PR #1146, opened 2026-07-30
fix(otlp-http): return upload failures from log and metric exporters

yasuradodo:
## Summary

- Make `OtlpHttpLogExporter.export` wait for the HTTP callback and return `.failure` when upload fails or times out.
- Make `OtlpHttpMetricExporter.export` do the same using `config.timeout`.
- Add regression coverage for immediate failure, delayed failure, and no-response timeout paths.

## Motivation

Persistence deletes or keeps stored batches based on the exporter’s synchronous return value. The HTTP log and metric exporters were returning `.success` immediately after scheduling an async upload, so durable batches could be deleted before the upload outcome was known.

[OtlpHttpTraceExporter.export](https://github.com/yasuradodo/opentelemetry-swift/blob/59ff8edf81d7ba77b691438beaba1337274bab5b/Sources/Exporters/OpenTelemetryProtocolHttp/trace/OtlpHttpTraceExporter.swift#L57-L100) already waits for the callback and returns the real result. This change aligns log and metric export behavior with that existing contract.

## Notes

Timeout behavior remains trace-like: export returns `.failure` on timeout without adding new in-memory timeout requeue behavior.

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocolHttp/logs/OtlpHttpLogExporter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolHttp/logs/OtlpHttpLogExporter.swift
@@ -55,6 +55,7 @@ public class OtlpHttpLogExporter: OtlpHttpExporterBase, LogRecordExporter, @unch
 
   public func export(logRecords: [OpenTelemetrySdk.ReadableLogRecord],
                      explicitTimeout: TimeInterval? = nil) -> OpenTelemetrySdk.ExportResult {
+    var resultValue: ExportResult = .success
     var sendingLogRecords: [ReadableLogRecord] = []
     exporterLock.withLockVoid {
       pendingLogRecords.append(contentsOf: logRecords)
@@ -70,7 +71,9 @@ public class OtlpHttpLogExporter: OtlpHttpExporterBase, LogRecordExporter, @unch
 
     var request = createRequest(body: body, endpoint: endpoint)
     exporterMetrics?.addSeen(value: sendingLogRecords.count)
-    request.timeoutInterval = min(explicitTimeout ?? TimeInterval.greatestFiniteMagnitude, config.timeout)
+    let timeout = min(explicitTimeout ?? TimeInterval.greatestFiniteMagnitude, config.timeout)
+    let semaphore = DispatchSemaphore(value: 0)
+    request.timeoutInterval = timeout
     httpClient.send(request: request) { [weak self] result in
       switch result {
       case .success:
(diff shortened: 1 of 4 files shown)


Sources/Exporters/OpenTelemetryProtocolHttp/logs/OtlpHttpLogExporter.swift
```diff
@@ -81,10 +84,17 @@ public class OtlpHttpLogExporter: OtlpHttpExporterBase, LogRecordExporter, @unch
           self?.pendingLogRecords.append(contentsOf: sendingLogRecords)
```
aranhave, line 84: Now that this method returns `.failure`, persistence keeps the same stored batch for retry, but this also adds the batch to `pendingLogRecords`. On the next persistence attempt, `export()` appends that stored batch to the pending copy and sends every record twice. The metric exporter has the same retry path. Could we make one layer own the retry and add a test that fails once, retries through persistence, and verifies the second request contains one copy?

yasuradodo: Thank you for pointing this out. 

My understanding is that `Batch*Processor`  such as [BatchLogRecordProcessor](https://github.com/open-telemetry/opentelemetry-swift-core/blob/2.5.1/Sources/OpenTelemetrySdk/Logs/Processors/BatchLogRecordProcessor.swift) will retry exporting the pending signals in the next export or flush. 

[OtlpHttpTraceExporter](https://github.com/open-telemetry/opentelemetry-swift/blob/9a6d6a8aed22c415bb1673206e337824635f818b/Sources/Exporters/OpenTelemetryProtocolHttp/trace/OtlpHttpTraceExporter.swift#L85-L87) already has the same issue.

I would like to fix the issue of sending duplicate logs when the persistence exporter is enabled in another PR because it's a different problem. 

What do you think?

aranhave: Thanks for explaining. My concern is specifically the persistence path: returning .failure already makes the persistence decorator keep the batch, while the exporter also queues it internally, so the next attempt can send it twice. The trace exporter having the same behavior looks like an existing issue but this PR would extend it to logs and metrics. I’m not a maintainer, so I’ll leave the scope decision to them and keep this thread open for now.

yasuradodo: Hey @aranhave 👋
Yesterday, I opened an issue for this topic. 
https://github.com/open-telemetry/opentelemetry-swift/issues/1147

aranhave: Thanks for opening #1147 👍 It captures the retry interaction well. Since #1146 extends the duplicate-send behavior from traces to logs and metrics, I’d leave this thread open until a maintainer confirms that the fix can land separately. You can also bring it up in today's SIG meeting for faster review on this.

nachoBonafonte:
My main concern, as explained in the SIG meeting is if waiting on the export may block the Metric or the log generation. For traces there is a BatchSpanProcessor which processes spans for export in a different thread each, so waiting on it has no collateral effects, but I am not sure if metrics has a similar functionality or all the processing is done in the same thread so waiting on the export will imply the rest of the processing waiting on it, and this export can take long if reported remotely. The two options would be updating the framework or adding to this exporters compatibility with the recently added async methods and handle the errors only and properly on the async methods.

PR #1150, opened 2026-08-10
fix(MetricKit): accept unwrapped MXCallStackTree JSON in stacktrace transform

### LINKED ISSUE #1148: MetricKit: exception.stacktrace silently falls back to Apple format — transformStackTrace expects a callStackTree wrapper that MXCallStackTree.jsonRepresentation() doesn't emit
### What happened

  With a stock `MetricKitInstrumentation()` (no configuration, so `useAppleStacktraceFormat` keeps its default of `false`), crash and hang diagnostics emit `exception.stacktrace` in **Apple's native format**, not the simplified OpenTelemetry format documented in [`StackTraceFormat.md`](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Instrumentation/MetricKit/StackTraceFormat.md).

  Observed on the wire (OTLP capture at our collector, iOS app, `MetricKitInstrumentation()` with no config):

  ```json
  {
    "callStacks" : [
      {
        "threadAttributed" : true,
        "callStackRootFrames" : [
          {
            "binaryUUID" : "18A84A89-5321-40C9-A763-8CD5B39C6684",
            "offsetIntoBinaryTextSegment" : 4096,
            "sampleCount" : 15,
            "binaryName" : "testBinaryName",
            "address" : 4294971392
          }
        ]
      }
    ],
    "callStackPerThread" : true
  }
  ```

  Those are Apple's keys (`callStackRootFrames`, `offsetIntoBinaryTextSegment`, `sampleCount`, `address`) — the transform did not run.

  ### What I expected

### What I expected

Per `StackTraceFormat.md`, the default output should be a flat `callStackFrames` array with `binaryName` / `binaryUUID` / `offsetAddress`, no `sampleCount`, and no `address`.

### Root cause

[`StackTraceTransform.swift#L9-L11`](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Instrumentation/MetricKit/StackTraceTransform.swift#L9-L11) decodes into a type that **requires a top-level `callStackTree` key**:

```swift
struct AppleCallStackTree: Codable {
    let callStackTree: AppleCallStackTreeContent
}
```

But `MXCallStackTree.jsonRepresentation()` — the input passed in at [`MetricKitInstrumentation.swift#L481`](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Instrumentation/MetricKit/MetricKitInstrumentation.swift#L481) — emits the content *without* that wrapper: the root keys are `callStacks` and `callStackPerThread`. The `callStackTree` key does appear in `MXDiagnostic.dictionaryRepresentation()`, where the tree is a nested field, which is plausibly where the expected shape came from.

So `decoder.decode` throws, `catch { return nil }` fires, and the caller's `transformStackTrace(appleJson) ?? appleJson` falls back to the raw Apple JSON — silently, for both crash ([`#L484-L489`](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Instrumentation/MetricKit/MetricKitInstrumentation.swift#L484-L489)) and hang ([`#L415-L420`](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Instrumentation/MetricKit/MetricKitInstrumentation.swift#L415-L420)) diagnostics.

The README documents this fallback as a safety net for unexpected OS output, but as things stand it is the *only* path: the OTel format is unreachable regardless of configuration.

### Why the tests don't catch it

Every fixture in `StackTraceTransformTests.swift` is hand-written **with** the `callStackTree` wrapper, including `testTransformStackTrace_RemovesCallStackTreeWrapper`. The only negative case is `testTransformStackTrace_HandlesInvalidJSON`, which passes the string `"not valid json"`. No test exercises real `MXCallStackTree.jsonRepresentation()` output, so the shape mismatch is invisible to CI.

### Suggested fix

Accept the unwrapped shape, keeping the wrapper as a fallback:

```swift
let content: AppleCallStackTreeContent
if let unwrapped = try? decoder.decode(AppleCallStackTreeContent.self, from: appleJsonData) {
    content = unwrapped
} else {
    content = try decoder.decode(AppleCallStackTree.self, from: appleJsonData).callStackTree
}
```

benjoseph-grafana:
### Issue summary
`transformStackTrace` required a top-level `callStackTree` wrapper key, but
`MXCallStackTree.jsonRepresentation()` emits the tree unwrapped -- the wrapper
only appears when a whole `MXDiagnosticPayload` is serialized. Decoding always
threw on device, so every crash and hang diagnostic silently fell back to
Apple's native format via the `?? appleJson` path, and `exception.stacktrace`
shipped nested `subFrames`, `sampleCount` and ASLR-dependent address fields
instead of the documented OTel format.

### Fix
Decode the unwrapped form first and keep the wrapped form as a fallback, so
both shapes transform.

### Tests
The existing tests could not catch this: FakeCallStackTree hand-wrote the
wrapper that the real API never produces, and both format tests keyed off the
presence of that wrapper. Because the fallback output also has `callStacks`
and also lacks `callStackTree`, testConfiguration_UseOTelStacktraceFormat
passed whether the transform succeeded or failed.

- Fix the fake to emit what MXCallStackTree.jsonRepresentation() returns
- Re-anchor both format tests on callStackRootFrames vs callStackFrames,
the actual discriminator between the two formats
- Add coverage for unwrapped input, asserting subFrames, sampleCount and
address are dropped

Verified against a real MXCrashDiagnostic on iOS 26.5 (iPhone17,1): the OTLP
log record now carries flat callStackFrames with binaryName, binaryUUID and
offsetAddress only.

Resolves https://github.com/open-telemetry/opentelemetry-swift/issues/1148

### DIFF
--- a/Sources/Instrumentation/MetricKit/StackTraceTransform.swift
+++ b/Sources/Instrumentation/MetricKit/StackTraceTransform.swift
@@ -68,12 +68,15 @@
         }
 
         do {
-            // Decode Apple's format
+            // Decode Apple's format. `MXCallStackTree.jsonRepresentation()` emits the tree
+            // unwrapped; the `callStackTree` wrapper only appears when a whole
+            // MXDiagnosticPayload is serialized. Accept both.
             let decoder = JSONDecoder()
-            let appleTree = try decoder.decode(AppleCallStackTree.self, from: appleJsonData)
+            let appleTree = try (try? decoder.decode(AppleCallStackTreeContent.self, from: appleJsonData))
+                ?? decoder.decode(AppleCallStackTree.self, from: appleJsonData).callStackTree
 
             // Transform to simplified format
-            let simplifiedCallStacks = appleTree.callStackTree.callStacks.map { appleStack in
+            let simplifiedCallStacks = appleTree.callStacks.map { appleStack in
                 // Flatten all root frames and their subframes
                 let allFrames = appleStack.callStackRootFrames.flatMap { flattenFrames($0) }
 
@@ -84,7 +87,7 @@
             }
 
             let simplifiedTree = SimplifiedCallStackTree(
-                callStackPerThread: appleTree.callStackTree.callStackPerThread,
+                callStackPerThread: appleTree.callStackPerThread,
                 callStacks: simplifiedCallStacks
             )
 
--- a/Sources/Instrumentation/MetricKit/StackTraceFormat.md
+++ b/Sources/Instrumentation/MetricKit/StackTraceFormat.md
@@ -85,7 +85,7 @@ This format is based on Apple's `MXCallStackTree.jsonRepresentation()` format bu
 
 ### 1. Removed `callStackTree` Wrapper
 
-The `callStackTree` wrapper adds an unnecessary level of nesting. Since the entire JSON document represents call stack data, the wrapper provides no additional information and only complicates parsing.
+`MXCallStackTree.jsonRepresentation()` emits the tree unwrapped; the `callStackTree` wrapper only appears when a whole `MXDiagnosticPayload` is serialized. Either way, the wrapper adds an unnecessary level of nesting: since the entire JSON document represents call stack data, it provides no additional information and only complicates parsing, so this format omits it and accepts both shapes as input.
 
 ### 2. Flattened Stack Frames
 
--- a/Tests/InstrumentationTests/MetricKitTests/MetricKitInstrumentationTests.swift
+++ b/Tests/InstrumentationTests/MetricKitTests/MetricKitInstrumentationTests.swift
@@ -642,8 +642,10 @@ class MetricKitDiagnosticTests: XCTestCase {
         let json = try? JSONSerialization.jsonObject(with: stacktraceData) as? [String: Any]
 
         XCTAssertNotNil(json)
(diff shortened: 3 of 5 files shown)


Sources/Instrumentation/MetricKit/StackTraceTransform.swift
```diff
@@ -68,12 +68,15 @@
         }
 
         do {
-            // Decode Apple's format
+            // Decode Apple's format. `MXCallStackTree.jsonRepresentation()` emits the tree
+            // unwrapped; the `callStackTree` wrapper only appears when a whole
+            // MXDiagnosticPayload is serialized. Accept both.
```
aranhave, line 73: Since this changes our understanding of Apple's direct JSON shape, can we update StackTraceFormat.md here too? Its "Removed callStackTree Wrapper" section still says MXCallStackTree.jsonRepresentation() includes that wrapper, which would leave the format docs contradicting this code and the new fixture.

benjoseph-grafana: done.

bryce-b (approved):

aranhave (approved):

PR #1156, opened 2026-08-15
feat(otlp): support dynamic export headers

### LINKED ISSUE #651: [OTLPExporter] Allow headers to be updated in initialized exporters
## Details

User reported issue #650, requests the ability to update auth tokens when they expire, but the exporters do not provide this functionality. We ought to add something akin to the android/java sdk, if possible. 

## solution

Let's add a callback to the OTLP configuration that checks prior to export any changes to the header file, or other mutable connection options (if feasible)

aranhave:
Closes #651

Added an optional headers provider to `OtlpConfiguration` so expiring auth tokens can be refreshed without rebuilding exporters. It is evaluated for each HTTP and gRPC export while preserving static and environment header behavior.

Also:
- Fixes duplicate headers when retrying HTTP log exports.
- Enforces `config.timeout` for gRPC metric exports, which was previously ignored.

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocolCommon/common/OtlpConfiguration.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolCommon/common/OtlpConfiguration.swift
@@ -26,17 +26,33 @@ public struct OtlpConfiguration: Sendable {
   // let endpoint : URL? = URL(string: "https://localhost:4317")
   // let certificateFile
   // let compression
+  /// Static headers included with every export when `headersProvider` is not set.
   public let headers: [(String, String)]?
+  /// Returns the current headers immediately before each export.
+  ///
+  /// This callback may be invoked concurrently. Callers must synchronize any mutable state
+  /// captured by the callback. When set, it takes precedence over `headers`. Headers supplied
+  /// through an exporter's `envVarHeaders` parameter retain their existing precedence.
+  public let headersProvider: (@Sendable () -> [(String, String)]?)?
   public let timeout: TimeInterval
   public let compression: CompressionType
   public let exportAsJson: Bool
   public init(timeout: TimeInterval = OtlpConfiguration.DefaultTimeoutInterval,
               compression: CompressionType = .gzip,
               headers: [(String, String)]? = nil,
-              exportAsJson: Bool = true) {
+              exportAsJson: Bool = true,
+              headersProvider: (@Sendable () -> [(String, String)]?)? = nil) {
     self.headers = headers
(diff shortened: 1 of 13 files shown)


Tests/ExportersTests/OpenTelemetryProtocol/OtlpHttpExporterBaseCoverageTests.swift
```diff
@@ -41,6 +41,72 @@ final class OtlpHttpExporterBaseCoverageTests: XCTestCase {
...
+
+    let firstRequest = try exporter.createRequest(
+      body: makeBody(),
+      endpoint: XCTUnwrap(URL(string: "http://example.com"))
+    )
+    provider.update([("Authorization", "Bearer second")])
+    let secondRequest = try exporter.createRequest(
+      body: makeBody(),
+      endpoint: XCTUnwrap(URL(string: "http://example.com"))
+    )
```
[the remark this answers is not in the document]
aranhave: `createRequest` is non-throwing but `XCTUnwrap` in the endpoint argument throws, so the `try` is required. Dropping it just flips the error to `call can throw but is not marked with 'try'`. Same for the other two spots.

Sources/Exporters/OpenTelemetryProtocolCommon/common/OtlpConfiguration.swift
```diff
@@ -26,17 +26,38 @@ public struct OtlpConfiguration: Sendable {
...
+    self.headersProvider = headersProvider
     self.timeout = timeout
     self.compression = compression
     self.exportAsJson = exportAsJson
   }
+
+  /// Returns the headers to include in the next export.
+  ///
+  /// When a provider is configured, its result takes precedence over static headers, including
+  /// when the provider returns `nil`.
```
benjoseph-grafana: What is the rationale behind this replacement? Wouldn't a merge with overwrite be a better approach?

benjoseph-grafana (approved):

Sources/Exporters/OpenTelemetryProtocolCommon/common/OtlpConfiguration.swift
aranhave: I originally treated the provider as the complete header snapshot but that makes static metadata awkward to use with rotating credentials. I’ve changed it to merge both, with provider values replacing matching names case-insensitively and envVarHeaders still taking precedence.

nachoBonafonte (approved):

PR #1157, opened 2026-08-16
fix(URLSession): swizzle only once to avoid duplicate spans

tomerhy:
## Problem

`URLSessionInstrumentation.init` unconditionally calls `injectInNSURLClasses()`. Each swizzle captures the currently installed implementation as its "original" and installs a new one in front of it:

- instance 1 captures Apple's implementation, installs `swizzled_A`
- instance 2 captures **`swizzled_A`**, installs `swizzled_B`

A request then runs `swizzled_B` → span → `swizzled_A` → **second span** → Apple's implementation. N instrumentations produce N spans per request.

This is reachable in any app where two SDKs each instrument `URLSession`, and in any SDK that re-creates the instrumentation after a configuration change.

## Fix

Guard the installation behind a process-wide flag so it runs at most once, regardless of how many `URLSessionInstrumentation` instances are created.

## Note for reviewers

With the guard in place, a second instrumentation's configuration is not applied — the first instance's configuration remains in effect. This is documented on `init`.

An alternative would be to keep a static reference to the most recently created instrumentation so the installed implementations consult the newest configuration. That is a lifecycle change rather than a bug fix, so I kept it out of this PR, but I'm happy to add it if you'd prefer that behaviour.

## Test

`testSecondInstrumentationDoesNotSwizzleAgain` asserts the installed `IMP` for `dataTask(with:)` is unchanged after a second instrumentation is created.

Verified the test fails without the fix (the `IMP` changes, confirming the second instance re-swizzled) and passes with it. Full `URLSessionInstrumentationTests` suite: 58 tests, 0 failures.

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -79,21 +79,19 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
   /// instance remains the one applied to instrumented requests.
   public init(configuration: URLSessionInstrumentationConfiguration) {
     self._configuration = configuration
-    guard Self.claimNSURLClassInjection() else { return }
-    injectInNSURLClasses()
-  }
 
-  /// Returns `true` for the first caller only.
-  ///
-  /// Each call to `injectInNSURLClasses()` captures the currently installed implementation as its
-  /// original and installs a new one in front of it. A second instrumentation would therefore
-  /// chain itself ahead of the first, and every request would be reported once per instance.
-  private static func claimNSURLClassInjection() -> Bool {
-    injectionLock.lock()
-    defer { injectionLock.unlock() }
-    guard !hasInjectedIntoNSURLClasses else { return false }
-    hasInjectedIntoNSURLClasses = true
-    return true
+    // The lock is held across the installation, not only across setting the flag. Each call to
+    // injectInNSURLClasses() captures the currently installed implementation as its original, so a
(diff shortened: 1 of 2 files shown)


Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -65,11 +72,30 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
...
   }
 
+  /// Creates the instrumentation and installs it into `URLSession`.
+  ///
+  /// Installation happens once per process. If an instrumentation has already been created, this
+  /// instance is still usable but does not re-install itself, and the configuration of the first
+  /// instance remains the one applied to instrumented requests.
   public init(configuration: URLSessionInstrumentationConfiguration) {
     self._configuration = configuration
+    guard Self.claimNSURLClassInjection() else { return }
```
aranhave: Can later initializers wait until the first installation finishes? The flag is set before `injectInNSURLClasses()` runs, so a concurrent second init can return while swizzling is still in progress.

tomerhy: You're right — the flag was set inside the lock but the swizzling ran after releasing it, so a second init could return mid-installation. Moved the injection inside the lock so a later initializer waits for the first to finish, in `65a34cb`. I couldn't get a deterministic test around this one since the flag is process-wide and already set by the time the tests run, so it's verified by inspection rather than a failing test.

nachoBonafonte (approved):

PR #1158, opened 2026-08-16
fix(URLSession): reuse the task span when resume fires more than once

tomerhy:
## Problem

`urlSessionTaskWillResume` starts a span for the task. But `resume()` can fire more than once for a single logical request — a suspend/resume cycle, `NSURLSession` superclass chaining, and redirect handling all re-enter it.

Each additional resume calls `processAndLogRequest` again, which overwrites `runningSpans[taskId]`. Two things go wrong:

1. The earlier span is orphaned — it is never ended, so it is never exported.
2. The `traceparent` header was injected when the **first** span was created and is already on the wire. The span that eventually gets reported carries a **different** span id.

The second is the damaging one: it isn't a dropped span, it's a *wrong* one. The server records a parent id that never appears in the exported trace, so correlation breaks silently while everything still looks healthy.

## Why this wasn't caught

The guard for this already exists in the file — it was gated to watchOS, with the comment noting that *"other platforms don't exhibit these re-entries in practice (verified via URLSessionInstrumentationTests)"*.

They do. A suspend/resume cycle produces **three** spans for a single request on macOS.

The existing tests couldn't see it: they assert `traceparent` is present in 14 places, but never that it matches the span actually reported. `testReportedSpanMatchesTraceparentSentOnTheWire` closes that gap.

## Fix

Remove the `#if os(watchOS)` fence so the existing guard applies on all platforms. No logic change beyond that.

## Tradeoff

The original comment gated this "to avoid the per-resume queue sync". That cost is now paid on every platform: one dictionary lookup inside `runningSpansQueue.sync`, once per resume. That seems clearly worth it against silently mis-attributed spans, and it is the same cost watchOS already pays — but I'm happy to discuss if you'd rather see it optimised (e.g. an associated-object flag on the task instead of a queue round-trip).

## Tests

- `testRepeatedResumeDoesNotStartASecondSpan` — fails without the fix (`3` spans vs expected `1`), passes with it.
- `testReportedSpanMatchesTraceparentSentOnTheWire` — asserts the reported span id equals the span id in the `traceparent` that went out. Passes before and after; it guards the invariant that had no coverage.

Full `URLSessionInstrumentationTests` suite: 59 tests, 0 failures.

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -771,21 +771,18 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
         requestMap[taskId]?.setRequest(request)
       }
 
-      #if os(watchOS)
-      // watchOS-only defensive guard: a span may have already been started for this
-      // task by a factory swizzle (e.g. dataTask(with:completionHandler:)) re-entered
-      // from data(for:), or by a previous resume (NSURLSession super-class chaining
-      // and redirect handling can both cause resume to fire more than once per logical
-      // request). Starting another one here would orphan the existing span:
-      // processAndLogRequest would overwrite runningSpans[taskId] and the first span
-      // would never be ended. Other platforms don't exhibit these re-entries in
-      // practice (verified via URLSessionInstrumentationTests), so this lookup is
-      // gated to watchOS to avoid the per-resume queue sync.
+      // A span may have already been started for this task by a factory swizzle
+      // (e.g. dataTask(with:completionHandler:)) re-entered from data(for:), or by a
+      // previous resume: suspend/resume, NSURLSession super-class chaining and redirect
+      // handling can all cause resume to fire more than once per logical request.
+      // Starting another one here would orphan the existing span, since
+      // processAndLogRequest overwrites runningSpans[taskId] and the first span would
+      // never be ended. It would also disagree with the wire, because the traceparent
+      // already sent to the server carries the first span's id.
       let alreadyTracked = URLSessionLogger.runningSpansQueue.sync {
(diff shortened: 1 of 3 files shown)


Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -771,21 +771,18 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
...
-      // gated to watchOS to avoid the per-resume queue sync.
+      // A span may have already been started for this task by a factory swizzle
+      // (e.g. dataTask(with:completionHandler:)) re-entered from data(for:), or by a
+      // previous resume: suspend/resume, NSURLSession super-class chaining and redirect
+      // handling can all cause resume to fire more than once per logical request.
+      // Starting another one here would orphan the existing span, since
+      // processAndLogRequest overwrites runningSpans[taskId] and the first span would
+      // never be ended. It would also disagree with the wire, because the traceparent
+      // already sent to the server carries the first span's id.
       let alreadyTracked = URLSessionLogger.runningSpansQueue.sync {
```
aranhave: Can the check and span claim be made atomic? Two threads can call `resume()` on the same task, both see no entry here, and both reach `processAndLogRequest`, recreating the duplicate span this is fixing.

tomerhy: You're right, and it was worse than the check — I reproduced it and got two spans every time. Turned out `idKeyForTask` also reads and assigns the task's id without synchronization, so the racing threads generated *different* ids for the same task and were tracked as two separate requests; making the claim atomic alone didn't help until that was fixed too. Both are in `51017b9`, with a test that resumes the same task from two threads and asserts a single span.

aranhave (approved):

nachoBonafonte (approved):

PR #1159, opened 2026-08-16
fix(URLSession): only instrument http and https requests

tomerhy:
## Problem

`processAndLogRequest` instruments every request it is handed. Nothing checks the URL scheme, so a request that is not an HTTP call still gets:

- a span named `HTTP {method}` carrying `http.*` semantic attributes, and
- a `traceparent` header injected into it when header injection is enabled.

Verified against `main`: `file:///tmp/example.txt`, `ws://localhost/socket` and `data:text/plain,hello` each start a span today.

This shows up wherever an app loads local files, opens a WebSocket through `URLSession`, or serves a custom scheme via `URLProtocol`. The result is telemetry describing HTTP requests that never happened, mixed in with real ones — and tracing headers attached to requests no HTTP server will ever read.

## Fix

Return early from `processAndLogRequest` unless the request URL scheme is `http` or `https`.

The check sits at the single funnel every instrumented path already goes through, so it covers the delegate, completion-handler, async and factory-swizzle routes in one place.

## Note for reviewers

The guard runs **before** `shouldInstrument`, so that callback is no longer invoked for non-HTTP requests. That seemed right — it is a filter for choosing among HTTP requests, not a way to opt a `file://` URL into HTTP semantics — but it does mean an app currently relying on custom-scheme instrumentation would stop receiving those spans. Happy to move the check after `shouldInstrument` instead if you would rather leave that escape hatch open.

## Tests

`testNonHTTPSchemesAreNotInstrumented` asserts no span is started for `file:`, `ws:` and `data:` URLs. It fails on unmodified `main` — all three produce a `SpanSdk` — and passes with the fix.

Full `URLSessionInstrumentationTests` suite: 58 tests, 0 failures. No existing test depended on non-HTTP requests being instrumented.

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionLogger.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionLogger.swift
@@ -48,8 +48,23 @@ class URLSessionLogger {
     }()
   #endif // os(iOS) && !targetEnvironment(macCatalyst)
 
+  /// The HTTP semantic conventions describe http and https traffic. A request on any other scheme
+  /// is not an HTTP call, so a span carrying HTTP attributes would misreport it.
+  static func isHTTPScheme(_ url: URL?) -> Bool {
+    guard let scheme = url?.scheme?.lowercased() else {
+      return false
+    }
+    return scheme == "http" || scheme == "https"
+  }
+
   /// This methods creates a Span for a request, and optionally injects tracing headers, returns a  new request if it was needed to create a new one to add the tracing headers
   @discardableResult static func processAndLogRequest(_ request: URLRequest, sessionTaskId: String, instrumentation: URLSessionInstrumentation, shouldInjectHeaders: Bool) -> URLRequest? {
+    // Schemes such as file:, data:, ws: and app-specific URLProtocol schemes would otherwise get a
+    // span describing an HTTP call that never happened, and tracing headers no server will read.
+    guard isHTTPScheme(request.url) else {
+      return nil
+    }
+
     #if os(watchOS)
     // watchOS routes a single user-level URLSession call (e.g. `data(for:)`) through
     // two distinct URLSessionTask objects: an outer task whose `resume()` is intercepted
--- a/Sources/Instrumentation/URLSession/URLSessionLogger.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionLogger.swift
@@ -48,20 +48,24 @@ class URLSessionLogger {
     }()
   #endif // os(iOS) && !targetEnvironment(macCatalyst)
 
-  /// The HTTP semantic conventions describe http and https traffic. A request on any other scheme
-  /// is not an HTTP call, so a span carrying HTTP attributes would misreport it.
-  static func isHTTPScheme(_ url: URL?) -> Bool {
+  /// Whether a request on this scheme is an HTTP call, and so describable by the HTTP semantic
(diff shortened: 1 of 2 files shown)


Sources/Instrumentation/URLSession/URLSessionLogger.swift
```diff
@@ -48,8 +48,23 @@ class URLSessionLogger {
...
+  static func isHTTPScheme(_ url: URL?) -> Bool {
+    guard let scheme = url?.scheme?.lowercased() else {
+      return false
+    }
+    return scheme == "http" || scheme == "https"
+  }
+
   /// This methods creates a Span for a request, and optionally injects tracing headers, returns a  new request if it was needed to create a new one to add the tracing headers
   @discardableResult static func processAndLogRequest(_ request: URLRequest, sessionTaskId: String, instrumentation: URLSessionInstrumentation, shouldInjectHeaders: Bool) -> URLRequest? {
+    // Schemes such as file:, data:, ws: and app-specific URLProtocol schemes would otherwise get a
```
aranhave: Should `ws` and `wss` stay instrumented for the WebSocket opening handshake? `URLSessionWebSocketTask` sends an HTTP handshake and preserves request headers, so this drops both its client span and `traceparent` propagation.

tomerhy: Tested this and it's not a concern — a real `ws://` task still gets its span and its `traceparent`. Turns out iOS rewrites the scheme behind the scenes: by the time the handshake reaches us, `currentRequest` already says `http://` (and `https://` for `wss`), so it passes the filter. Good catch though, since nothing documents that — I've added a test in `c4b91d4` pinning it, so it can't quietly break later.

aranhave: I tested the `URLRequest` overload too, and it behaves differently. `webSocketTask(with: URLRequest)` keeps `ws/wss` in both `originalRequest` and `currentRequest`, so this guard exits before creating the span or injecting `traceparent`. The new test only covers the URL overload, which Foundation rewrites to HTTP. Switching that test to the `URLRequest` overload fails.

```sh
swift -e 'import Foundation; let s = URLSession(configuration: .ephemeral); let t = s.webSocketTask(with: URLRequest(url: URL(string: "ws://localhost:33333/socket")!)); print(t.currentRequest?.url?.scheme ?? "nil")'
```

Can we preserve `ws/wss` here and cover the `URLRequest` overload too? That avoids depending on the URL-overload rewrite.

tomerhy: You're right, and my test was covering the wrong overload — I ran your one-liner and `webSocketTask(with: URLRequest)` does keep `ws` on both `originalRequest` and `currentRequest`, so the guard was dropping the span and the traceparent for it. The URL overload only passed because Foundation rewrites it to `http`, which is exactly the case my test happened to hit.

Allowed `ws`/`wss` through — the handshake is an HTTP request after all — and switched the test to the `URLRequest` overload, which fails without the change (nil scheme, no traceparent). Also renamed the helper to `isInstrumentableScheme`, since a function called `isHTTPScheme` returning true for `ws` would be misleading. In `5ed1ad7`.

nachoBonafonte (approved):

aranhave (approved):

PR #1161, opened 2026-08-18
Fix link format in Example Network Sample README.md

sauerborn-git:
The Link was pointing to a dead end, fixed it to point to the correct Readme.

### DIFF
--- a/Examples/Network Sample/README.md
+++ b/Examples/Network Sample/README.md
@@ -1,4 +1,4 @@
 ### Network sample
 
 This example shows the `URLSessionInstrumentation` in action. 
-It only shows the minimum configuration: Just by initializing the class, all network request that happen after it will be captured by Opentelemetry as Spans. Check [`URLSessionInstrumentation` README](Sources/Instrumentation/URLSession/README.md) for more information.
+It only shows the minimum configuration: Just by initializing the class, all network request that happen after it will be captured by Opentelemetry as Spans. Check [`URLSessionInstrumentation` README](/Sources/Instrumentation/URLSession/README.md) for more information.


Examples/Network Sample/README.md
```diff
@@ -1,4 +1,4 @@
 ### Network sample
 
 This example shows the `URLSessionInstrumentation` in action. 
-It only shows the minimum configuration: Just by initializing the class, all network request that happen after it will be captured by Opentelemetry as Spans. Check [`URLSessionInstrumentation` README](Sources/Instrumentation/URLSession/README.md) for more information.
+It only shows the minimum configuration: Just by initializing the class, all network request that happen after it will be captured by Opentelemetry as Spans. Check [`URLSessionInstrumentation` README](/Sources/Instrumentation/URLSession/README.md) for more information.
```
aranhave, line 4: This still resolves from the GitHub domain root, so it opens `github.com/Sources/...` rather than this repository. Can we use `../../Sources/Instrumentation/URLSession/README.md` here?

sauerborn-git: Works fine for me, it resolves to the repo root, you can test it here: 
https://github.com/sauerborn-git/opentelemetry-swift/blob/f7e24d71ac4582771ffd5e0ec8404430a4dcec22/Examples/Network%20Sample/README.md

aranhave: Thanks for confirming! I checked the rendered target again , seems like github does resolve it from the repo root.

aranhave (approved):

nachoBonafonte (approved):

PR #1162, opened 2026-08-19
fix(URLSession): instrument session delegates the startup scan cannot find

tomerhy:
## Problem

`injectInNSURLClasses` finds delegate classes by scanning each class's method list for one of six selectors. A session delegate that implements **none** of them is never discovered, so none of the completion reporting is ever installed for it.

`urlSessionTaskWillResume` still starts a span for every task on that session. Nothing ever ends it, so it stays in `runningSpans` for the lifetime of the process — an unbounded leak of live spans, and no telemetry for any request made through that session.

This is reachable with a delegate as ordinary as:

```swift
final class MyDelegate: NSObject, URLSessionDelegate {
  func urlSession(_ session: URLSession, didBecomeInvalidWithError error: Error?) {}
}
```

Verified against `main`: one span leaks per request on such a session.

## Fix

Instrument the session's delegate class when a task resumes, at which point the delegate is known.

Because a class can now be reached by both the startup scan and a resuming task, injection is tracked in a set — injecting twice would install the swizzle on top of itself and report every completion once per installation, the same failure mode as #1157.

## Why the existing tests didn't catch it

`tearDown` clears `runningSpans` before asserting it is empty, so a leaked span is invisible to the suite. The new test inspects `startedRequestSpans` immediately after the task completes, before any cleanup.

## Note for reviewers

Two smaller things this touches on, which I have deliberately left alone:

- `injectTaskDidFinishCollectingMetricsIntoDelegateClass` adds the method with `class_addMethod` when the class lacks it, but `injectTaskDidCompleteWithErrorIntoDelegateClass` returns instead. That asymmetry means completion reporting on a discovered-but-incomplete delegate relies on the metrics path, which is gated to macOS 13 / iOS 16. Worth aligning, but it is a separate change.
- Lazy injection at resume works because Obj-C dispatch is dynamic; I verified a delegate class instrumented after its session was created still receives the added callbacks.

## Tests

`testSpanIsEndedWhenSessionDelegateImplementsNoInstrumentedSelector` fails on unmodified `main` (`1 still running`) and passes with the fix.

Full `URLSessionInstrumentationTests` suite: 58 tests, 0 failures.

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -142,12 +142,17 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
 
   /// Instruments `cls` unless it already has been. Injecting twice would report each task
   /// completion once per installation.
+  ///
+  /// The lock is held across the injection, not just the bookkeeping: a caller that finds the class
+  /// already recorded must be able to rely on its callbacks being installed, otherwise a task
+  /// resuming concurrently could start before the delegate has been modified.
   private func instrumentDelegateClassIfNeeded(_ cls: AnyClass) {
     Self.instrumentedDelegateClassesLock.lock()
-    let isNew = Self.instrumentedDelegateClasses.insert(ObjectIdentifier(cls)).inserted
-    Self.instrumentedDelegateClassesLock.unlock()
+    defer { Self.instrumentedDelegateClassesLock.unlock() }
 
-    guard isNew else { return }
+    guard Self.instrumentedDelegateClasses.insert(ObjectIdentifier(cls)).inserted else {
+      return
+    }
     injectIntoDelegateClass(cls: cls)
   }
 
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -565,6 +565,16 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
     let selector = #selector(
       URLSessionDataDelegate.urlSession(_:task:didCompleteWithError:))
     guard let original = class_getInstanceMethod(cls, selector) else {
+      // The method is optional and the delegate omits it, so nothing would report completion for
+      // its tasks. Add it, as injectTaskDidFinishCollectingMetricsIntoDelegateClass already does —
+      // that fallback only exists on newer platforms, so this is the one that covers the rest.
+      let block:
+        @convention(block) (Any, URLSession, URLSessionTask, Error?) -> Void = { _, session, task, error in
+          self.urlSession(session, task: task, didCompleteWithError: error)
+        }
+      let imp = imp_implementationWithBlock(
+        unsafeBitCast(block, to: AnyObject.self))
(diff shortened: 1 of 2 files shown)


Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -135,6 +140,17 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
     injectIntoNSURLSessionTaskResume()
   }
 
+  /// Instruments `cls` unless it already has been. Injecting twice would report each task
+  /// completion once per installation.
+  private func instrumentDelegateClassIfNeeded(_ cls: AnyClass) {
+    Self.instrumentedDelegateClassesLock.lock()
+    let isNew = Self.instrumentedDelegateClasses.insert(ObjectIdentifier(cls)).inserted
```
aranhave: Can this stay in an installing state until the swizzling finishes? A concurrent task can see the class in this set, skip injection, and start before its completion callbacks have been modified.

Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -763,6 +779,14 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
     }
 
     let taskId = idKeyForTask(task)
+
+    // A session delegate that implements none of the selectors scanned in injectInNSURLClasses is
+    // never discovered there, so without this nothing would report this task's completion and the
+    // span started below would never be ended.
+    if let sessionDelegate = (task.value(forKey: "session") as? URLSession)?.delegate {
+      instrumentDelegateClassIfNeeded(type(of: sessionDelegate))
```
aranhave, line 826: How does this end the span on iOS 15 or macOS 12? A delegate with no `didCompleteWithError` is not modified there, and the metrics fallback is only installed on iOS 16/macOS 13 and newer.

Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
tomerhy: Good catch — you're right, there's a window. I add the class to the set before the swizzling actually runs, so another task can see it as already done and start before its completion callbacks are installed. I'll move the injection inside the lock so a second caller waits for it to finish — happy to use an explicit installing state instead if you'd rather keep per-class parallelism.

Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
tomerhy: You're right — I disabled the metrics fallback locally to stand in for iOS 15, and the span still leaked, so this was only working on iOS 16+. I've added `didCompleteWithError` when the delegate doesn't implement it, same as the metrics injector already does, so completion gets reported on the older versions too. Pushed in `ff2e738`.

aranhave (approved):
Looks good now. Holding the lock through installation closes the race, and adding didCompleteWithError covers the older OS path too. Thanks for addressing both!

nachoBonafonte (approved):

aranhave (approved):

nachoBonafonte:
Hi @tomerhy can you please resolve the conflicts?

PR #1163, opened 2026-08-19
fix(URLSession): honour shouldRecordPayload on the completion handler paths

tomerhy:
## Problem

`shouldRecordPayload` is opt-in and defaults to **off**. The session delegate path respects that — `urlSession(_:dataTask:didReceive:)` and `urlSession(_:dataTask:didReceive:completionHandler:)` both guard on it before accumulating anything.

The completion handler paths do not. Both wrappers in `injectIntoNSURLSessionAsyncDataAndDownloadTaskMethods` pass the response object straight through:

```swift
URLSessionLogger.logResponse(response, dataOrFile: object, ...)
URLSessionLogger.logError(error!, dataOrFile: object, ...)
```

`dataOrFile` is handed to the `receivedResponse` / `receivedError` callbacks. So an application that never enabled payload recording still has response bodies delivered into its telemetry callbacks for every `dataTask(with:completionHandler:)` request — and those callbacks exist to attach information to spans, which is where it would end up.

## Fix

Withhold the payload from the telemetry callbacks unless recording is enabled:

```swift
let payload = (self.configuration.shouldRecordPayload?(session) ?? false) ? object : nil
```

The caller's own completion handler still receives `object` unchanged — only what reaches the instrumentation callbacks is gated. Applied to both wrappers.

## Question for maintainers

The documentation on `shouldRecordPayload` says *"This callback is only necessary when using session delegate"*, so the current behaviour may be intentional rather than an oversight. I think it is worth changing regardless: an option that reads as "payload recording is off by default" not applying to two of the three capture paths is surprising, and it is the kind of surprise that matters because the data involved is response bodies.

That said, this **is** a behaviour change — anyone currently relying on bodies arriving without setting the predicate would stop receiving them. If you would rather keep that and correct the documentation instead, say so and I will turn this into a docs PR.

## Tests

`testCompletionHandlerHonoursShouldRecordPayload` asserts that with recording disabled the caller's completion handler still receives its data while `receivedResponse` receives `nil`. It fails on unmodified `main`, where a `Data` instance arrives despite recording being off.

The test uses the existing `/success` route, whose body is empty — enough to show the path is ungated, since an empty `Data` still arrives where `nil` is expected. Note that no route in `HttpTestServer` returns a non-empty body, which is why payload behaviour has had no coverage until now. Adding one seemed out of scope here, but it is worth doing separately.

Full `URLSessionInstrumentationTests` suite: 58 tests, 0 failures.

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -353,13 +353,16 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
           if completionBlock != nil {
             if objc_getAssociatedObject(argument, &idKey) == nil {
               let completionWrapper: (Any?, URLResponse?, Error?) -> Void = { object, response, error in
+                // Payload recording is opt-in and off by default. The caller still receives
+                // `object` below; only what is handed to the telemetry callbacks is withheld.
+                let payload = (self.configuration.shouldRecordPayload?(session) ?? false) ? object : nil
                 if error != nil {
                   let status = (response as? HTTPURLResponse)?.statusCode ?? 0
-                  URLSessionLogger.logError(error!, dataOrFile: object, statusCode: status,
+                  URLSessionLogger.logError(error!, dataOrFile: payload, statusCode: status,
                                             instrumentation: self, sessionTaskId: sessionTaskId)
                 } else {
                   if let response {
-                    URLSessionLogger.logResponse(response, dataOrFile: object, instrumentation: self,
+                    URLSessionLogger.logResponse(response, dataOrFile: payload, instrumentation: self,
                                                  sessionTaskId: sessionTaskId)
                   }
                 }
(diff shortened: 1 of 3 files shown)


Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -353,13 +353,16 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
           if completionBlock != nil {
             if objc_getAssociatedObject(argument, &idKey) == nil {
               let completionWrapper: (Any?, URLResponse?, Error?) -> Void = { object, response, error in
+                // Payload recording is opt-in and off by default. The caller still receives
+                // `object` below; only what is handed to the telemetry callbacks is withheld.
+                let payload = (self.configuration.shouldRecordPayload?(session) ?? false) ? object : nil
```
aranhave, line 382: Can we update the `shouldRecordPayload` API docs too? They still say this callback is only needed with a session delegate, but completion-handler callbacks now use it as their opt-in gate.

tomerhy: @aranhave done

aranhave: Thanks, the docs look right now. Since this changes existing behavior, I’ll leave the final call to the maintainers.

nachoBonafonte (approved):

nachoBonafonte:
This looks good to me, also maybe adding clearly to the documentation that this parameter is not functional for asynchronous methods, which is a known but I think undocumented limitation

aranhave (approved):

bryce-b:
@tomerhy This merge conflict needs to be resolved before merging.
The other PRs may need this resolution as well.

PR #1167, opened 2026-08-20
fix(URLSession): end the span for tasks nothing else reports

tomerhy:
We hit this in production and have been carrying a fix for it in our SDK, so I'm upstreaming it.

If a task has no session delegate and no completion handler, the factory swizzle starts its span and then nothing ever ends it. There's nothing left to observe the completion, so the span just sits in `runningSpans` for the rest of the process. Same thing happens with networking layers that manage their own tasks.

Easy to see on `main`:

```swift
let session = URLSession(configuration: .ephemeral)
let task = session.dataTask(with: request)   // no completion handler, no delegate
task.resume()
// after the task completes: startedRequestSpans.count == 1
```

`setState:` is the one place every task goes through on its way to `.completed`, so I swizzle it and end the span there if it's still running. If a delegate or a completion handler already reported the task, its span is gone from `runningSpans` and this is a no-op.

One detail worth calling out: tasks created with a completion handler get marked and skipped here. That handler reports the response itself and it has the body, so letting the fallback get there first would mean reporting the same request with less data.

The existing tests couldn't catch the leak because `tearDown` clears `runningSpans` before it asserts it's empty. The new test looks right after the task completes instead. It fails on `main` with `1 still running` and passes with this.

Full URLSession suite is green, and the whole package suite (597 tests) too.

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -25,6 +25,7 @@ struct NetworkRequestState {
 }
 
 nonisolated(unsafe) private var idKey: Void?
+nonisolated(unsafe) private var completionHandlerKey: Void?
 
 public final class URLSessionInstrumentation: @unchecked Sendable {
   private var requestMap = [String: NetworkRequestState]()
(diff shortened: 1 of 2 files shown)


aranhave (approved):

nachoBonafonte (approved):
This is a great idea, it has always been a limitation. Have you tested on async URL handling, does it work also there?

PR #1168, opened 2026-08-20
feat(URLSession): give receivedResponse the request it belongs to

tomerhy:
Another one we've been carrying in our SDK.

`receivedResponse` hands you the response, the payload and the span, but not the request they came from. If you want to correlate a response with what was actually sent — request headers, the final URL, the body you posted — you have to keep your own table of in-flight requests keyed by task and clean it up yourself. We ended up doing exactly that, which is why I'd rather it just came through the callback.

The request is already sitting in `requestMap`, it was just being cleared before the logging ran. So this keeps the entry until after the callbacks have run and passes it along.

```swift
receivedResponse: { response, dataOrFile, span, request in
    // request is the one this response belongs to
}
```

Two things for you to weigh:

- This changes the signature of `receivedResponse`, so anyone who wrote that closure with explicit parameters has to add one. If you'd rather not break it, I can add a separate callback instead and leave the existing one alone — happy either way, just tell me which you prefer.
- `receivedError` still doesn't get the request. I left it out to keep this small, but it has the same problem if you want it.

New test asserts the request arrives and matches the URL that was requested. It fails if the lookup returns nil, so it's actually checking something. URLSession suite green, full package suite green (597 tests).

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentationConfiguration.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentationConfiguration.swift
@@ -31,7 +31,8 @@ public struct URLSessionInstrumentationConfiguration {
               shouldInjectTracingHeaders: ((URLRequest) -> (Bool)?)? = nil,
               injectCustomHeaders: ((inout URLRequest, Span?) -> Void)? = nil,
               createdRequest: ((URLRequest, Span) -> Void)? = nil,
-              receivedResponse: ((URLResponse, DataOrFile?, Span, URLRequest?) -> Void)? = nil,
+              receivedResponse: ((URLResponse, DataOrFile?, Span) -> Void)? = nil,
+              receivedResponseWithRequest: ((URLResponse, DataOrFile?, Span, URLRequest?) -> Void)? = nil,
               receivedError: ((Error, DataOrFile?, HTTPStatus, Span) -> Void)? = nil,
               delegateClassesToInstrument: [AnyClass]? = nil,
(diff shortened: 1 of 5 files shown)


Sources/Instrumentation/URLSession/URLSessionInstrumentationConfiguration.swift
```diff
@@ -31,7 +31,7 @@ public struct URLSessionInstrumentationConfiguration {
               shouldInjectTracingHeaders: ((URLRequest) -> (Bool)?)? = nil,
               injectCustomHeaders: ((inout URLRequest, Span?) -> Void)? = nil,
               createdRequest: ((URLRequest, Span) -> Void)? = nil,
-              receivedResponse: ((URLResponse, DataOrFile?, Span) -> Void)? = nil,
+              receivedResponse: ((URLResponse, DataOrFile?, Span, URLRequest?) -> Void)? = nil,
```
aranhave: This changes the public callback type and breaks existing three-argument `receivedResponse` closures. Since this is a released 2.x API, can we keep the current callback and add a separate request-aware one instead?

tomerhy: Fair enough — that's the tradeoff I flagged, and you're right that breaking a released callback isn't worth it. Left `receivedResponse` exactly as it was and added `receivedResponseWithRequest` next to it; both get called if both are implemented. The test config in the suite is back to the three-argument form and compiles unchanged, so nothing existing has to move. Happy to rename the new one if you'd prefer something else.

nachoBonafonte (approved):
Great improvement

aranhave (approved):

PR #1169, opened 2026-08-20
feat(URLSession): allow payload recording to be decided per request

tomerhy:
Last of the batch from our SDK.

`shouldRecordPayload` asks about the *session*. In our app, and I'd guess in most, one session is shared across everything, so that callback can only answer for all requests at once. We want bodies from a couple of endpoints and definitely not from the rest, and there's no way to say that today — it's all or nothing.

So this adds `shouldRecordPayloadForRequest`, which gets the request instead of the session:

```swift
shouldRecordPayloadForRequest: { request in
    request.url?.path.hasPrefix("/api/orders") == true
}
```

It's asked first, and falls back to `shouldRecordPayload` when it isn't implemented or returns nil, so nothing changes for anyone already using the session-level one.

Two notes:

- This sits next to #1163, which makes the completion handler paths honour `shouldRecordPayload` at all. If you'd rather land that one first and have me rebase this on top, that's fine by me.
- The test needed a response with an actual body, and no route in `HttpTestServer` returns one, so I added a small `URLProtocol` stub scoped to a single sentinel host (`payload.stub`). It only claims requests for that host, so registering it can't affect anything else in the suite. Happy to move it somewhere more general if you'd prefer it reusable.

Test asserts the callback is asked with the request that's being sent, and fails if the per-request lookup is skipped. URLSession suite green, full package suite green (597 tests).

### DIFF
--- a/Sources/Instrumentation/URLSession/README.md
+++ b/Sources/Instrumentation/URLSession/README.md
@@ -12,7 +12,9 @@ This behaviour can be modified or augmented by using the optional callbacks defi
 
 `shouldInstrument: ((URLRequest) -> (Bool)?)?` :  Filter which requests you want to instrument, all by default
 
-`shouldRecordPayload: ((URLSession) -> (Bool)?)?`: Implement if you want the session to record payload data, false by default.
+`shouldRecordPayload: ((URLSession) -> (Bool)?)?`: Implement if you want the session to record payload data, false by default. It gates the payload passed to `receivedResponse` and `receivedError` on every path, whether the request uses the session delegate, a completion handler or async/await.
+
+`shouldRecordPayloadForRequest: ((URLRequest) -> (Bool)?)?`: Decide payload recording per request rather than per session, which is useful when one session serves endpoints that should be treated differently. Asked first, falling back to `shouldRecordPayload` when it is not implemented or returns nil.
 
 `shouldInjectTracingHeaders: ((URLRequest) -> (Bool)?)?`: Allows filtering which requests you want to inject headers to follow the trace, true by default. You must also return true if you want to inject custom headers.
 
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -353,13 +353,16 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
           if completionBlock != nil {
             if objc_getAssociatedObject(argument, &idKey) == nil {
               let completionWrapper: (Any?, URLResponse?, Error?) -> Void = { object, response, error in
+                // Payload recording is opt-in and off by default. The caller still receives
+                // `object` below; only what is handed to the telemetry callbacks is withheld.
+                let payload = (self.configuration.shouldRecordPayload?(session) ?? false) ? object : nil
                 if error != nil {
                   let status = (response as? HTTPURLResponse)?.statusCode ?? 0
-                  URLSessionLogger.logError(error!, dataOrFile: object, statusCode: status,
+                  URLSessionLogger.logError(error!, dataOrFile: payload, statusCode: status,
                                             instrumentation: self, sessionTaskId: sessionTaskId)
                 } else {
                   if let response {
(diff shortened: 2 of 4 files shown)


Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -858,6 +858,19 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
     return id!
   }
 
+  /// Whether the payload should be recorded, asking per request first and falling back to the
+  /// per-session callback.
+  private func shouldRecordPayload(for session: URLSession, taskId: String?) -> Bool {
+    let config = configuration
+    if let perRequest = config.shouldRecordPayloadForRequest,
```
aranhave, line 901: Can this gate the completion-handler paths too? I merged this with #1163 locally, and those wrappers still call `shouldRecordPayload(session)` directly, so `shouldRecordPayloadForRequest` is ignored for data, download, and upload tasks using completion handlers. A test for that path would pin the new API consistently.

Sources/Instrumentation/URLSession/URLSessionInstrumentationConfiguration.swift
```diff
@@ -66,6 +68,13 @@ public struct URLSessionInstrumentationConfiguration {
   /// This callback is only necessary when using session delegate
   public var shouldRecordPayload: ((URLSession) -> (Bool)?)?
 
+  /// Implement this callback to decide payload recording per request rather than per session.
+  ///
+  /// A session is usually shared across every call an app makes, so `shouldRecordPayload` can only
+  /// answer for all of them at once. This is consulted first, and `shouldRecordPayload` is used when
+  /// it is not implemented or returns nil, so existing behaviour is unchanged.
+  public var shouldRecordPayloadForRequest: ((URLRequest) -> (Bool)?)?
```
aranhave, line 83: Please add this callback to `Sources/Instrumentation/URLSession/README.md` too. That guide lists the configuration callbacks but still only exposes the session-level payload gate.

Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
tomerhy: Good catch — you're right, I only changed the delegate guards. I pulled #1163 into this branch like you did locally and routed all four call sites through the same helper, so the per-request callback now decides for the completion handler paths too. Added a test for a completion handler task that checks the callback is asked and the body actually comes through. Note this branch now carries #1163's commits, so if that one lands first I'll rebase and they'll drop out.

Sources/Instrumentation/URLSession/URLSessionInstrumentationConfiguration.swift
tomerhy: Added it, plus a line on `shouldRecordPayload` saying it gates every path, not just the delegate one. Worth flagging that #1168 changes `receivedResponse` to take the request, so that README entry goes stale too — I'll update it there.

nachoBonafonte (approved):
LGTM

Sources/Instrumentation/URLSession/README.md
```diff
@@ -12,7 +12,9 @@ This behaviour can be modified or augmented by using the optional callbacks defi
 
 `shouldInstrument: ((URLRequest) -> (Bool)?)?` :  Filter which requests you want to instrument, all by default
 
-`shouldRecordPayload: ((URLSession) -> (Bool)?)?`: Implement if you want the session to record payload data, false by default.
+`shouldRecordPayload: ((URLSession) -> (Bool)?)?`: Implement if you want the session to record payload data, false by default. It gates the payload passed to `receivedResponse` and `receivedError` on every path, whether the request uses the session delegate, a completion handler or async/await.
```
aranhave: Can we either support async/await here or remove it from this sentence? `AsyncTaskDelegate` always passes `dataOrFile: nil` and never calls either payload predicate. I reproduced it with `data(from:)`: the per-request callback was not invoked and `receivedResponse` got no payload.

aranhave (approved):
Looks good! Just need to fix the conflicts.

PR #1171, opened 2026-08-20
fix(otlp-http): add requeueOnFailure to avoid duplicate Persistence exports

### LINKED ISSUE #1147: Persistence + OTLP HTTP exporters can upload the same batch twice (disk retry + in-memory pending)
## Description

When an OTLP HTTP exporter is wrapped by a Persistence exporter decorator, a failed upload can be retried by two layers at once.

1. Persistence keeps the batch on disk when the decorated exporter returns `.failure` (`needsRetry: true`) and retries it later.
2. The OTLP HTTP exporters also re-queue the same batch into an in-memory pending list (`pendingSpans` / `pendingLogRecords` / `pendingMetrics`) on HTTP failure.

On the next persistence-driven `export()`, the OTLP exporter appends the disk batch to its pending list and sends the combined list. As a result, the failed batch can be uploaded twice.

This already happens with traces via `OtlpHttpTraceExporter`. [#1146](https://github.com/open-telemetry/opentelemetry-swift/pull/1146) makes the log and metric exporters return real upload failures so Persistence can keep failed batches. That change is needed for Persistence, but it also extends this double-retry behavior to logs and metrics.

Related discussion: https://github.com/open-telemetry/opentelemetry-swift/pull/1146#discussion_r3691251752

## Steps to reproduce

1. Wrap `OtlpHttpTraceExporter` with `PersistenceSpanExporterDecorator`.
2. Disable the network.
3. Emit spans. They are written to disk; the Persistence worker later calls the OTLP exporter, which fails, stores them in `pendingSpans`, and Persistence keeps the file for retry.
4. Re-enable the network.
5. Force flush, or wait for the next Persistence export cycle.

## Expected behavior

Each failed batch is retried once and uploaded once.

## Actual behavior

The same spans can be sent from both the Persistence retry path and the OTLP in-memory pending queue. This often shows up as duplicate payloads on flush or on the next export.

## Notes

Without Persistence, the in-memory pending queue is a useful retry mechanism. With Persistence wrapping the exporter, it is unclear which layer should own retry for a failed batch.

I am happy to contribute a fix once maintainers decide the preferred direction for how these two retry paths should interact.

yasuradodo:
Fixes #1147

## Summary

When `PersistenceSpanExporterDecorator` wraps an OTLP HTTP exporter, a failed upload was retried twice: Persistence kept the batch on disk while the HTTP exporter also re-appended it to the in-memory pending queue. The next export merged both copies and sent duplicate spans.

This adds `requeueOnFailure: Bool = true` on `OtlpHttpExporterBase` (trace, log, and metric HTTP exporters). Set `requeueOnFailure: false` when wrapping with Persistence so disk retry owns failed batches.

## Test plan

- [x] `swift test --filter 'OtlpHttpLogExporterCoverageTests|OtlpHttpTraceExporterCoverageTests|OtlpHttpExporterFlushTimeoutTests'`
- [x] `swift test --filter PersistenceSpanExporterStackedWithOtlpHttpTests`

<details>
<summary>Known limitations</summary>

Log and metric HTTP exporters still return `.success` before the HTTP callback completes. With `requeueOnFailure: false`, in-memory duplication is prevented, but Persistence disk retry for logs/metrics still needs separate failure-signaling work (#1146). Trace is correct end-to-end.
</details>

### DIFF
--- a/Tests/ExportersTests/PersistenceExporter/PersistenceSpanExporterStackedWithOtlpHttpTests.swift
+++ b/Tests/ExportersTests/PersistenceExporter/PersistenceSpanExporterStackedWithOtlpHttpTests.swift
@@ -1,0 +1,152 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import Foundation
+#if canImport(FoundationNetworking)
+  import FoundationNetworking
+#endif
+import OpenTelemetryApi
+import OpenTelemetryProtocolExporterCommon
+import OpenTelemetryProtocolExporterHttp
+import OpenTelemetrySdk
+@testable import PersistenceExporter
+import SwiftProtobuf
+import XCTest
+
+/// When Persistence wraps an OTLP HTTP trace exporter, both layers used to retry
+/// the same failed batch. With `requeueOnFailure: false`, the exporter leaves its
+/// pending queue empty so Persistence can retry from disk without duplicating spans.
+final class PersistenceSpanExporterStackedWithOtlpHttpTests: XCTestCase {
+  @UniqueTemporaryDirectory private var temporaryDirectory: Directory
+
+  private final class FailOnceHTTPClient: HTTPClient, @unchecked Sendable {
+    struct SendFailure: Error {}
+
+    private let lock = NSLock()
+    private var _sentBodies: [Data] = []
+    private let firstRequestSent: XCTestExpectation
+    private let secondRequestSent: XCTestExpectation
+
+    var sentBodies: [Data] {
+      lock.lock()
+      defer { lock.unlock() }
+      return _sentBodies
+    }
+
+    init(firstRequestSent: XCTestExpectation, secondRequestSent: XCTestExpectation) {
+      self.firstRequestSent = firstRequestSent
+      self.secondRequestSent = secondRequestSent
+    }
+
+    func send(request: URLRequest,
+              completion: @escaping (Result<HTTPURLResponse, Error>) -> Void) {
+      lock.lock()
+      _sentBodies.append(request.httpBody ?? Data())
+      let requestCount = _sentBodies.count
+      lock.unlock()
+
+      // First request fails; later requests succeed (Persistence retry).
+      if requestCount == 1 {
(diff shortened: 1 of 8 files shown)


Tests/ExportersTests/PersistenceExporter/PersistenceSpanExporterStackedWithOtlpHttpTests.swift
```diff
@@ -0,0 +1,152 @@
...
+import OpenTelemetryProtocolExporterHttp
+import OpenTelemetrySdk
+@testable import PersistenceExporter
+import SwiftProtobuf
+import XCTest
+
+/// When Persistence wraps an OTLP HTTP trace exporter, both layers used to retry
+/// the same failed batch. With `requeueOnFailure: false`, the exporter leaves its
+/// pending queue empty so Persistence can retry from disk without duplicating spans.
+final class PersistenceSpanExporterStackedWithOtlpHttpTests: XCTestCase {
```
yasuradodo, line 21: Testing only Span (Trace) because Log and Metric have another problem. 
More context: https://github.com/open-telemetry/opentelemetry-swift/pull/1146

nachoBonafonte (approved):

yasuradodo:
Hi @nachoBonafonte 

Thank you for taking care of my pull requests 🙏 

As we talked in the last SIG meeting,  I would like to get a new version with this fix quickly.
This is mainly because "Persistence Exporter" doesn't work in the currently released versions. 

With this fix, at least, I can start using the exporter for traces.
I understand this is not a high priority for the Swift OTel project,  and I am sorry to ask you again, but I would really appreciate it if we could release 2.5.2 with the fix soon. 

If there's anything I can do to help release a new version, I'd be happy to support.

nachoBonafonte:
Hi @yasuradodo, I have just released 2.5.2 as prerelease, that means it won't be downloaded by default for the latest version but you can use it by using directly the version number. Let me know if it works for you

yasuradodo:
> Hi @yasuradodo, I have just released 2.5.2 as prerelease, that means it won't be downloaded by default for the latest version but you can use it by using directly the version number. Let me know if it works for you

Thank you very much! It works 🙏 
Have a nice weekend 👋

Issue #895, opened 2025-08-21
Document short falls of network instrumentation regarding async requests


bryce-b:
related issue https://github.com/open-telemetry/opentelemetry-swift/issues/860
TODO:
* Add a readme to the NSURLSession instrumentation library that explains some of the edge cases due to limitations of the instrumentation. see related issue.

PR #1172, opened 2026-08-20
docs(URLSession): document the initialization cost and how to avoid it

tomerhy:
Docs only.

Creating `URLSessionInstrumentation` searches every class loaded in the process to find session delegates. There's already a way to skip that — pass `delegateClassesToInstrument` — but neither that option nor `ignoredClassPrefixes` appears in the README, so the only discoverable behaviour is the expensive default.

This documents both options and adds a short section on what the search costs and how to avoid it.

The reason I went looking: on #895 @williazz reports the instrumentation blocking app launch by ~500 ms across devices and simulators, and says the only workaround they found was deferring initialization, which then drops early requests. The escape hatch they needed already exists in the API, it just isn't written down anywhere.

I've tried to be straight about the trade-off rather than just recommending it — if you pass an explicit list, a delegate you leave out isn't instrumented and its requests aren't captured. And I noted that deferring initialization is not an equivalent workaround, for the reason williazz hit.

No code changes. The snippet compiles (checked against the current initializer).

### DIFF
--- a/Sources/Instrumentation/URLSession/README.md
+++ b/Sources/Instrumentation/URLSession/README.md
@@ -30,8 +30,6 @@ This behaviour can be modified or augmented by using the optional callbacks defi
 
 `delegateClassesToInstrument: [AnyClass]?`: The session delegate classes to instrument. When this is `nil`, the instrumentation discovers them by examining **every class loaded in the process** at initialization, which is the default. Passing your delegate classes explicitly skips that search — see [Initialization cost](#initialization-cost) below.
 
-`ignoredClassPrefixes: [String]?`: Class name prefixes to leave out of that search.
-
 `baggageProvider: ((inout URLRequest, Span) -> (Baggage)?)?`: Provides baggage instance for instrumented requests that is merged with active baggage. The callback receives URLRequest and Span parameters to create dynamic baggage based on request context. The resulting baggage is injected into request headers using the configured propagator.
 
 ## Initialization cost
--- a/Sources/Instrumentation/URLSession/README.md
+++ b/Sources/Instrumentation/URLSession/README.md
@@ -28,5 +28,26 @@ This behaviour can be modified or augmented by using the optional callbacks defi
 
 `receivedError: ((Error, DataOrFile?, HTTPStatus, Span) -> Void)?` -  Called after an error is received,  it allows to add extra information to the Span
 
+`delegateClassesToInstrument: [AnyClass]?`: The session delegate classes to instrument. When this is `nil`, the instrumentation discovers them by examining **every class loaded in the process** at initialization, which is the default. Passing your delegate classes explicitly skips that search — see [Initialization cost](#initialization-cost) below.
+
+`ignoredClassPrefixes: [String]?`: Class name prefixes to leave out of that search.
+
 `baggageProvider: ((inout URLRequest, Span) -> (Baggage)?)?`: Provides baggage instance for instrumented requests that is merged with active baggage. The callback receives URLRequest and Span parameters to create dynamic baggage based on request context. The resulting baggage is injected into request headers using the configured propagator.
 
+## Initialization cost
+
+Creating `URLSessionInstrumentation` searches every class loaded in the process for session delegate methods, so that delegates are instrumented before any request runs. The cost of that search grows with the number of classes an app links, and in large apps it has been measured in the hundreds of milliseconds. Because it happens during initialization, that time lands on whatever is waiting for it, usually app launch.
+
+If you know your session delegate classes, pass them and the search is skipped entirely:
+
+```swift
+URLSessionInstrumentation(
+  configuration: URLSessionInstrumentationConfiguration(
+    delegateClassesToInstrument: [MyAPIClientDelegate.self, MyUploadDelegate.self]
+  )
+)
+```
+
+Only the classes you list are instrumented, so a delegate you leave out is not, and requests made through it are not captured.
+
+Note that deferring the initialization to get it off the launch path is not equivalent: requests made before the instrumentation exists are not captured, so early network calls go missing instead.


williazz:
Did you reproduce the bug by any chance?

tomerhy:
@williazz yeah, I measured it. In a test process with ~35k classes loaded, walking every class's method list takes **~285 ms** (3 runs: 285/282/290) and finds about 19 matching classes. `objc_getClassList` itself is only ~5-15 ms, so it's the per-class method walk that costs, and it scales with how many classes your app links rather than with anything URLSession-related. Your ~500 ms on a real app with more frameworks linked lines up with that.

Two caveats on my numbers: I timed the enumeration and method walk in isolation rather than the shipped `init` (which also swizzles the matches), and the match count was collected without synchronization so treat it as approximate.

Passing `delegateClassesToInstrument` skips the search entirely, which is what this PR documents, so that should unblock you now — with the trade-off that a delegate you don't list isn't instrumented.

Longer term I think the answer is discovering delegate classes on demand instead of scanning everything up front. I'll open an issue for that so the performance side is tracked on its own rather than buried in a docs PR.

tomerhy:
Sharing how I measured, since my number came from timing the scan in isolation rather than the shipped code path.

If you want the number for your own app, this measures the real `init`:

```swift
var classCount: UInt32 = 0
_ = objc_copyClassList(&classCount)

let t0 = ProcessInfo.processInfo.systemUptime
let instrumentation = URLSessionInstrumentation(configuration: URLSessionInstrumentationConfiguration())
let ms = (ProcessInfo.processInfo.systemUptime - t0) * 1000
print("classes: \(classCount), URLSessionInstrumentation init: \(ms) ms")
```

One warning from my own mistake: don't time a second `init` to compare against `delegateClassesToInstrument`. Since #1157 the swizzling is installed once per process, so the second call returns immediately and you measure ~0 ms and conclude the explicit list is free when you've really measured nothing. Comparing the two paths needs separate launches.

And this is the harness behind the ~285 ms, so the caveat is visible rather than just asserted — it reimplements what `injectInNSURLClasses` does, minus the swizzling:

```swift
let selectors = [
  #selector(URLSessionDataDelegate.urlSession(_:dataTask:didReceive:)),
  #selector(URLSessionDataDelegate.urlSession(_:dataTask:didReceive:completionHandler:)),
  #selector(URLSessionDataDelegate.urlSession(_:task:didCompleteWithError:)),
  #selector(URLSessionTaskDelegate.urlSession(_:task:didFinishCollecting:))
]

let t0 = ProcessInfo.processInfo.systemUptime
let classes = InstrumentationUtils.objc_getClassList()
let listed = ProcessInfo.processInfo.systemUptime

DispatchQueue.concurrentPerform(iterations: classes.count) { i in
  var methodCount: UInt32 = 0
  guard let methodList = class_copyMethodList(classes[i], &methodCount) else { return }
  defer { free(methodList) }
  for j in 0 ..< Int(methodCount) where selectors.contains(method_getName(methodList[j])) { break }
}
let done = ProcessInfo.processInfo.systemUptime

print("getClassList \((listed - t0) * 1000) ms, method walk \((done - listed) * 1000) ms")
```

On my machine, 35,417 classes: `getClassList` ~5-15 ms, method walk 270-285 ms, over three runs. Worth noting it uses four of the six selectors the real code checks, so if anything it understates the walk slightly.

Sources/Instrumentation/URLSession/README.md
```diff
@@ -28,5 +28,26 @@ This behaviour can be modified or augmented by using the optional callbacks defi
 
 `receivedError: ((Error, DataOrFile?, HTTPStatus, Span) -> Void)?` -  Called after an error is received,  it allows to add extra information to the Span
 
+`delegateClassesToInstrument: [AnyClass]?`: The session delegate classes to instrument. When this is `nil`, the instrumentation discovers them by examining **every class loaded in the process** at initialization, which is the default. Passing your delegate classes explicitly skips that search — see [Initialization cost](#initialization-cost) below.
+
+`ignoredClassPrefixes: [String]?`: Class name prefixes to leave out of that search.
```
aranhave: Can we remove this claim for now? `injectInNSURLClasses()` calls `objc_getClassList()` directly, and `ignoredClassPrefixes` is never read after configuration, so setting it does not exclude classes or reduce the initialization scan.

tomerhy: You're right, removed. Confirmed it's assigned in the initializer and never read anywhere — the search uses the hardcoded `excludeList` instead, so setting `ignoredClassPrefixes` has no effect at all. I shouldn't have documented it without checking it was wired up.

Worth deciding separately whether to wire it up or deprecate it, since right now it's public API that silently does nothing. Happy to open an issue for that if useful.

Sources/Instrumentation/URLSession/README.md
```diff
@@ -28,5 +28,24 @@ This behaviour can be modified or augmented by using the optional callbacks defi
...
+
+```swift
+URLSessionInstrumentation(
+  configuration: URLSessionInstrumentationConfiguration(
+    delegateClassesToInstrument: [MyAPIClientDelegate.self, MyUploadDelegate.self]
+  )
+)
+```
+
+Only the classes you list are instrumented, so a delegate you leave out is not, and requests made through it are not captured.
```
aranhave, line 49: Small wording fix: completion-handler requests are still captured even when their session's delegate is omitted. I verified this with a local HTTP request: both span creation and completion ran. Would it be clearer to say that callbacks on unlisted delegates are not instrumented?

aranhave (approved):

Sources/Instrumentation/URLSession/README.md
tomerhy: @aranhave You're right, reworded. The completion-handler path is swizzled on `URLSession` itself (`injectIntoNSURLSessionCreateTaskWithParameterMethods`), not through the delegate, so those requests are captured regardless of what's in the list. One thing I noticed while checking - if you omit a delegate you actually use, the span still starts at task creation but nothing ends it, since `didCompleteWithError` is only swizzled on listed classes - so I added a line warning about that too.

PR #1183, opened 2026-08-28
feat(sessions): add explicit linked reset

### LINKED ISSUE #1180: Sessions: clarify activity semantics and add explicit reset
### Decision

The current default remains unchanged: application telemetry is session activity. `getSession()` refreshes the inactivity deadline, and the span and log processors continue to use that behavior.

An opt-in refresh policy or signal filter may still be useful, but it is a separate follow-up. This issue and [#1183](https://github.com/open-telemetry/opentelemetry-swift/pull/1183) are now limited to explicit linked reset.

### Problem

There is no public operation for ending the current session and immediately starting a linked replacement. Apps need this boundary after events such as sign-out or an account change.

### Proposed direction

Add an explicit reset operation to `SessionManager`. Reset should end the current session, create and persist one replacement, and link it through `session.previous_id` without changing the existing activity or timeout behavior.

The exact API name remains open to review in [#1183](https://github.com/open-telemetry/opentelemetry-swift/pull/1183).

### Acceptance criteria

- [ ] `getSession()` and automatic span and log attribution continue to refresh inactivity by default.
- [ ] Explicit reset creates one replacement whose `session.previous_id` is the previous session ID.
- [ ] Reset emits one end event for the previous session and one start event for the replacement, in that order.
- [ ] The replacement is persisted before reset returns.
- [ ] Concurrent access and reset preserve one ordered session chain without duplicate lifecycle events.
- [ ] Expiry behavior remains unchanged.
- [ ] Tests cover reset, persistence, lifecycle event ordering, expiry, and concurrent access.

### Out of scope

A configurable refresh filter is not part of this change. If pursued, it should preserve all-signal refresh as the default.

This follows the lifecycle configuration introduced in [#1107](https://github.com/open-telemetry/opentelemetry-swift/pull/1107).

aranhave:
## Summary

- Add `SessionManager.resetSession()` for explicit session boundaries such as sign-out or account changes.
- Create and persist a replacement session linked through `session.previous_id`.
- Keep session end and start events ordered during concurrent and reentrant resets.
- Avoid re-entering session access while publishing session lifecycle telemetry.
- Preserve the existing behavior where `getSession()`, spans, and logs refresh session activity.

## Validation

- `swift test --enable-code-coverage`: 633 tests passed.
- 131 session tests passed with Thread Sanitizer on an ARM64 iPhone 17 simulator running iOS 26.5, with no sanitizer findings.

Closes #1180

### DIFF
--- a/Sources/Instrumentation/Sessions/SessionLogRecordProcessor.swift
+++ b/Sources/Instrumentation/Sessions/SessionLogRecordProcessor.swift
@@ -34,7 +34,7 @@ public class SessionLogRecordProcessor: LogRecordProcessor {
 
     // Only add session attributes if they don't already exist
     if logRecord.attributes[SemanticConventions.Session.id.rawValue] == nil || logRecord.attributes[SemanticConventions.Session.previousId.rawValue] == nil {
-      let session = sessionManager.getSessionForAttribution()
+      let session = sessionManager.getSession()
 
       // Add session.id if not already present
       if logRecord.attributes[SemanticConventions.Session.id.rawValue] == nil {
--- a/Sources/Instrumentation/Sessions/SessionLogRecordProcessor.swift
+++ b/Sources/Instrumentation/Sessions/SessionLogRecordProcessor.swift
@@ -23,7 +23,8 @@ public class SessionLogRecordProcessor: LogRecordProcessor {
   /// Called when a log record is emitted - adds session attributes and forwards to next processor
   public func onEmit(logRecord: ReadableLogRecord) {
     if logRecord.eventName == SessionConstants.sessionStartEvent ||
-      logRecord.eventName == SessionConstants.sessionEndEvent {
+      logRecord.eventName == SessionConstants.sessionEndEvent,
+      logRecord.attributes[SemanticConventions.Session.id.rawValue] != nil {
       // Lifecycle events already carry the historical session context they describe.
       nextProcessor.onEmit(logRecord: logRecord)
       return
--- a/Sources/Instrumentation/Sessions/SessionLogRecordProcessor.swift
+++ b/Sources/Instrumentation/Sessions/SessionLogRecordProcessor.swift
@@ -22,11 +22,18 @@ public class SessionLogRecordProcessor: LogRecordProcessor {
 
   /// Called when a log record is emitted - adds session attributes and forwards to next processor
   public func onEmit(logRecord: ReadableLogRecord) {
+    if logRecord.eventName == SessionConstants.sessionStartEvent ||
+      logRecord.eventName == SessionConstants.sessionEndEvent {
+      // Lifecycle events already carry the historical session context they describe.
+      nextProcessor.onEmit(logRecord: logRecord)
+      return
+    }
+
     var enhancedRecord = logRecord
 
     // Only add session attributes if they don't already exist
     if logRecord.attributes[SemanticConventions.Session.id.rawValue] == nil || logRecord.attributes[SemanticConventions.Session.previousId.rawValue] == nil {
-      let session = sessionManager.getSession()
+      let session = sessionManager.getSessionForAttribution()
 
       // Add session.id if not already present
       if logRecord.attributes[SemanticConventions.Session.id.rawValue] == nil {
--- a/Sources/Instrumentation/Sessions/SessionManager.swift
+++ b/Sources/Instrumentation/Sessions/SessionManager.swift
@@ -14,11 +14,6 @@ public class SessionManager: @unchecked Sendable {
     let previousSessionToEnd: Session?
   }
 
-  private enum EffectOperation {
-    case save(Session)
-    case transition(SessionTransition)
-  }
-
   private struct SessionAccess {
     let session: Session
     let shouldDrainEffects: Bool
@@ -29,7 +24,8 @@ public class SessionManager: @unchecked Sendable {
   private var persistedPreviousSession: Session?
   private let lock = NSLock()
   private let effectsLock = NSLock()
-  private var pendingEffects: [EffectOperation] = []
+  private let effectDrainQueue = DispatchQueue(label: "io.opentelemetry.sessions.effects")
+  private var pendingTransitions: [SessionTransition] = []
   private var isDrainingEffects = false
 
   /// Initializes the session manager and restores any previous session from disk
@@ -76,7 +72,7 @@ public class SessionManager: @unchecked Sendable {
   /// - Returns: The newly created session
   @discardableResult
   public func resetSession() -> Session {
-    let nextSession: Session = lock.withLock {
+    let access: SessionAccess = lock.withLock {
       let now = Date()
       let previousSession = session ?? persistedPreviousSession
       let previousSessionToEnd = previousSession.map { previous in
@@ -85,15 +81,18 @@ public class SessionManager: @unchecked Sendable {
       let nextSession = locked_startSession(previousId: previousSession?.id, at: now)
       session = nextSession
       persistedPreviousSession = nil
-      enqueueEffect(.transition(SessionTransition(
+      SessionStore.saveImmediately(session: nextSession)
+      let shouldDrainEffects = enqueueTransition(SessionTransition(
         session: nextSession,
         previousSessionToEnd: previousSessionToEnd
-      )))
-      return nextSession
+      ))
+      return SessionAccess(session: nextSession, shouldDrainEffects: shouldDrainEffects)
     }
 
-    drainPendingEffects()
-    return nextSession
+    if access.shouldDrainEffects {
+      drainClaimedTransition()
+    }
+    return access.session
   }
 
   /// Gets the current session without extending its inactivity deadline
@@ -164,19 +163,20 @@ public class SessionManager: @unchecked Sendable {
 
       let refreshedSession = locked_refreshSession(session: session, at: now)
       self.session = refreshedSession
-      enqueueEffect(.save(refreshedSession))
-      return SessionAccess(session: refreshedSession, shouldDrainEffects: true)
+      SessionStore.scheduleSave(session: refreshedSession)
+      return SessionAccess(session: refreshedSession, shouldDrainEffects: false)
     }
 
     let previousSession = session ?? persistedPreviousSession
     let nextSession = locked_startSession(previousId: previousSession?.id, at: now)
     session = nextSession
     persistedPreviousSession = nil
-    enqueueEffect(.transition(SessionTransition(
+    SessionStore.saveImmediately(session: nextSession)
+    let shouldDrainEffects = enqueueTransition(SessionTransition(
       session: nextSession,
       previousSessionToEnd: previousSession
-    )))
-    return SessionAccess(session: nextSession, shouldDrainEffects: true)
+    ))
+    return SessionAccess(session: nextSession, shouldDrainEffects: shouldDrainEffects)
   }
 
   /// Creates an ended snapshot whose end time is fixed to `date`.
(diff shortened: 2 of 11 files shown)


Sources/Instrumentation/Sessions/SessionManager.swift
```diff
@@ -21,58 +35,79 @@ public class SessionManager: @unchecked Sendable {
...
+    return accessSession(recordActivity: true)
+  }
+
+  /// Gets the session used to attribute telemetry without recording activity.
+  ///
+  /// Span, log, metric, and custom processors can use this path so passive telemetry cannot keep
+  /// a session alive. It creates or rotates an expired session before returning it.
+  /// - Returns: The current active session
+  @discardableResult
+  public func getSessionForAttribution() -> Session {
```
williazz: Issue: `peekSession()` already does this

Sources/Instrumentation/Sessions/SessionManager.swift
```diff
@@ -21,58 +35,79 @@ public class SessionManager: @unchecked Sendable {
...
+    return accessSession(recordActivity: true)
+  }
+
+  /// Records meaningful user activity and extends the current session's inactivity deadline.
+  ///
+  /// If the previous session has already expired, this creates a linked replacement before
+  /// recording the activity.
+  /// - Returns: The active session after recording activity
+  @discardableResult
+  public func recordActivity() -> Session {
```
williazz: Issue: What's the use case of `recordActivity`? The name is a little confusing because we are not recording an activity here. For clarity, maybe this should just be `refreshSession()`

Sources/Instrumentation/Sessions/SessionSpanProcessor.swift
```diff
@@ -28,10 +28,10 @@ public class SessionSpanProcessor: SpanProcessor {
   ///   - parentContext: The parent span context (unused)
   ///   - span: The span being started
   public func onStart(parentContext: SpanContext?, span: ReadableSpan) {
-    let session = sessionManager.getSession()
+    let session = sessionManager.getSessionForAttribution()
```
williazz:
```suggestion
    let session = sessionManager.getSession()
```

This is a breaking change. Also, are we sure this is the correct design decision? Let's say you have a user session that starts then is backgrounded for 10 hours, and is doing some background activity. SessionStart fires at t=0:00; SessionEnd fires at t=0:30m; then it's backgrounded from T=0:30 to 10:00. But background logs/spans from 0:30-10:00 are still attributed to the same expired session. I'm not sure if that's wise. 

From https://github.com/open-telemetry/opentelemetry-swift/issues/1180, it seems like the main issue is that backgrounded events get mischaracterized as active user events. To solve for the problem, I'd suggest that it's wiser to just add some metadata about the app state (foreground, background, etc.) as apart of the session span/log processors.

williazz (changes requested):

Sources/Instrumentation/Sessions/SessionManager.swift
aranhave: `peekSession()` is slightly different: it can return `nil` or an expired session. This path always returns a current session, creating or rotating it before attribution.

Sources/Instrumentation/Sessions/SessionManager.swift
aranhave: This is for lifecycle and interaction hooks to report meaningful user activity. It can also create or rotate a session, so `refreshSession()` felt narrower than the actual behavior. Happy to follow whichever name the maintainers prefer.

Sources/Instrumentation/Sessions/SessionSpanProcessor.swift
aranhave: @williazz The background case works a little differently here: after timeout, passive access creates a linked replacement before attribution, so it never returns the expired session. Later background spans and logs also don’t extend that replacement. App-state metadata would still be useful, but it wouldn’t fix the timeout issue because `getSession()` refreshes on every record. Does that address the concern, or should background telemetry have no session after expiry?

bryce-b:
I think this changes are at odds with the [session semantic convention](https://github.com/open-telemetry/semantic-conventions/blob/0ccc22b0046dfc24a35ec87969fc02b702f8c361/docs/general/session.md). I've left a comment on the related issue.

aranhave:
@williazz  I’ve updated this around the decision in #1180. Automatic telemetry still refreshes the session,and this PR is now focused on explicit linked reset. @bryce-b, would you both mind taking another look?

Sources/Instrumentation/Sessions/SessionLogRecordProcessor.swift
```diff
@@ -22,6 +22,14 @@ public class SessionLogRecordProcessor: LogRecordProcessor {
 
   /// Called when a log record is emitted - adds session attributes and forwards to next processor
   public func onEmit(logRecord: ReadableLogRecord) {
+    if logRecord.eventName == SessionConstants.sessionStartEvent ||
```
williazz, line 25: This is already covered by lines 35 and below but seems good to double check

Tests/InstrumentationTests/SessionTests/SessionEventInstrumentationTests.swift
```diff
@@ -388,7 +390,7 @@ final class SessionEventInstrumentationTests: XCTestCase {
         record.attributes["session.id"] == AttributeValue.string(sessions[0].id)
     }
     XCTAssertNotNil(firstStartRecord)
-    XCTAssertNil(firstStartRecord!.attributes["session.previous_id"])
+    XCTAssertNil(firstStartRecord?.attributes["session.previous_id"])
```
williazz: Nit: The optional chaining looks like a regression. We want to check that session.previous_id exists and is nil

Sources/Instrumentation/Sessions/SessionStore.swift
```diff
@@ -21,63 +21,62 @@ internal final class SessionStore {
...
 
   /// Immediately saves a session to UserDefaults
   /// - Parameter session: The session to save
   static func saveImmediately(session: Session) {
-    // Persist session. UserDefaults is thread-safe so the writes themselves
-    // don't need our lock; only the in-memory bookkeeping does.
+    lock.withLock { locked_save(session: session) }
+  }
+
+  /// Persists a session while `lock` is held.
```
williazz, line 78: These refactors are nice but kind of makes it hard to detect regressions. Could we please keep the diff narrower in the future? Or staged in different commits

Tests/InstrumentationTests/SessionTests/SessionPublicAPITests.swift
```diff
@@ -0,0 +1,9 @@
+import XCTest
+import Sessions
+
+final class SessionPublicAPITests: XCTestCase {
```
williazz, line 4: Nit: could this please be consolidated with the other tests?

Sources/Instrumentation/Sessions/SessionStore.swift
williazz: Btw the lock synchronization is nice. Did you detect any actual data races or crashes before your change? I recall that userdefaults is thread safe and wasn't causing any obvious issues

Sources/Instrumentation/Sessions/SessionManager.swift
williazz: Looks like we removed `getSessionForAttribution`

williazz (commented):
Thanks @aranhave - I'm good with the direction in this PR 

1. Seems like the purpose of this PR has shifted from "separating activity from access" to adding `SessionManager.resetSession()` and improving thread safety. Can we update the pr title and description to match the new purpose?
2. Because we are also improving/altering the thread safety - Did you test the new changes on an emulator or actual device? I've added a thread sanitizer approval workflow here - https://github.com/open-telemetry/opentelemetry-swift/pull/1188, but I recall the approval workflows are not sufficient to catch actual data races, so manual testing is needed

Sources/Instrumentation/Sessions/SessionSpanProcessor.swift
williazz: @bryce-b said it much better below in https://github.com/open-telemetry/opentelemetry-swift/pull/1183#issuecomment-5529606524 for why we use `getSession()`, but it is just a matter of accepted opinion on what a sessoin is. I think it makes sense to also allow for alternate definitions of user sessions, but either make that opt-in behavior or separate instrumentation. In fact, maybe we should also update documentation so it's clear that session here means application lifecycle. 

I still think it's worth pursuing the background metadata state since it looks like you are looking for this sort of feature support

Sources/Instrumentation/Sessions/SessionManager.swift
williazz: Looks like we removed this but it's still fair game IMO to add `refreshSession()` / `extendSession()` for syntactic sugar on `getSession()`. Just don't call it recordActivity because it sounds like we are recording telemetry about an activity, which we are not.

Sources/Instrumentation/Sessions/SessionStore.swift
aranhave: You’re right, this would have been easier to review as a separate commit or PR. I’ll split changes like this out next time.

aranhave: I didn’t start from a reported crash. `UserDefaults` is thread-safe, but the session is written across several keys and the pending-save state has its own check-then-act sequence. The lock keeps those operations together, and the contention test checks that readers never see a mixed session.

Tests/InstrumentationTests/SessionTests/SessionPublicAPITests.swift
aranhave: I kept this separate because it imports `Sessions` without `@testable`. Folding it into the existing test files would stop proving that `resetSession()` is public.

Sources/Instrumentation/Sessions/SessionLogRecordProcessor.swift
aranhave: I double-checked this by removing the guard. The fallback still calls `getSession()` when a lifecycle event has `session.id` but no `session.previous_id`, which re-enters session access. Both focused lifecycle tests fail without it, so I kept the guard.

aranhave:
> Thanks @aranhave - I'm good with the direction in this PR
> 
> 1. Seems like the purpose of this PR has shifted from "separating activity from access" to adding `SessionManager.resetSession()` and improving thread safety. Can we update the pr title and description to match the new purpose?
> 2. Because we are also improving/altering the thread safety - Did you test the new changes on an emulator or actual device? I've added a thread sanitizer approval workflow here - #1188, but I recall the approval workflows are not sufficient to catch actual data races, so manual testing is needed

Thanks @williazz. I narrowed this to linked reset and kept the existing all-signal refresh behavior. I also ran all 131 session tests with Thread Sanitizer on an ARM64 iPhone 17 simulator running iOS 26.5, with no sanitizer findings. The title and description now match the revised scope.

Also, could you rerun the failed visionOS job when you get a chance? It timed out in an unrelated DataExportWorkerTests test that this PR doesn’t touch.

Tests/InstrumentationTests/SessionTests/SessionPublicAPITests.swift
williazz: Interesting, okay. I guess other public methods can live in here too then

Sources/Instrumentation/Sessions/SessionManager.swift
```diff
@@ -21,58 +35,58 @@ public class SessionManager: @unchecked Sendable {
...
-    if sessionDidExpire {
-      if let previousSessionToEnd {
-        SessionEventInstrumentation.addSession(session: previousSessionToEnd, eventType: .end)
+  /// Ends the current session and starts a linked replacement.
+  ///
+  /// Use this after a sign-out, account change, or another explicit session boundary.
+  /// The replacement is persisted before lifecycle events are emitted.
+  /// - Returns: The newly created session
+  @discardableResult
+  public func resetSession() -> Session {
```
williazz, line 53: After signout, I think the appropriate action is to provide an `endSession()` method that fires a session end event, but does not start a new session. That said, I think this resetSession() hook is good a useful

williazz (approved):

Sources/Instrumentation/Sessions/SessionManager.swift
aranhave: Thanks, Will! Makes sense for sign-out. I’ll keep this PR focused on reset, with endSession() as a separate follow-up.

williazz: I do have another question - is this the intended behavior now?

1. Session A starts
2. Session A `Session.Start` event fires
3. resetSession()
4. Session B replaces Session A
5. After inactivity, Session B ends on next event
6. Session B `Session.End` event fires
7. Session C starts
8. Session C `Session.Start` event fires

In this case, Session A will be missing a `Session.End`, Session B will be missing a `Session.Start`, but Session C will be whole.

aranhave: Reset also emits A’s `session.end` followed by B’s `session.start`, so those happen between steps 4 and 5. B’s later expiry then ends B and starts C. The [reset test](https://github.com/open-telemetry/opentelemetry-swift/blob/d5bc61ad1346a92c06222b7ff579d8d4a6873b10/Tests/InstrumentationTests/SessionTests/SessionManagerTests.swift#L80-L104) checks that first pair.

williazz (approved):

Sources/Instrumentation/Sessions/SessionStore.swift
```diff
@@ -21,63 +21,62 @@ internal final class SessionStore {
...
-            return pending
+            locked_save(session: pending)
           }
-          return nil
-        }
-        if let pendingToSave {
-          saveImmediately(session: pendingToSave)
         }
       }
+      locked_save(session: session)
```
bryce-b, line 57: It seems like adding this `locked_save` bypasses the timed save, so I'm not sure what the purpose of the timer is at this point? Was this intentional? could you elaborate on this a bit?

bryce-b (approved):
Looks good to me, one minor clarification I'd like in `SessionStore.swift`.

Sources/Instrumentation/Sessions/SessionStore.swift
aranhave: That call only runs when saveTimer is nil. We already saved the first session immediately. this keeps that save and timer setup under the same lock. Later scheduleSave calls only update pendingSession, so the 30-second timer still handles those writes.

PR #1184, opened 2026-08-28
feat(sessions): add versioned persistence

### LINKED ISSUE #1181: Sessions: add versioned, injectable persistence and define process ownership
### Problem

[`SessionStore`](https://github.com/open-telemetry/opentelemetry-swift/blob/1ada76dc2334fc6d37880635b3f0d10cc3b40a16/Sources/Instrumentation/Sessions/SessionStore.swift#L8-L146) is a static store backed by several independent keys in `UserDefaults.standard`.

This creates three limits:

- The stored shape has no schema version or explicit migration path.
- Apps cannot inject a separate store, namespace, or App Group suite.
- Ownership is undefined when multiple `SessionManager` instances, an app and its extensions, or multiple processes can write session state.

Adding fields as separate keys can also leave a mixed record if only part of an update is observed.

### Proposed direction

Introduce an injectable persistence abstraction and store one versioned session record. Define the default ownership model and how an app may opt into sharing through an App Group or another store.

The exact storage protocol and encoding are open for discussion. The default should remain simple for a single app process, while shared storage should have documented coordination behavior rather than silently racing.

### Ordering and scope

This follows the explicit linked-reset work in [#1180](https://github.com/open-telemetry/opentelemetry-swift/issues/1180) and [#1183](https://github.com/open-telemetry/opentelemetry-swift/pull/1183), because reset introduces a replacement-session persistence path that this storage contract must preserve.

This issue does not change what counts as session activity. Automatic telemetry continues to refresh inactivity by default. Session sampling is handled separately in [#1182](https://github.com/open-telemetry/opentelemetry-swift/issues/1182).

### Acceptance criteria

- [ ] Session persistence uses a versioned record that cannot mix fields from different writes.
- [ ] Existing `otel-session-*` keys have a defined migration or fallback path.
- [ ] Corrupt records and unknown schema versions fail safely without destroying data from a newer writer.
- [ ] Callers can inject a persistence backend or `UserDefaults` suite and namespace.
- [ ] Default behavior for the main app, extensions, and multiple managers is documented.
- [ ] Opt-in shared storage defines how concurrent writers are coordinated or rejected.
- [ ] Tests cover legacy data, corrupt and unknown records, isolated stores, concurrent managers, and shared-store interleaving.

Implementation is tracked in [#1184](https://github.com/open-telemetry/opentelemetry-swift/pull/1184).

aranhave:
## Summary

- persists each session as one versioned record and migrates the legacy split `UserDefaults` keys
- adds injectable persistence, suite and namespace isolation, and an explicit ownership model
- saves session creation and reset immediately, retries failed immediate writes, and coalesces activity updates
- persists each transition before publishing its lifecycle events

## Ownership

Each persistence record has one writer. `.shared` access is rejected because the current backends cannot atomically coordinate a session transition across processes. Apps and extensions that need independent session chains should use separate namespaces.

## Stack

Review after #1183. This PR closes #1181.

### DIFF
--- a/Sources/Instrumentation/Sessions/README.md
+++ b/Sources/Instrumentation/Sessions/README.md
@@ -4,7 +4,7 @@ Automatic session tracking for OpenTelemetry Swift applications. Creates unique 
 
 ## Features
 
-- **Automatic Session Management** - Creates and manages session lifecycles with configurable timeouts
+- **Session Management** - Creates and manages session lifecycles with configurable timeouts and explicit activity
 - **Session Events** - Emits OpenTelemetry log records for session start/end events
 - **Span Attribution** - Automatically adds session IDs to all spans via span processor
 - **Persistence** - Sessions persist across app restarts using UserDefaults
@@ -18,22 +18,31 @@ Automatic session tracking for OpenTelemetry Swift applications. Creates unique 
 import Sessions
 import OpenTelemetrySdk
 
+let sessionManager = SessionManager()
+SessionManagerProvider.register(sessionManager: sessionManager)
+
 // Record session start and end events
 SessionEventInstrumentation.install()
 
 // Add session attributes to spans
-let sessionSpanProcessor = SessionSpanProcessor()
+let sessionSpanProcessor = SessionSpanProcessor(sessionManager: sessionManager)
 let tracerProvider = TracerProviderBuilder()
     .add(spanProcessor: sessionSpanProcessor)
     .build()
 
-// Add session atttributes to log records
+// Add session attributes to log records
 let sessionProcessor = SessionLogRecordProcessor(
-    nextProcessor: SimpleLogRecordProcessor(logRecordExporter: ConsoleLogRecordExporter())
+    nextProcessor: SimpleLogRecordProcessor(logRecordExporter: ConsoleLogRecordExporter()),
+    sessionManager: sessionManager
 )
 let builder = LoggerProviderBuilder()
   .with(processors: [sessionProcessor])
   .with(resource: resource)
+
+// Call this from your app's foreground callback and other meaningful interactions.
+func applicationBecameActive() {
+    sessionManager.recordActivity()
+}
 ```
 
 **Custom Configuration**:
@@ -51,7 +60,7 @@ SessionManagerProvider.register(sessionManager: sessionManager)
 **Getting Session Information**:
 
 ```swift
-// Get the current session without extending inactivity
+// Get the current session and record activity, preserving the existing API behavior
 let session = SessionManagerProvider.getInstance().getSession()
 print("Session ID: \(session.id)")
 
@@ -71,11 +80,13 @@ if let session = SessionManagerProvider.getInstance().peekSession() {
 
 ### SessionManager
 
-Manages session lifecycle with automatic expiration and explicit activity tracking.
+Manages session lifecycle with automatic expiration and explicit activity tracking. Direct
+`getSession()` calls continue to record activity for compatibility, while automatic span and log
+attribution is passive.
 
 ```swift
 let manager = SessionManager(configuration: SessionConfig(sessionTimeout: 1800))
-let session = manager.getSession() // Creates or retrieves without extending
+let session = manager.getSession() // Creates or retrieves and records activity
 manager.recordActivity() // Extends inactivity after meaningful activity
 manager.resetSession() // Starts a linked replacement
 let current = manager.peekSession() // Peek without creating or extending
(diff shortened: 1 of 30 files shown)


aranhave:
@bryce-b could you rerun the [failed macOS job](https://github.com/open-telemetry/opentelemetry-swift/actions/runs/34520396732/job/103020331060)? It aborts in the exporter retry test because the "Export has started" expectation is fulfilled more than once. I do not have permission to rerun it. I will investigate if it repeats. Thanks! :)

benjoseph-grafana (approved):

Sources/Instrumentation/Sessions/README.md
```diff
@@ -251,15 +266,25 @@ The session's end time is carried on the log record's `timestamp` field, not as
...
+- Malformed records are cleared so persistence can recover on the next write
+- Session starts and resets attempt an immediate save; rejected writes are retried
+- `endSession()` attempts to clear the saved session so it is not restored or linked on restart; rejected removals are retried
 - Access updates are coalesced and saved on a 30-second timer to minimize disk I/O
 
+### Ownership
+
+The default manager uses `UserDefaults.standard` and is intended to be the only writer in its process. Use `SessionManagerProvider` instead of creating multiple default managers. App extensions use their own standard defaults container and therefore start an independent session chain.
+
+`UserDefaultsSessionPersistence` accepts a suite and namespace, including an App Group suite, but one app or extension must own that record at a time. Requesting `.shared` access is currently rejected for every backend because a session transition requires an atomic cross-process read, update, and write. Use separate namespaces for independent processes until shared transitions have an explicit coordination API.
```
robert-northmind: Just a small comment on wording:
When we say "backend" in this sentence: 
"rejected for every backend because a session transition..", i guess we refer to the persistence/storage implementation. Is there a better name we could use here? Since maybe "backend" is implying some sort of server involved?

Sources/Instrumentation/Sessions/README.md
```diff
@@ -251,15 +266,25 @@ The session's end time is carried on the log record's `timestamp` field, not as
...
-Sessions are automatically persisted to UserDefaults and can be resumed on app restart:
+Sessions are automatically persisted and can be resumed on app restart:
 
 - By default, active persisted sessions continue from their previous state
 - Set `restorePersistedSession` to `false` to start a new session on clean start while linking and ending the persisted session
 - Expired sessions create new sessions with proper `previous_id` linking
-- Session starts and resets are saved immediately
-- `endSession()` clears the saved session so it is not restored or linked on restart
+- The built-in backend stores one versioned `Data` record instead of separate fields
+- Existing `otel-session-*` fields are migrated when first read
```
robert-northmind: Small doc nit about migration and custom persistence.

The legacy keys (`idKey`, `startTimeKey`, …) in `UserDefaultsSessionPersistence` are built from the configured `namespace`, and they're read from the `userDefaults` instance that was passed in. So the old per-field data is only migrated when the persistence uses `UserDefaults.standard` with the default `otel-session` namespace. If an app switches to a custom namespace or a suite (like the App Group example above), it looks for keys like `mobile-session-id` that no earlier SDK version wrote. The previous session is then silently not carried over, and the old `otel-session-*` keys stay in `.standard`.

Starting fresh seems like reasonable behavior to me. Could we say so explicitly, so nobody expects the session to carry over when they switch to a custom backend? Something like:

```suggestion
- Existing `otel-session-*` fields in `UserDefaults.standard` are migrated when first read by the default persistence. A `UserDefaultsSessionPersistence` with a different suite or namespace only migrates legacy fields stored under its own suite and namespace, so switching to a custom backend starts a new session chain.
```

Tests/InstrumentationTests/SessionTests/SessionStoreTests.swift
```diff
@@ -86,183 +95,489 @@ final class SessionStoreTests: XCTestCase {
...
+    let record = try decodeRecord()
+    XCTAssertEqual(record.session.id, session.id)
   }
 
   func testTeardownClearsUserDefaults() {
-    let session = Session(id: "test-session", expireTime: Date(timeIntervalSinceNow: 1800), previousId: nil, startTime: Date(), sessionTimeout: 1800, maxLifetime: 4 * 60 * 60)
-    SessionStore.saveImmediately(session: session)
+    let session = Session(id: "test-session", expireTime: Date(timeIntervalSinceNow: 1800), startTime: Date())
+    store.saveImmediately(session: session)
+    XCTAssertNotNil(persistence.read())
```
robert-northmind: Since saves now only write the new single record, the old per-field keys are never created in this test, so the `XCTAssertNil` checks after `teardown()` would pass even if the legacy cleanup were removed. Could we write all six old keys (including `previousIdKey` and `maxLifetimeKey`) before calling `teardown()`, so this actually tests the migration cleanup? Same idea for `testClearRemovesVersionedAndLegacySession`, which currently sets only four of the six. 
Non-blocking.

robert-northmind (approved):
Looks good. just left some smaller comments

robert-northmind (approved):
Looks great I think 🏅

nachoBonafonte (approved):

PR #1186, opened 2026-08-31
fix(URLSession): honour ignoredClassPrefixes

tomerhy:
Follow-up to the `ignoredClassPrefixes` finding in #1172.

The option is accepted in the initializer and stored, and then never read, so setting it excludes nothing. The delegate search only ever consults the hardcoded `excludeList`.

This wires it up. I also moved the check to before `class_copyMethodList` rather than after the method walk, because that's the expensive part — as it stood, a class you excluded still cost the full walk and was only dropped afterwards. Now excluding it actually saves the work, which matters given the search cost @williazz reported on #1172 (I measured ~285 ms for the walk over ~35k classes).

A couple of edge cases I made explicit rather than leaving to chance: an empty prefix string doesn't exclude everything, an empty array excludes nothing, and one matching prefix out of several is enough.

**On testing, since this doesn't meet the bar I'd normally hold to.** Installation happens once per process since #1157, so a test can't build a second instrumentation with a different configuration and re-run the search. The test therefore covers the exclusion decision itself, which is what the search consults for every class, rather than driving it end to end. It fails if prefix matching is removed, so it isn't vacuous, but it is a unit test of the predicate and not proof that the search consults it. Happy to hear if you'd rather see that wired differently so it can be tested properly.

The other open question is whether you want this option at all. I asked on #1172 whether to wire it up or deprecate it, and this PR takes the first option because it keeps the documented contract — but if you'd rather remove it, say so and I'll close this in favour of a deprecation.

617 tests pass.

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
@@ -131,38 +131,11 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
     let classes =
       configuration.delegateClassesToInstrument
         ?? InstrumentationUtils.objc_getClassList()
-    let selectorsCount = selectors.count
-    let ignoredPrefixes = configuration.ignoredClassPrefixes
-    DispatchQueue.concurrentPerform(iterations: classes.count) { iteration in
-      let theClass: AnyClass = classes[iteration]
-      guard theClass != Self.self else { return }
-      // Checked before the method list is copied, so an excluded class costs nothing to skip.
-      guard !Self.isExcludedFromInstrumentation(theClass, ignoredPrefixes: ignoredPrefixes) else {
-        return
-      }
-      var selectorFound = false
-      var methodCount: UInt32 = 0
-      guard let methodList = class_copyMethodList(theClass, &methodCount) else {
-        return
-      }
-      defer { free(methodList) }
 
-      var foundClasses: [AnyClass] = []
-      for j in 0 ..< selectorsCount {
-        for i in 0 ..< Int(methodCount)
-          where method_getName(methodList[i]) == selectors[j] {
-          selectorFound = true
-          foundClasses.append(theClass)
-        }
-        if selectorFound {
-          break
-        }
-      }
-
-      foundClasses.forEach { cls in
-        injectIntoDelegateClass(cls: cls)
-      }
-    }
+    Self.delegateClassesToInject(from: classes,
+                                 matching: selectors,
+                                 ignoredPrefixes: configuration.ignoredClassPrefixes)
+      .forEach { injectIntoDelegateClass(cls: $0) }
 
     if #available(OSX 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *) {
       injectIntoNSURLSessionCreateTaskMethods()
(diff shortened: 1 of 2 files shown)


Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -119,9 +132,14 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
       configuration.delegateClassesToInstrument
         ?? InstrumentationUtils.objc_getClassList()
     let selectorsCount = selectors.count
+    let ignoredPrefixes = configuration.ignoredClassPrefixes
     DispatchQueue.concurrentPerform(iterations: classes.count) { iteration in
       let theClass: AnyClass = classes[iteration]
       guard theClass != Self.self else { return }
+      // Checked before the method list is copied, so an excluded class costs nothing to skip.
+      guard !Self.isExcludedFromInstrumentation(theClass, ignoredPrefixes: ignoredPrefixes) else {
```
aranhave: The predicate test still passes if this guard is removed, which brings back the original bug. Can we exercise this wiring in a separate test process, or extract the scan step so this path is covered?

Tests/InstrumentationTests/URLSessionTests/URLSessionInstrumentationTests.swift
```diff
@@ -1242,4 +1242,38 @@ class URLSessionInstrumentationTests: XCTestCase {
...
+    XCTAssertTrue(URLSessionInstrumentation.isExcludedFromInstrumentation(cls, ignoredPrefixes: [String(name.prefix(6))]),
+                  "a matching prefix excludes the class")
+    XCTAssertTrue(URLSessionInstrumentation.isExcludedFromInstrumentation(cls, ignoredPrefixes: ["NotAMatch", String(name.prefix(6))]),
+                  "any one of the prefixes matching is enough")
+  }
+
+  /// The built in exclude list keeps working, now that it is applied in the same place.
+  public func testExcludeListStillExcludes() throws {
+    guard let proxy = NSClassFromString("__NSCFURLProxySessionConnection") else {
+      throw XCTSkip("__NSCFURLProxySessionConnection is not present on this platform")
```
aranhave: Can we avoid adding `XCTSkip` here? This class is not present on every target, so the test silently disappears there, and the repository test guidance explicitly disallows new conditional skips.

Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
tomerhy: Good catch — and you're right that flagging it in the description wasn't good enough, I shipped a test I'd already described as insufficient. Went with your second suggestion and extracted the selection into `delegateClassesToInject(from:matching:ignoredPrefixes:)`, so the search calls it and the test drives it directly. Removing the exclusion check from that function now fails the test, where before it passed. In `6b9d3ef`.

Tests/InstrumentationTests/URLSessionTests/URLSessionInstrumentationTests.swift
tomerhy: Removed. The honest reason is that I hadn't read `AGENTS.md` before starting — I've read it now, so that's a class of mistake rather than a one-off and it shouldn't recur. I left the two existing watchOS skips in this file alone since they predate this PR.

aranhave (approved):

Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -123,45 +136,56 @@ public final class URLSessionInstrumentation: @unchecked Sendable {
     let classes =
       configuration.delegateClassesToInstrument
         ?? InstrumentationUtils.objc_getClassList()
-    let selectorsCount = selectors.count
+
+    Self.delegateClassesToInject(from: classes,
+                                 matching: selectors,
+                                 ignoredPrefixes: configuration.ignoredClassPrefixes)
```
aranhave, line 142: Resuming a task still calls `instrumentDelegateClassIfNeeded` directly, without checking `ignoredClassPrefixes`. That instruments a delegate the startup scan just excluded. Could the shared injection helper enforce the prefix check too, with a test that resumes a task using an ignored delegate?

bryce-b (changes requested):
@aranhave has a [valid concern](https://github.com/open-telemetry/opentelemetry-swift/pull/1186/changes#r3971323433), could this please be addressed then we can merge this PR.

PR #1187, opened 2026-09-03
fix(otlp-http): return upload failures from log exporter

yasuradodo:
## Summary

- Make `OtlpHttpLogExporter.export` wait for the HTTP callback and return `.failure` when upload fails or times out.
- Add regression coverage for immediate failure, delayed failure, and no-response timeout paths.

This is the logs-only portion of #1146. The metric exporter change is deferred until the SIG concern about blocking metric processing is resolved.

## Motivation

Persistence deletes or keeps stored batches based on the exporter's synchronous return value. The HTTP log exporter was returning `.success` immediately after scheduling an async upload, so durable batches could be deleted before the upload outcome was known.

`OtlpHttpTraceExporter.export` already waits for the callback and returns the real result. This change aligns log export behavior with that existing contract.

## Notes

Timeout behavior remains trace-like: export returns `.failure` on timeout without adding new in-memory timeout requeue behavior.

<details>
<summary>Related work</summary>

- Supersedes the log portion of #1146
- Metrics follow-up blocked on SIG feedback about synchronous wait blocking metric collection threads
</details>

## Test plan

- [x] `swift test --filter OtlpHttpLogExporterCoverageTests`

### DIFF
--- a/Sources/Exporters/OpenTelemetryProtocolHttp/logs/OtlpHttpLogExporter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolHttp/logs/OtlpHttpLogExporter.swift
@@ -62,6 +62,7 @@ public class OtlpHttpLogExporter: OtlpHttpExporterBase, LogRecordExporter, @unch
 
   public func export(logRecords: [OpenTelemetrySdk.ReadableLogRecord],
                      explicitTimeout: TimeInterval? = nil) -> OpenTelemetrySdk.ExportResult {
+    var resultValue: ExportResult = .success
     var sendingLogRecords: [ReadableLogRecord] = []
     exporterLock.withLockVoid {
       pendingLogRecords.append(contentsOf: logRecords)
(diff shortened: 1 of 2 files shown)


Sources/Exporters/OpenTelemetryProtocolHttp/logs/OtlpHttpLogExporter.swift
```diff
@@ -62,6 +62,7 @@ public class OtlpHttpLogExporter: OtlpHttpExporterBase, LogRecordExporter, @unch
 
   public func export(logRecords: [OpenTelemetrySdk.ReadableLogRecord],
```
yasuradodo, line 63: _[Radom place]_

Trace uses [BatchSpanProcessor](https://github.com/open-telemetry/opentelemetry-swift-core/blob/main/Sources/OpenTelemetrySdk/Trace/SpanProcessors/BatchSpanProcessor.swift), and log uses [BatchLogRecordProcessor](https://github.com/open-telemetry/opentelemetry-swift-core/blob/main/Sources/OpenTelemetrySdk/Logs/Processors/BatchLogRecordProcessor.swift), but metrics don't have a similar one. Metrics uses [PeriodicMetricReaderSdk](https://github.com/open-telemetry/opentelemetry-swift-core/blob/main/Sources/OpenTelemetrySdk/Metrics/Exporter/PeriodicMetricReaderSdk.swift
)

The blocker for metrics is PeriodicMetricReaderSdk’s single scheduleQueue, which does both collection and export. 

But this PR is only about the log exporter. So this should be a safe change.

aranhave (approved):

nachoBonafonte (approved):

PR #1188, opened 2026-09-03
test: add thread sanitizer approval gate

williazz:
## Summary

Contributing this for parity with https://github.com/open-telemetry/opentelemetry-swift-core/pull/70 . I've also noticed in the past from my work on session management that our approval workflows are not great at catching crashes, so hopefully this will help some.

Adds thread sanitizer approval gate running against concurrency test suites to catch data races and regressions. 

The results are pretty positive. I only found some small issues in MetricHandlers, FileOrchestrator, OtlpTraceJsonExporter and ExporterMetrics which I resolved with locks. 

## Testing

See approval flows

Full transparency, I just had AI create a bunch of concurrency tests so that the thread sanitizer actually has some context to work with. They only run when `OTEL_CONCURRENCY_TESTS` is set to a truthy value, so they don't run unless explicitly intended.

### DIFF
--- a/Sources/Exporters/Persistence/Storage/FilesOrchestrator.swift
+++ b/Sources/Exporters/Persistence/Storage/FilesOrchestrator.swift
@@ -91,7 +91,7 @@ final class FilesOrchestrator: Sendable {
 
   func getReadableFile(excludingFilesNamed excludedFileNames: Set<String> = []) -> ReadableFile? {
     do {
-      let filesWithCreationDate = try directory.files()
+      let filesWithCreationDate = try orchestratedFiles()
         .map { (file: $0, creationDate: fileCreationDateFrom(fileName: $0.name)) }
         .compactMap { try deleteFileIfItsObsolete(file: $0.file, fileCreationDate: $0.creationDate) }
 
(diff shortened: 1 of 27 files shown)


Sources/Exporters/Persistence/Storage/FilesOrchestrator.swift
```diff
@@ -146,7 +162,7 @@ final class FilesOrchestrator: Sendable {
 
       while sizeFreed < sizeToFree, !filesWithSizeSortedByCreationDate.isEmpty {
         let fileWithSize = filesWithSizeSortedByCreationDate.removeFirst()
-        try fileWithSize.file.delete()
+        try? fileWithSize.file.delete() // may already be gone, see above
```
aranhave, line 165: A failed delete still counts as freed space, so a permission or I/O error can let storage keep growing past the limit. Can we ignore only missing-file errors here (and in the size lookup above)?

nachoBonafonte (approved):

Sources/Exporters/Persistence/Storage/FilesOrchestrator.swift
williazz: Nice catch. I noticed that it throws CocoaError.fileNoSuchFile (https://developer.apple.com/documentation/foundation/cocoaerror/filenosuchfile) so used that to handle the edge case

aranhave (approved):

Issue #107, opened 2021-02-03
Add Otel-protocol metric exporter 


bryce-b:
I'm going to start working on a metric exporter for the open-telemetry protocol. If there is any existing work or caveats I need to be aware of, please let me know.

nachoBonafonte:
That's really great news!!!
Main caveat is that the opentelemetry-swift metrics API is a bit outdated compared to the open-telemetry protocol, but you can update whatever you think makes sense.

bryce-b:
👍

bryce-b:
@nachoBonafonte I see that your original metric implementation is based off the .net solution, but they've backed away from metrics for the time being, it seems. I'm going to be referencing the java implementation going forward, and try to minimize the amount of changes overall. Right now, I'm looking at adding a Resource reference & InstrumentationLibrary reference to the MetricData object to allow for the generation of [ResourceMetrics](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L28)
Let me know if there are any concerns. If I run into trouble, I'll post in here for help. 
Thanks!

nachoBonafonte:
Yes, it is based off .net implementation, java implementation looked a bit too convoluted for me. But I am not very happy in how it ended up, so feel free to change all that makes sense for you, probably metrics is the area where I have less experience.

bryce-b:
linking #108

bryce-b:
@nachoBonafonte @SergeyKanzhelev This is off-topic for this thread, but I was intending on going to the "Governance Office Hours" on Monday and ask about starting a SIG for this agent. Is that something you guys would be interested in?

nachoBonafonte:
Yes, I think it can be useful for doing some planning and synchronization. Until now I was the only one doing regular work on the project so planning and organization has not been my highest priority.

nachoBonafonte:
Done in #108

PR #1178, opened 2026-08-27
refactor(otlp-http): compose generic OtlpHttpExporterBase over inheritance

yasuradodo:
## Motivation
I'm working on this https://github.com/open-telemetry/opentelemetry-swift/pull/1146#issuecomment-5207600952
But the code needs a lot of changes. 
To avoid a large PR, I started by refactoring the related code. 

## Summary

- Replace OTLP HTTP exporter inheritance from `OtlpHttpExporterBase` with composition via a generic `OtlpHttpExporterBase<Signal>`.
- Centralize pending queue handling (`drainPending`, `requeue`, `snapshotPending`, `dropFlushed`) and HTTP transport in the internal base helper.
- Mark trace, log, and metric HTTP exporters as `final`; keep their public initializers and export behavior unchanged.

## Test plan

- [x] `swift build --build-tests`
- [x] `swift test --filter 'OtlpHttp|OtlpHTTP'` (69 tests)

PR #1189, opened 2026-09-04
fix(otlp-http): return upload failures from metric exporter

yasuradodo:
Fixes the metrics portion of #1146: `OtlpHttpMetricExporter.export()` always returned `.success` even when the HTTP upload failed.

This PR is independent of the log exporter fix in #1187.

## Summary

- Propagate upload failures from `OtlpHttpMetricExporter.export()` instead of always returning `.success`
- Add a `PeriodicMetricReader` background export integration test that exercises OTLP HTTP metric export failure handling
- Pin `opentelemetry-swift-core` to `yasuradodo/opentelemetry-swift-core` branch `tech/periodic-metric-reader-background-export` until [open-telemetry/opentelemetry-swift-core/pull/107](https://github.com/open-telemetry/opentelemetry-swift-core/pull/107) lands

## Dependencies

- Depends on core PR: https://github.com/open-telemetry/opentelemetry-swift-core/pull/107
- `Package.swift` pins `yasuradodo/opentelemetry-swift-core` at branch `tech/periodic-metric-reader-background-export`

## Test plan

- [x] `swift test --filter OtlpHttpMetricExporter`
- [x] `swift test --filter PeriodicMetricReader`
- [x] Verify metric export failure surfaces as `.failure` (not `.success`) when the OTLP endpoint is unreachable
- [x] After core PR #107 merges, update `Package.swift` to point at the released core version and re-run the integration test

<details>
<summary>Commits</summary>

1. `fix(otlp-http): return upload failures from metric exporter`
2. `test(metrics): add PeriodicMetricReader background export integration test`
3. `build(deps): pin opentelemetry-swift-core to background-export branch`

</details>

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -30,7 +30,10 @@ let package = Package(
     .executable(name: "StableMetricSample", targets: ["StableMetricSample"])
   ],
   dependencies: [
-    .package(url: "https://github.com/open-telemetry/opentelemetry-swift-core.git", from: "2.5.1"),
+    .package(
+      url: "https://github.com/yasuradodo/opentelemetry-swift-core.git",
+      branch: "tech/periodic-metric-reader-background-export"
+    ),
     .package(url: "https://github.com/apple/swift-nio.git", from: "2.101.3"),
     .package(url: "https://github.com/grpc/grpc-swift.git", exact: "1.27.6"),
     .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.38.1"),
--- a/Sources/Exporters/OpenTelemetryProtocolHttp/metric/OtlpHttpMetricExporter.swift
+++ b/Sources/Exporters/OpenTelemetryProtocolHttp/metric/OtlpHttpMetricExporter.swift
@@ -92,6 +92,7 @@ public class OtlpHttpMetricExporter: OtlpHttpExporterBase, MetricExporter, @unch
   // MARK: - StableMetricsExporter
 
   public func export(metrics: [MetricData]) -> ExportResult {
+    var resultValue: ExportResult = .success
     var sendingMetrics: [MetricData] = []
     exporterLock.withLockVoid {
       pendingMetrics.append(contentsOf: metrics)
@@ -104,8 +105,10 @@ public class OtlpHttpMetricExporter: OtlpHttpExporterBase, MetricExporter, @unch
           metricData: sendingMetrics)
       }
     exporterMetrics?.addSeen(value: sendingMetrics.count)
+    let timeout = min(TimeInterval.greatestFiniteMagnitude, config.timeout)
+    let semaphore = DispatchSemaphore(value: 0)
     var request = createRequest(body: body, endpoint: endpoint)
-    request.timeoutInterval = min(TimeInterval.greatestFiniteMagnitude, config.timeout)
+    request.timeoutInterval = timeout
     httpClient.send(request: request) { [weak self] result in
       switch result {
       case .success:
@@ -118,10 +121,17 @@ public class OtlpHttpMetricExporter: OtlpHttpExporterBase, MetricExporter, @unch
           }
         }
         OpenTelemetry.instance.feedbackHandler?("\(error)")
+        resultValue = .failure
       }
+      semaphore.signal()
     }
 
-    return .success
+    let waitResult = semaphore.wait(timeout: .now() + timeout)
+    if waitResult == .timedOut {
+      exporterMetrics?.addFailed(value: sendingMetrics.count)
+      return .failure
+    }
+    return resultValue
   }
 
   public func flush() -> ExportResult {
--- a/Tests/ExportersTests/OpenTelemetryProtocol/PeriodicMetricReaderIntegrationTests.swift
+++ b/Tests/ExportersTests/OpenTelemetryProtocol/PeriodicMetricReaderIntegrationTests.swift
@@ -1,0 +1,175 @@
+//
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import OpenTelemetryApi
+import OpenTelemetryProtocolExporterHttp
+@testable import OpenTelemetrySdk
+import SharedTestUtils
+import XCTest
+
+final class PeriodicMetricReaderIntegrationTests: XCTestCase {
+  private let exportInterval: TimeInterval = 0.05
+  private let metricsPath = "/v1/metrics"
+  private var meterProvider: MeterProviderSdk!
+  private var testServer: HttpTestServer!
+
+  override func setUp() {
+    super.setUp()
+    testServer = HttpTestServer()
+    XCTAssertNoThrow(try testServer.start())
+  }
+
+  override func tearDown() {
+    _ = meterProvider?.shutdown()
+    meterProvider = nil
+    testServer.stop()
+    testServer = nil
+    super.tearDown()
+  }
+
+  func testTimerKeepsEnqueueingWhileExportIsBlocked() {
+    let exporter = CountingBlockingMetricExporter(aggregationTemporality: .delta)
+    let reader = PeriodicMetricReaderBuilder(exporter: exporter)
+      .setInterval(timeInterval: exportInterval)
+      .build()
+    meterProvider = MeterProviderSdk.builder()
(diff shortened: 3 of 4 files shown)


Package.swift
```diff
@@ -30,7 +30,10 @@ let package = Package(
     .executable(name: "StableMetricSample", targets: ["StableMetricSample"])
   ],
   dependencies: [
-    .package(url: "https://github.com/open-telemetry/opentelemetry-swift-core.git", from: "2.5.1"),
+    .package(
+      url: "https://github.com/yasuradodo/opentelemetry-swift-core.git",
+      branch: "tech/periodic-metric-reader-background-export"
+    ),
```
yasuradodo, line 36: TODO: Use `2.5.2` after the release

- Merge [swift-core#107](https://github.com/open-telemetry/opentelemetry-swift-core/pull/107 )
- Release a new version of core

Tests/ExportersTests/OpenTelemetryProtocol/PeriodicMetricReaderIntegrationTests.swift
```diff
@@ -0,0 +1,175 @@
...
+
+  override func tearDown() {
+    _ = meterProvider?.shutdown()
+    meterProvider = nil
+    testServer.stop()
+    testServer = nil
+    super.tearDown()
+  }
+
+  func testTimerKeepsEnqueueingWhileExportIsBlocked() {
```
yasuradodo, line 33: We may not want to have this test here because this is for [swift-core/PeriodicMetricReaderBuilder](https://github.com/open-telemetry/opentelemetry-swift-core/blob/06f8a460a66f813758d22f09025d85df45450a63/Sources/OpenTelemetrySdk/Metrics/Exporter/PeriodicMeterReaderBuilder.swift#L11-L28).
I'd like to keep it for now because it's easier for the reviewer to get the context.


| swift-core/2.5.1 | [swift-core/pull/107](https://github.com/open-telemetry/opentelemetry-swift-core/pull/107) |
----|----
Image | Image

Sources/Exporters/OpenTelemetryProtocolHttp/metric/OtlpHttpMetricExporter.swift
```diff
@@ -118,10 +121,17 @@ public class OtlpHttpMetricExporter: OtlpHttpExporterBase, MetricExporter, @unch
...
         OpenTelemetry.instance.feedbackHandler?("\(error)")
+        resultValue = .failure
       }
+      semaphore.signal()
     }
 
-    return .success
+    let waitResult = semaphore.wait(timeout: .now() + timeout)
+    if waitResult == .timedOut {
+      exporterMetrics?.addFailed(value: sendingMetrics.count)
```
aranhave, line 131: The timeout and a late failure callback both increment the failure counter. One request produced `seen=1, failed=2` locally. Could we count it once?

Package.swift
```diff
@@ -30,7 +30,7 @@ let package = Package(
     .executable(name: "StableMetricSample", targets: ["StableMetricSample"])
   ],
   dependencies: [
-    .package(url: "https://github.com/open-telemetry/opentelemetry-swift-core.git", from: "2.5.1"),
+    .package(url: "https://github.com/open-telemetry/opentelemetry-swift-core.git", from: "2.6.0"),
```
yasuradodo, line 33: This doesn't work until the release PR is merged. 
- https://github.com/open-telemetry/opentelemetry-swift-core/pull/115

Sources/Exporters/OpenTelemetryProtocolHttp/metric/OtlpHttpMetricExporter.swift
yasuradodo: You are right, and this is a race condition bug. 
If, somehow, the timeout is reached earlier than the actual callback, sometimes we get extra counts (either success or failure).

But this should be out of scope because we already have this problem in other places
- OtlpHttpMetricExporter.flush
- OtlpHttpLogExporter.export and flush
- OtlpHttpTraceExporter.export and flush

This PR introduces the bug you found, but it's a common bug in all HTTP exporters in `export` and `flush` methods. So I think we should fix them in a way that fixes all of them at once and keeps consistent behavior. 

I also think my refactoring https://github.com/open-telemetry/opentelemetry-swift/pull/1178 would help fix the bug.

What do you think?

aranhave: Fixing them together makes sense. My concern is that this PR adds the same bug to metric export. Could we prevent the double count here and add a regression test, then include that fix in #1178?

yasuradodo: Two Questions

Q1. Could you explain why you don't want to fix the bug later? 
The bug affects all other HTTP exporters and also other exporters' flush methods. To me, it's a different topic and a different bug. 

Q2. How would you fix the bug? 
I don't think there is a quick and clean solution. Probably, we would need to touch `ExporterMetrics`, which all other exporters use, to keep it maintainable.

yasuradodo:
Hi @nachoBonafonte @bryce-b 

https://github.com/open-telemetry/opentelemetry-swift-core/pull/115 has not been merged yet, so we cannot merge this PR yet.

But it would be great if you could review my solution, including the point @aranhave raised.

Sources/Exporters/OpenTelemetryProtocolHttp/metric/OtlpHttpMetricExporter.swift
aranhave: I raised it here because the new timeout path adds double-counting to metric export. I’d prefer not to introduce that here, even though it already exists in the other exporters.
I was thinking of a per-request, lock-protected guard so only the timeout or callback records the outcome, not both. That could leave ExporterMetrics unchanged, while preserving late-failure requeueing. I haven’t tested that approach yet. it would need coverage for late success and failure.
I’m open to a shared follow-up if that’s the cleaner option. Could we link an issue covering these cases and confirm whether the fix would be part of #1178 or separate?

yasuradodo: This PR fixes a separate bug: `OtlpHttpMetricExporter.export()` currently returns `.success` before the HTTP request completes, even when the request eventually fails. That incorrect result can cause callers, including persistence layers, to treat failed exports as successful and discard data. This PR aligns metric export with trace export and the existing flush behavior by waiting for the callback and returning `.failure` on failure or timeout.

I consider the double-counting issue to be _a fundamental design limitation_ in our exporter metrics handling. [ExporterMetrics](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Exporters/OpenTelemetryProtocolCommon/common/ExporterMetrics.swift) exposes independent `addSuccess` and `addFailed` operations, but it does not represent an export attempt or enforce exactly one terminal outcome per attempt. Consequently, a timeout and a late callback can both report outcomes for the same request.

Adding a metric-export-specific guard here would address one occurrence while leaving the same race in trace export and the HTTP exporters’ flush methods. I think the correct fix is to introduce a shared attempt/outcome abstraction and apply it consistently across all HTTP exporters, rather than patching only this path.

**I propose keeping this PR focused on correcting the return value from metric export**, then tracking the exporter-metrics design bug separately with explicit coverage for late success and late failure. 

I submitted the issue https://github.com/open-telemetry/opentelemetry-swift/issues/1212 and linked it to this PR.

aranhave: Thanks for opening #1212. I’m fine with keeping this PR focused on the export result and tracking the double-counting fix separately.

nachoBonafonte: Yes, I think addressing the double counting later will be better as it is orthogonal to this code

nachoBonafonte (approved):

nachoBonafonte:
Hi @yasuradodo the core PR has been merged. We plan to release a release of open telemetry-swift for version 2.6 before next week, if you want these PR merged please address the changes needed and ping me for faster review. Thanks.

PR #1192, opened 2026-09-07
Add HTTP error-only response payload recording

### LINKED ISSUE #1191: URLSessionInstrumentation: allow recording response payloads only for HTTP errors
In OpenTelemetry Swift 2.5.2, enabling payload recording causes `URLSessionInstrumentation` to buffer the response body regardless of whether the response succeeds or fails.

We would like to keep recording spans and response metadata for every request while choosing between:

- buffering all response payloads; or
- buffering payloads only for HTTP error responses (`400...599`).

## Current behavior

Every received response-data chunk is appended when `shouldRecordPayload` returns `true`:

```swift
guard configuration.shouldRecordPayload?(session) ?? false else { return }
// ...
requestMap[taskId]?.dataProcessed?.append(dataCopy)
```

Source:

https://github.com/open-telemetry/opentelemetry-swift/blob/2.5.2/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift#L677-L712

The existing callback receives only the `URLSession`, so it cannot decide based on the HTTP response status:

```swift
shouldRecordPayload: ((URLSession) -> Bool?)?
```

## Motivation

Our application needs failed-response bodies for error diagnosis, but it generally does not need successful-response bodies. The networking layer already processes the response body, so recording every payload can retain an additional in-memory copy of successful responses.

In a URLSessionInstrumentation crash with 4.7k+ events 80% of the crash events observed occurred on devices with 3 GB of RAM, particularly the iPhone SE (2nd generation) and base-model iPads. The recurring crash path contains Foundation `Data` allocation/appending, Alamofire's response handling, and OpenTelemetry's injected callback:

```text
Foundation Data allocation / append / assertion
Alamofire SessionDelegate.urlSession(_:dataTask:didReceive:)
OpenTelemetry URLSessionInstrumentation.injectTaskDidReceiveDataIntoDelegateClass
CFNetwork
```

This device distribution and stack pattern strongly support our memory-pressure suspicion. The stack does not prove that OpenTelemetry's own buffer is the exact allocation that fails, but avoiding duplicate successful-response bodies would reduce memory pressure in this path.

## Proposed behavior

Add a payload-recording mode equivalent to:

```swift
public enum ResponsePayloadRecordingMode {
  case all
  case httpErrorsOnly
}
```

Alternatively, instead of adding an enum, the configuration could expose the received response or status code through a callback:

```swift
shouldRecordPayloadForResponse: (
  URLSession,
  URLSessionDataTask,
  HTTPURLResponse
) -> Bool
```

This would let each application define its own rule, for example:

```swift
shouldRecordPayloadForResponse: { _, _, response in
  (400...599).contains(response.statusCode)
}
```

This callback would be more flexible than an enum because callers could choose any status-code range while OpenTelemetry would only need to provide the response-aware decision point.

The exact API design is open for discussion. Whichever API is chosen, the important behavior is:

- An all-payload setting preserves the current behavior.
- An error-only policy starts retaining response data only after receiving a matching HTTP response, such as `400...599`.
- Spans, status codes, headers, timing information, and other response metadata continue to be recorded for successful requests.

This would let applications retain the diagnostic value of server-error payloads without buffering every successful response body. It would remove the extra successful-response copies from our workload, although unusually large error responses would still remain unbounded by this specific proposal.

## Related GitHub discussions

- OpenTelemetry Swift PR #1169, per-request payload selection:
  https://github.com/open-telemetry/opentelemetry-swift/pull/1169

  Request-level selection cannot determine the eventual HTTP response status.

- Datadog iOS issue #3005, full response-body buffering:
  https://github.com/DataDog/dd-sdk-ios/issues/3005

AmerAtiah:
## Summary

- Add response payload recording modes for all responses or HTTP errors only.
- Preserve existing behavior by keeping all-response recording as the default.
- Avoid allocating and appending successful response payloads when HTTP-error-only recording is selected.
- Document the new configuration and cover its behavior with tests.

## Testing

- swift test (620 tests passed)

Fixes open-telemetry/opentelemetry-swift#1191

### DIFF
--- a/Sources/Instrumentation/URLSession/URLSessionInstrumentationConfiguration.swift
+++ b/Sources/Instrumentation/URLSession/URLSessionInstrumentationConfiguration.swift
@@ -23,8 +23,28 @@ public enum HTTPSemanticConvention {
   case httpDup  // Emit both old and stable (migration period)
 }
 
+/// Controls which response payloads are retained by URLSession instrumentation.
+public enum ResponsePayloadRecordingMode {
+  /// Records payloads for every response.
+  case all
+
+  /// Records payloads only for HTTP responses with a status code from 400 through 599.
+  case httpErrorsOnly
+
+  func shouldRecordPayload(for response: URLResponse?) -> Bool {
+    switch self {
+    case .all:
+      return true
+    case .httpErrorsOnly:
+      guard let response = response as? HTTPURLResponse else { return false }
+      return (400 ... 599).contains(response.statusCode)
+    }
+  }
+}
+
 public struct URLSessionInstrumentationConfiguration {
   public init(shouldRecordPayload: ((URLSession) -> (Bool)?)? = nil,
+              responsePayloadRecordingMode: ResponsePayloadRecordingMode = .all,
               shouldInstrument: ((URLRequest) -> (Bool)?)? = nil,
               nameSpan: ((URLRequest) -> (String)?)? = nil,
               spanCustomization: ((URLRequest, SpanBuilder) -> Void)? = nil,
@@ -39,6 +59,7 @@ public struct URLSessionInstrumentationConfiguration {
               ignoredClassPrefixes: [String]? = nil,
               semanticConvention: HTTPSemanticConvention = .old) {
     self.shouldRecordPayload = shouldRecordPayload
+    self.responsePayloadRecordingMode = responsePayloadRecordingMode
     self.shouldInstrument = shouldInstrument
     self.shouldInjectTracingHeaders = shouldInjectTracingHeaders
     self.injectCustomHeaders = injectCustomHeaders
@@ -65,6 +86,9 @@ public struct URLSessionInstrumentationConfiguration {
   /// Implement this callback if you want the session to record payload data, false by default.
   /// This callback is only necessary when using session delegate
   public var shouldRecordPayload: ((URLSession) -> (Bool)?)?
+
+  /// Controls whether all response payloads or only HTTP error payloads are recorded.
+  public var responsePayloadRecordingMode: ResponsePayloadRecordingMode
 
   /// Implement this callback to filter which requests you want to inject headers to follow the trace,
   /// also must implement it if you want to inject custom headers
--- a/Sources/Instrumentation/URLSession/README.md
+++ b/Sources/Instrumentation/URLSession/README.md
@@ -12,7 +12,9 @@ This behaviour can be modified or augmented by using the optional callbacks defi
 
 `shouldInstrument: ((URLRequest) -> (Bool)?)?` :  Filter which requests you want to instrument, all by default
 
-`shouldRecordPayload: ((URLSession) -> (Bool)?)?`: Implement if you want the session to record payload data, false by default.
+`shouldRecordPayload: ((URLSession) -> (Bool)?)?`: Return true to enable response payload recording for the session, false by default. Use `responsePayloadRecordingMode` to select which response payloads are recorded.
+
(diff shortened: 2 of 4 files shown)


Sources/Instrumentation/URLSession/URLSessionInstrumentationConfiguration.swift
```diff
@@ -66,6 +87,9 @@ public struct URLSessionInstrumentationConfiguration {
   /// This callback is only necessary when using session delegate
   public var shouldRecordPayload: ((URLSession) -> (Bool)?)?
 
+  /// Controls whether all response payloads or only HTTP error payloads are recorded.
+  public var responsePayloadRecordingMode: ResponsePayloadRecordingMode
```
aranhave, line 91: With `.httpErrorsOnly` enabled, `dataTask(with:completionHandler:)` still sends a 200 response's full body to `receivedResponse` (reproduced locally). Should this mode filter that path too, or should the docs say it only controls delegate buffering? #1163 covers the existing flag, but not this new mode.

aranhave (changes requested):

bryce-b (approved):

Sources/Instrumentation/URLSession/URLSessionInstrumentationConfiguration.swift
bryce-b: it seems to me `recievedResponse` will fire with the full body regardless of `shouldRecordPayload` is true or false, let alone the new configuration.

bryce-b:
@AmerAtiah can you elaborate a little bit more about what you want to accomplish with this PR? From my understanding these changes will only affect the response data called back to the  `URLSessionInstrumentationConfiguration` callbacks: 
-   `public var receivedResponse: ((URLResponse, DataOrFile?, Span) -> Void)?`
-  ` public var receivedError: ((Error, DataOrFile?, HTTPStatus, Span) -> Void)?`

and only for requests that are being handled by a URLSession delegate. 
If you're wanting to append the response data to the actual created span you can do that in these specific URLSessionInstrumentationConfguration callbacks

Sources/Instrumentation/URLSession/URLSessionInstrumentationConfiguration.swift
aranhave: Yes 💯 , that’s the existing gap #1163 addresses. My question is whether `.httpErrorsOnly` should also filter completion-handler payloads; #1163 alone doesn’t apply the new mode.

AmerAtiah: Good point, I focused on the delegate buffering because that was the path shown in our crash stack.  I've updated the pr so .httpErrorOnly also applies to the completion handler paths.

AmerAtiah:
> @AmerAtiah can you elaborate a little bit more about what you want to accomplish with this PR? From my understanding these changes will only affect the response data called back to the `URLSessionInstrumentationConfiguration` callbacks:
> 
> * `public var receivedResponse: ((URLResponse, DataOrFile?, Span) -> Void)?`
> * ` public var receivedError: ((Error, DataOrFile?, HTTPStatus, Span) -> Void)?`
> 
> and only for requests that are being handled by a URLSession delegate. If you're wanting to append the response data to the actual created span you can do that in these specific URLSessionInstrumentationConfguration callbacks

@bryce-b The goal here is a small, targeted memory optimization rather than automatically adding response bodies to spans. 

In our app, we enable payload recording because we need the body when a request fails. However, some successful responses are quite large, and OpenTelemetry creates an additional buffer for them on the delegate path. Since we don’t use successful payloads in our telemetry, that extra memory is unnecessary.

The original change only covered the delegate path because that was the path shown in our crash stack. After the feedback here, I also updated the completion-handler paths so the behavior is consistent. The app’s own completion handler still receives the original payload; this only controls what is passed to the instrumentation callbacks.

PR #1193, opened 2026-09-08
fix(Sessions): delegate forceFlush and shutdown to the wrapped processor

### LINKED ISSUE #1179: SessionLogRecordProcessor drops forceFlush and shutdown instead of delegating to nextProcessor
### What happened

`SessionLogRecordProcessor` wraps another `LogRecordProcessor` and forwards `onEmit` to it, but its
`forceFlush` and `shutdown` return `.success` without calling the wrapped processor:

```swift
// Sources/Instrumentation/Sessions/SessionLogRecordProcessor.swift
/// Shuts down the processor - no cleanup needed
public func shutdown(explicitTimeout: TimeInterval?) -> ExportResult {
  return .success
}

/// Forces a flush of any pending data - no action needed
public func forceFlush(explicitTimeout: TimeInterval?) -> ExportResult {
  return .success
}
```

The comments say "no action needed", which is true of the session processor's own state — it holds
nothing to flush. But it is the only thing the `LoggerProvider` can see. Anything it wraps becomes
unreachable for both operations.

### Why this is a problem

The recommended wiring in the module's own README puts a batching processor behind it:

```swift
let sessionProcessor = SessionLogRecordProcessor(
    nextProcessor: SimpleLogRecordProcessor(logRecordExporter: ConsoleLogRecordExporter())
)
let builder = LoggerProviderBuilder().with(processors: [sessionProcessor])
```

Substitute `BatchLogRecordProcessor` — the normal choice for a network exporter — and
`forceFlush()` silently becomes a no-op. `BatchLogRecordProcessor.forceFlush` drains its queue and
exports synchronously, so the capability exists; it just cannot be reached through the wrapper.
`shutdown` has the same problem, so a clean teardown also drops whatever is queued.

On mobile this is the expensive case: the batch processor holds up to `scheduleDelay` (5 s by
default) of records, and on iOS that buffer is lost when the app is backgrounded or terminated.
Log records are where exceptions live, so the signals most worth saving are the ones lost.

There is no error and no diagnostic — `forceFlush` returns `.success`.

### Why I think it is a bug rather than a design choice

- The protocol documents both methods as "Processes all span events that have not yet been
  processes" (`Sources/OpenTelemetrySdk/Logs/Processors/LogRecordProcessor.swift`). A conforming
  type that processes nothing and reports success does not meet that contract.
- `MultiLogRecordProcessor`, the other composing processor in the SDK, does delegate:

  ```swift
  public func forceFlush(explicitTimeout: TimeInterval? = nil) -> ExportResult {
    var result: ExportResult = .success
    logRecordProcessors.forEach {
      result.mergeResultCode(newResultCode: $0.forceFlush(explicitTimeout: explicitTimeout))
    }
    return result
  }
  ```

- `SessionSpanProcessor` has the same stubbed `forceFlush`, but that one is harmless: it takes no
  `nextProcessor` and is registered as a sibling, so `TracerProviderSdk.forceFlush` still reaches
  `BatchSpanProcessor` directly. The log-side processor is a chain, so the stub swallows the call.
  Only the log side needs fixing.

### Reproduction

```swift
let batch = BatchLogRecordProcessor(logRecordExporter: myExporter)
let session = SessionLogRecordProcessor(nextProcessor: batch)
OpenTelemetry.registerLoggerProvider(
  loggerProvider: LoggerProviderBuilder().with(processors: [session]).build()
)

logger.logRecordBuilder().setBody(.string("hello")).emit()
_ = session.forceFlush(explicitTimeout: nil)   // returns .success
// myExporter.export was never called; the record is still queued in `batch`.
```

Calling `batch.forceFlush(explicitTimeout: nil)` directly exports it, confirming the record was
reachable and the wrapper is what dropped the call.

### Suggested fix

```swift
public func shutdown(explicitTimeout: TimeInterval?) -> ExportResult {
  nextProcessor.shutdown(explicitTimeout: explicitTimeout)
}

public func forceFlush(explicitTimeout: TimeInterval?) -> ExportResult {
  nextProcessor.forceFlush(explicitTimeout: explicitTimeout)
}
```

Happy to open a PR with this plus a test if the approach looks right.

### Version

`2.5.2`, and unchanged on `main` as of 2026-08-27. iOS 26.5, Xcode 26.6, SPM.

benjoseph-grafana:
SessionLogRecordProcessor wraps another LogRecordProcessor and forwards onEmit to it, but forceFlush and shutdown returned .success without calling it. Since the session processor is the only thing the caller holds a reference to, anything it wraps became unreachable for both operations.

With the README's recommended wiring around a BatchLogRecordProcessor, forceFlush was silently a no-op and clean teardown dropped whatever was queued. On iOS that buffer is lost when the app is backgrounded or terminated, and log records are where exceptions live. There was no error and no diagnostic.

Both methods now delegate to nextProcessor, matching what MultiLogRecordProcessor already does.

Fixes #1179

### DIFF
--- a/Tests/InstrumentationTests/SessionTests/SessionLogRecordProcessorTests.swift
+++ b/Tests/InstrumentationTests/SessionTests/SessionLogRecordProcessorTests.swift
@@ -186,6 +186,7 @@ final class SessionLogRecordProcessorTests: XCTestCase {
   func testForceFlushExportsRecordsQueuedInWrappedBatchProcessor() {
     let exporter = InMemoryLogRecordExporter()
     let (provider, session) = makeBatchingProvider(exporter: exporter)
+    defer { _ = session.shutdown(explicitTimeout: nil) }
 
     provider.get(instrumentationScopeName: "test")
       .logRecordBuilder()
--- a/Tests/InstrumentationTests/SessionTests/SessionLogRecordProcessorTests.swift
+++ b/Tests/InstrumentationTests/SessionTests/SessionLogRecordProcessorTests.swift
@@ -128,14 +128,94 @@ final class SessionLogRecordProcessorTests: XCTestCase {
     }
   }
 
-  func testShutdownReturnsSuccess() {
+  func testShutdownDelegatesToNextProcessor() {
     let result = logRecordProcessor.shutdown(explicitTimeout: 5.0)
+
     XCTAssertEqual(result, .success)
+    XCTAssertEqual(mockNextProcessor.shutdownCalls.count, 1)
+    XCTAssertEqual(mockNextProcessor.shutdownCalls[0], 5.0)
   }
 
-  func testForceFlushReturnsSuccess() {
+  func testForceFlushDelegatesToNextProcessor() {
     let result = logRecordProcessor.forceFlush(explicitTimeout: 5.0)
+
     XCTAssertEqual(result, .success)
+    XCTAssertEqual(mockNextProcessor.forceFlushCalls.count, 1)
+    XCTAssertEqual(mockNextProcessor.forceFlushCalls[0], 5.0)
+  }
+
+  func testShutdownPropagatesNilTimeout() {
+    _ = logRecordProcessor.shutdown(explicitTimeout: nil)
+
+    XCTAssertEqual(mockNextProcessor.shutdownCalls.count, 1)
+    XCTAssertNil(mockNextProcessor.shutdownCalls[0])
+  }
+
+  func testForceFlushPropagatesNilTimeout() {
+    _ = logRecordProcessor.forceFlush(explicitTimeout: nil)
+
+    XCTAssertEqual(mockNextProcessor.forceFlushCalls.count, 1)
+    XCTAssertNil(mockNextProcessor.forceFlushCalls[0])
+  }
+
+  func testShutdownPropagatesFailureFromNextProcessor() {
+    mockNextProcessor.shutdownResult = .failure
+
+    let result = logRecordProcessor.shutdown(explicitTimeout: 5.0)
+
+    XCTAssertEqual(result, .failure)
+  }
+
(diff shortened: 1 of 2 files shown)


robert-northmind (approved):

Tests/InstrumentationTests/SessionTests/SessionLogRecordProcessorTests.swift
```diff
@@ -128,14 +128,94 @@ final class SessionLogRecordProcessorTests: XCTestCase {
...
+  /// the only handle the caller has, since `LoggerProviderSdk` exposes neither operation.
+  private func makeBatchingProvider(exporter: LogRecordExporter) -> (LoggerProviderSdk, SessionLogRecordProcessor) {
+    let batch = BatchLogRecordProcessor(logRecordExporter: exporter, scheduleDelay: 60)
+    let session = SessionLogRecordProcessor(nextProcessor: batch, sessionManager: MockSessionManager())
+    return (LoggerProviderBuilder().with(processors: [session]).build(), session)
+  }
+
+  func testForceFlushExportsRecordsQueuedInWrappedBatchProcessor() {
+    let exporter = InMemoryLogRecordExporter()
+    let (provider, session) = makeBatchingProvider(exporter: exporter)
```
aranhave, line 188: nit: Add a `defer` to shut down `session` after the assertions? This test starts a real `BatchLogRecordProcessor` worker, and `forceFlush()` drains the queue without stopping it.

aranhave (approved):
LGTM. Just a small nit.

Tests/InstrumentationTests/SessionTests/SessionLogRecordProcessorTests.swift
benjoseph-grafana: done.

bryce-b (approved):

